PR #3, opened 2025-04-01
fix: update broken URL in the hooks.json error message

mwbrooks:
### Summary

This pull request updates the URL in the `hooks.json` error message to use a working, valid URL. 

While the Slack CLI now support `.slack/hooks.json`, it continues to remain backwards compatible with `slack.json`. We're in a bit of a messy-in-the-middle situation where the Slack CLI often references `.slack/hooks.json` but our samples continue to use `slack.json` for backwards compatibility. 

I've updated the error messages to also suggest using `init` to create the file, when it's missing.

### Preview

**Error when a hook script returns an error code:**
image

**Error when the hooks file is missing:**
image

### Reviewer

```bash
# Create a new Deno SDK app
$ lack create my-app --template https://github.com/slack-samples/deno-starter-template
$ cd my-app/

# Error when a hook script returns an error code:
$ vim slack.json
# → Update "deno" to "dino", or some other binary name that doesn't exist
$ lack run
# → Expect an error message similar to the preview screenshot

# Error when the hooks file is missing:
$ mv slack.json slack.json.bk
$ lack run
# → Expect an error message similar to the preview screenshot

# Cleanup
$ lack delete -f
$ cd ../
$ rm -rf my-app/
```

### Requirements


internal/slackerror/errors.go
```diff
@@ -1292,12 +1296,18 @@ Otherwise start your app for local development with: %s`,
 		Code:    ErrSDKHookNotFound,
 		Message: fmt.Sprintf("A script in %s was not found", style.Highlight(filepath.Join(".slack", "hooks.json"))),
 		Remediation: strings.Join([]string{
-			fmt.Sprintf("Hook scripts are defined in the Slack hooks file ('%s').", filepath.Join(".slack", "hooks.json")),
-			"Every app requires a Slack hooks file and you can find a working example at:",
+			"Hook scripts are defined in a Slack hooks file:",
+			"- slack.json",
+			"- or",
```
zimeg:
```suggestion
			"Hook scripts are defined in a Slack hooks file found at either:",
			"- slack.json",
```

🤔 Hmm I think the list of items makes the "or" somewhat confusing for me.

I'm not sure if this is great wording, but I'm hoping to make the "pick one" option more clear in the line above!

internal/slackerror/errors.go
```diff
@@ -1292,12 +1296,18 @@ Otherwise start your app for local development with: %s`,
...
+			"- " + filepath.Join(".slack", "hooks.json"),
```
zimeg, line 1301: 🎉 Love the `filepath` support here!

internal/slackerror/errors.go
```diff
@@ -1292,12 +1296,18 @@ Otherwise start your app for local development with: %s`,
...
+			"",
+			"Every app requires a Slack hooks file and you can find an example at:",
+			"",
```
zimeg:
```suggestion
```

internal/slackerror/errors.go
```diff
@@ -1292,12 +1296,18 @@ Otherwise start your app for local development with: %s`,
...
+			style.Highlight("https://github.com/slack-samples/deno-starter-template/blob/main/slack.json"),
 			"",
-			style.Highlight("https://github.com/slack-samples/deno-starter-template/blob/main/.slack/hooks.json"),
+			"You can create a hooks file manually or with " + style.Commandf("init", false),
 			"",
```
zimeg, line 1307:
```suggestion
```

Nit: IMO newlines between some of these details separates some related ideas, but these might be helpful for parsing outputs... 👁️‍🗨️ 

No blocker but removing some might be interesting!

internal/slackerror/errors.go
```diff
@@ -1292,12 +1296,18 @@ Otherwise start your app for local development with: %s`,
...
+			"- slack.json",
+			"- or",
+			"- " + filepath.Join(".slack", "hooks.json"),
+			"",
+			"Every app requires a Slack hooks file and you can find an example at:",
+			"",
+			style.Highlight("https://github.com/slack-samples/deno-starter-template/blob/main/slack.json"),
 			"",
-			style.Highlight("https://github.com/slack-samples/deno-starter-template/blob/main/.slack/hooks.json"),
+			"You can create a hooks file manually or with " + style.Commandf("init", false),
```
zimeg:
```suggestion
			"You can create a hooks file manually or with the " + style.Commandf("init", false) + " command.",
```

This might align with the sentance following better? Of course no blocker!

internal/slackerror/errors.go
```diff
@@ -806,9 +806,13 @@ Otherwise start your app for local development with: %s`,
...
 	ErrInvalidAppDirectory: {
-		Code:        ErrInvalidAppDirectory,
-		Message:     "This is an invalid Slack app project directory",
-		Remediation: fmt.Sprintf("A valid Slack project includes the Slack hooks file: %s", filepath.Join(".slack", "hooks.json")),
+		Code:    ErrInvalidAppDirectory,
+		Message: "This is an invalid Slack app project directory",
+		Remediation: strings.Join([]string{
+			fmt.Sprintf("A valid Slack project includes the Slack hooks file: %s", filepath.Join(".slack", "hooks.json")),
+			"",
+			"If this is a Slack project, you can initialize it with " + style.Commandf("init", false),
```
zimeg, line 814:
```suggestion
			"You can initialize an existing Slack project with the " + style.Commandf("init", false) + " command",
```

I like the wording to the reader! I made a similar line removal here with some adjusted words, but might want to discuss the extra line more. Without having a similar suggestion heading this feels somewhat out of place to me...

Not meaning to block this from merging but am also hoping to lean into fewer linebreaks in output overall? It's at least interesting to me.

internal/slackerror/errors.go
mwbrooks: That's definitely a fair suggestion. I wasn't a fan of the "or" but I wanted to keep it readable.

Riffing on your suggestion, commit b9588ea updates the text to end with "Slack hooks files" but tries to emphasize that it can be one of the following:

```
   Hook scripts are defined in one of these Slack hooks files:
   - slack.json
   - .slack/hooks.json
```

internal/slackerror/errors.go
mwbrooks: I hear ya on too much separation, but I think it's important to leave a newline after a URL when possible. Depending on the person's terminal, the URL may not be styled and may wrap to multiple lines. So, leaving some whitespace after the URL can help it stand out and make the entire sentence more readable.

For example, here is a smaller window size. I've merged the previous suggestion to remove the whitespace before the URL. However, if we removed the whitespace after, it will become very unreadable:
image

internal/slackerror/errors.go
mwbrooks: Usually, we can't add text after styling a command, because it will not restore the previous colour. However, with all error messages we format the command with ticks and no colour, so I think this suggestion works! ✨

PR #6, opened 2025-04-02
ci: include compiled archives in snapshot build releases

zimeg:
### Summary

This PR uses a packaged CLI build in CI runs to match releases, using the `-s` and `-W` [flags](https://pkg.go.dev/cmd/link) to omit debug info 🤖

### Preview

The updated build is shown in the download size!

Before changes:

```
$ du dist/*
25276   dist/slack-macos_darwin_amd64_v1/bin
25248   dist/slack_linux_amd64_v1/bin
25592   dist/slack_windows_amd64_v1/bin
```

After changes:

```
$ du dist/*
17768   dist/slack-macos_darwin_amd64_v1/bin
17352   dist/slack_linux_amd64_v1/bin
17712   dist/slack_windows_amd64_v1/bin
```

### Notes

A few comments were exploring other CI happenings that other PRs hope to address 🙏 

### Requirements

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -151,7 +151,7 @@ jobs:
       - run:
           name: Run GoReleaser. Without actually publishing.
           command: |
-            export LDFLAGS="-s -w -X 'github.com/slackapi/slack-cli/internal/pkg/version.Version=$BUILD_VERSION'"
+            LDFLAGS="-s -w -X 'github.com/slackapi/slack-cli/internal/pkg/version.Version=$BUILD_VERSION'"
             make LDFLAGS="$LDFLAGS" build-snapshot
       - run:
           name: Show binaries are generated correctly


scripts/archive.sh
```diff
@@ -12,7 +12,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+set -eo
```
zimeg: 📣 This causes the archive builds **without** a known tag to fail - either before a PR-specific release is created or for `dev-build`.

https://github.com/slackapi/slack-cli/runs/39814853762:

```
Creating macOS archives:
-> Source archive: /tmp/circleci-workspace/artifacts/slack_cli_3.0.2-5-g6078ccd_macOS_64-bit.zip
-> Creating dist directory: /tmp/circleci-workspace/artifacts/slack_mac_64-bit
-> Extracting: /tmp/circleci-workspace/artifacts/slack_mac_64-bit
unzip:  cannot find or open /tmp/circleci-workspace/artifacts/slack_cli_3.0.2-5-g6078ccd_macOS_64-bit.zip, /tmp/circleci-workspace/artifacts/slack_cli_3.0.2-5-g6078ccd_macOS_64-bit.zip.zip or /tmp/circleci-workspace/artifacts/slack_cli_3.0.2-5-g6078ccd_macOS_64-bit.zip.ZIP.

Exited with code exit status 9
```

.circleci/config.yml
```diff
@@ -33,7 +33,7 @@ commands:
       - run:
           name: Output BUILD_VERSION env var based on recent git tag and type of build
           command: |
-            BUILD_VERSION=$(git describe --tags --match 'v*.*.*' --exclude '*feature');
```
zimeg: 📣 This caused uploads to the `dev-build` release to fail with this tag not matching `goreleaser` outputs:

```
    • archiving                                      name=dist/slack_cli_3.0.2-main-build-test-feature-6-gcc496f3_macOS_64-bit.tar.gz
    • archiving                                      name=dist/slack_cli_3.0.2-main-build-test-feature-6-gcc496f3_linux_64-bit.tar.gz
    • archiving                                      name=dist/slack_cli_3.0.2-main-build-test-feature-6-gcc496f3_windows_64-bit.zip
```

```
$ git describe --tags --match 'v*.*.*' --exclude '*feature'
v3.0.2-7-gcc496f3
```

zimeg: Now, the `BUILD_VERSION` variable matches the tag used elsewhere:

```
$ git describe --tags --match 'v*.*.*'
v3.0.2-main-build-test-feature-6-gcc496f3
```

zimeg: FWIW I tried to filter the following but had no luck:

- v1.23.4
- v1.23.4-example-feature


But not:

- v1.23.3-example
- v1.23.3-example-feature-1
- branch

And cannot find a good command combination...

I think it'd be better to **not** increment after a feature tag since IIRC these are meant for "stable experiements", but finding the right command and making related changes elsewhere might require more follow up 🔍

.circleci/config.yml
```diff
@@ -152,7 +152,7 @@ jobs:
           name: Run GoReleaser. Without actually publishing.
           command: |
             export LDFLAGS="-s -w -X 'github.com/slackapi/slack-cli/internal/pkg/version.Version=$BUILD_VERSION'"
-            make build-snapshot
+            make LDFLAGS="$LDFLAGS" build-snapshot
```
zimeg, line 155: The `-s` and `-w` flag weren't being used when building the snapshot! I'm hoping this improves our runtime speeds around 👾 

**Before**:

```
$ du dist/*
...
13324   dist/slack_cli_3.0.2-main-build-test-feature-6-gcc496f3_linux_64-bit.tar.gz
13332   dist/slack_cli_3.0.2-main-build-test-feature-6-gcc496f3_macOS_64-bit.tar.gz
14376   dist/slack_cli_3.0.2-main-build-test-feature-6-gcc496f3_windows_64-bit.zip
```

**After**:

```
$ du dist/*
...
7156    dist/slack_cli_3.0.2-main-build-test-feature-6-gcc496f3_linux_64-bit.tar.gz
7252    dist/slack_cli_3.0.2-main-build-test-feature-6-gcc496f3_macOS_64-bit.tar.gz
7208    dist/slack_cli_3.0.2-main-build-test-feature-6-gcc496f3_windows_64-bit.zip
```

.circleci/config.yml
mwbrooks: Hmm... I think it's important for us to filter out the `-feature` tags. Otherwise, someone who tags a feature will cause the `dev-build` to start building off their feature flag instead of the previous stable release.

mwbrooks: > 📣 This caused uploads to the `dev-build` release to fail with this tag not matching `goreleaser` outputs:
> 
> ```
>     • archiving                                      name=dist/slack_cli_3.0.2-main-build-test-feature-6-gcc496f3_macOS_64-bit.tar.gz
>     • archiving                                      name=dist/slack_cli_3.0.2-main-build-test-feature-6-gcc496f3_linux_64-bit.tar.gz
>     • archiving                                      name=dist/slack_cli_3.0.2-main-build-test-feature-6-gcc496f3_windows_64-bit.zip
> ```
> 
> ```
> $ git describe --tags --match 'v*.*.*' --exclude '*feature'
> v3.0.2-7-gcc496f3
> ```

I'm catching up on your thinking here, but from my understanding the `dev-build` should not be off the above feature flag:

```
name=dist/slack_cli_3.0.2-main-build-test-feature-6-gcc496f3_macOS_64-bit.tar.gz
```

It should be off the latest stable on main, which would look something like:

```
# Making the number -12- up, but it should be equal to or higher than the commit count of the feature flag
name=dist/slack_cli_3.0.2-12-gcc496f3_macOS_64-bit.tar.gz
```

.circleci/config.yml
```diff
@@ -313,10 +313,10 @@ jobs:
...
               LAST_SEMVER_TAG=$(git describe --tags --match 'v*.*.*' --abbrev=0 | cut -d"-" -f1)
               # When generating release notes, setting the target tag_name property to an existing tag has GitHub ignore the target_commitish property
-              # So, set it to some non-existent tag name for release-note-generating purposes.
+              # So, set the tag to the commit hash for release-note-generating purposes.
               # See https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#generate-release-notes-content-for-a-release--parameters
               # Note the following _does not change releases or tags_ - it only creates release notes, just like clicking "Generate Release Notes" on the GitHub Releases page.
-              CHANGELOG=$(curl -L -X POST -H "Accept: application/vnd.github+json" -H "Authorization: Bearer $GITHUB_RELEASE_TOKEN" -H "X-GitHub-Api-Version: 2022-11-28" https://api.github.com/repos/slackapi/slack-cli/releases/generate-notes -d "{\"tag_name\":\"does-not-exist\",\"target_commitish\":\"${TARGET}\",\"previous_tag_name\":\"${LAST_SEMVER_TAG}\"}" | jq .body)
```
mwbrooks: I think this value was originally set, because the `dev-build` tag never changes. If we used the `dev-build` tag, then the CHANGELOG would include all commits back to the original, very old `dev-build` tag. However, we may not have been using the `target_commitish` tag originally and that may solve the problem.

What's the reason for this change? I just want to make sure it doesn't introduce any bugs.

.circleci/config.yml
zimeg: @mwbrooks Thank you for confirming this strangeness!

I was wanting to avoid making a wider change to tags following a feature build, but now does seem like a good time to get this right.

I'll update this with scripts as needed 🫡

.circleci/config.yml
zimeg: I'm TBH not so certain that `${TARGET}` will generate the CHANGELOG we expect, but this has hopes of creating release notes with:

> Full Changelog: https://github.com/slackapi/slack-cli/compare/v3.0.3...70e82233e3a7d8afcec6b83ab37f78abfc34a17e

Instead of:

> Full Changelog: https://github.com/slackapi/slack-cli/compare/v3.0.3...does-not-exist

Since this change was found in PR explorations here, I'll move it to another PR and confirm this does what we expect 🙏

zimeg: 📝 This was undone in https://github.com/slackapi/slack-cli/pull/6/commits/3e3cf9f6035c0d51ff0c6ec3e0e8bbf17aa69a58 that has hope to follow up in another PR!

.circleci/config.yml
zimeg: 📝 The current changes of this PR aren't so related to the tag naming, though still might pose problems on `main`, so I'll mark this as **resolved** with hopes of revising!

scripts/archive.sh
zimeg: 📝 This was undone in https://github.com/slackapi/slack-cli/pull/6/commits/3e3cf9f6035c0d51ff0c6ec3e0e8bbf17aa69a58 but hopes to move to another PR soon for discussion and related changes-

.circleci/config.yml
mwbrooks: Great catch @zimeg! 👏🏻 Before this change, the `export LDFLAGS` was unused and instead it was using the `LDFLAGS` set by `Makefile`'s startup.

PR #7, opened 2025-04-02
refactor: Add slackcontext package to DRY context management

mwbrooks:
### Summary

This pull request introduces the `package slackcontext` to allow context to be managed in one place.

**Usage:**
* Any package can **get** context/state values using `slackcontext.<Function>`
* Any package can **set** context/state values using `slackcontext.<SetFunction>`
* Every command is executed with the context (`.ExecuteContext(ctx)`) so `cmd.Context()` returns the correct context

**Thoughts:**
* We don't want to over-use `ctx`
* There are split opinions in the Golang community on what can go into `ctx`
* Common theme is that it should contain the state for a specific "request"
  * In HTTP world, that would be a request to a HTTP route
  * In CLI world, it's less clear, but likely the execution of the command
* Since `ctx` is immutable, any function that modifies it must return the new `ctx`
* For this reason, it's best to reduce the amount that `ctx` is updated
* Right now, we are only updating the `ctx` before `rootCmd.ExecuteContext(ctx)`

**Rule of thumb:**
* For now, I think it'll be best to assume `ctx` doesn't change after `rootCmd.ExecuteContext(ctx`) is called
* The `ctx` should not contain required function args, such as `fs` or `os` so we'll need to think about where these should live long-term

### Reviewers

Sorry about the massive PR. Most of the PR is updating tests to use `slackcontext.MockContext(ctx)` instead of `context.Background()`. This is because values such as `Version` are required by the API Client.

There should be no change to functionality.

### Requirements


cmd/env/list.go
```diff
@@ -51,7 +51,7 @@ func NewEnvListCommand(clients *shared.ClientFactory) *cobra.Command {
 			return preRunEnvListCommandFunc(ctx, clients)
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return runEnvListCommandFunc(clients)
+			return runEnvListCommandFunc(clients, cmd)
```
mwbrooks, line 54: While `runEnvListCommandFunc(ctx, clients, cmd)` would be a more appropriate format, the other `env` commands only pass `cmd` and fetch the context with `cmd.Context()`.

This change is leaning on consistency and a smaller change footprint, since the PR is already quite large.

cmd/env/list.go
```diff
@@ -74,8 +74,9 @@ func preRunEnvListCommandFunc(ctx context.Context, clients *shared.ClientFactory
 // runEnvListCommandFunc outputs environment variables for a selected app
 func runEnvListCommandFunc(
 	clients *shared.ClientFactory,
+	cmd *cobra.Command,
 ) error {
-	ctx := context.Background()
+	ctx := cmd.Context()
```
mwbrooks, line 79: Fixed getting the real context, so that the `env list` command can access the Slack CLI Version. Currently, all API requests for the `env list` command don't include the Version in the user-agent.

cmd/root.go
```diff
@@ -261,43 +262,35 @@ func InitConfig(clients *shared.ClientFactory, rootCmd *cobra.Command) error {
...
 	clients.Config.LogstashHostResolved = clients.AuthInterface().ResolveLogstashHost(ctx, clients.Config.ApiHostResolved, clients.CliVersion)
 
-	// TODO: should we store system ID in config or in context?
 	// Init System ID
 	if systemID, err := clients.Config.SystemConfig.InitSystemID(ctx); err != nil {
 		clients.IO.PrintDebug(ctx, "Error initializing user-level config system_id: %s", err.Error())
 	} else {
 		// Used by Logstash
+		// TODO(slackcontext) Consolidate storing SystemID to slackcontext
 		clients.Config.SystemID = systemID
```
mwbrooks, line 271: This would be a larger refactor that can wait until we 🪓 axe storing state in `clients.Config`

cmd/root.go
```diff
@@ -261,43 +262,35 @@ func InitConfig(clients *shared.ClientFactory, rootCmd *cobra.Command) error {
...
 		// Debug logging
 		clients.IO.PrintDebug(ctx, "system_id: %s", clients.Config.SystemID)
 	}
 
-	// TODO: should we store project ID in config or in context?
 	// Init Project ID, if current directory is a project
 	if projectID, _ := clients.Config.ProjectConfig.InitProjectID(ctx, false); projectID != "" {
 		// Used by Logstash
+		// TODO(slackcontext) Consolidate storing ProjectID to slackcontext
 		clients.Config.ProjectID = projectID
```
mwbrooks, line 283: Likewise, this can wait until we 🪓 axe storing state in `clients.Config`

cmd/root.go
```diff
@@ -334,10 +327,8 @@ func InitConfig(clients *shared.ClientFactory, rootCmd *cobra.Command) error {
 // listens for process interrupts and sends to IOStreams' GetInterruptChannel() for use in
 // in communicating process interrupts elsewhere in the code.
 func Execute(ctx context.Context, rootCmd *cobra.Command, clients *shared.ClientFactory) {
-	// TODO: ensure context is only used with setters and not with getters. After investigating, report:
-	// - internal/contextutil/contextutil.go has a `VersionFromContext` getter method which is used in various API methods when setting the user agent on HTTP requests, and when sending data up to logstash
-	// - internal/config/context.go has a variety of set and get methods related to app, token, team/enterprise/session/user/trace IDs, domains
```
mwbrooks, line 339: These files have been refactored to `slackcontext` and the comment is no longer relevant.

cmd/root.go
```diff
@@ -366,6 +357,7 @@ func Execute(ctx context.Context, rootCmd *cobra.Command, clients *shared.Client
 			}()
 		case <-ctx.Done():
 			// No interrupt signal sent, command executed to completion
+			// FIXME - `.Done()` channel is triggered by `cancel()` and not a successfully completion
```
mwbrooks, line 360: This is a bug that i noticed while learning more about context. The `.Done()` is misleading - it's called when the context is cancelled NOT when the execution is complete.

cmd/root_test.go
```diff
@@ -21,12 +21,15 @@ import (
...
 	"github.com/slackapi/slack-cli/internal/shared"
+	"github.com/slackapi/slack-cli/internal/slackcontext"
 	"github.com/slackapi/slack-cli/test/testutil"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
 func TestRootCommand(t *testing.T) {
+	ctx := slackcontext.MockContext(t.Context())
+
```
mwbrooks, line 32: I've introduced a `slackcontext.MockContext(ctx)` helper to setup a context with the guaranteed, default values set before a command executes.

internal/api/client.go
```diff
@@ -114,7 +114,11 @@ func (c *Client) postForm(ctx context.Context, endpoint string, formValues url.V
 	if err != nil {
 		return nil, err
 	}
-	cliVersion := contextutil.VersionFromContext(ctx)
+
+	cliVersion, err := slackcontext.Version(ctx)
+	if err != nil {
+		return nil, err
+	}
```
mwbrooks, line 121: These errors are bubbled up to the original caller. Previously, when version didn't exist, the API calls would be made with a `""` empty string.

internal/config/context.go
```diff
@@ -16,42 +16,15 @@ package config
...
 import (
 	"context"
-
-	"github.com/slackapi/slack-cli/internal/shared/types"
 )
 
 type contextKey string
 
-const CONTEXT_ENV contextKey = "env"
 const CONTEXT_TOKEN contextKey = "token"
```
mwbrooks, line 23: The remaining functions in this file don't belong in the `slackcontext` because they deal with run-time state - selected auth, team, app, etc. We may want to refactor this into a standardized selected app struct that can be passed around.

internal/slackcontext/slackcontext_test.go
```diff
@@ -0,0 +1,331 @@
...
+	"testing"
+
+	"github.com/google/uuid"
+	"github.com/opentracing/opentracing-go"
+	"github.com/slackapi/slack-cli/internal/slackerror"
+	"github.com/slackapi/slack-cli/internal/tracer"
+	"github.com/stretchr/testify/require"
+)
+
+func Test_SlackContext_OpenTracingSpan(t *testing.T) {
```
mwbrooks, line 28: Slightly disappointed that removing the untested `contextutil.go` and some of `context.go` didn't result in a higher test coverage increase. Oh well 🤷🏻

main.go
```diff
@@ -23,63 +23,68 @@ import (
...
 	"github.com/slackapi/slack-cli/internal/tracer"
 	"github.com/uber/jaeger-client-go"
 )
 
 func main() {
 	defer recoveryFunc()
 
+	// Create the parent context for the CLI execution
 	var ctx = context.Background()
+
```
mwbrooks, line 41: This should be the one and only time we call `context.Background()`. In a future PR, I'll scrub the code base for all references to it, because I'm sure there are still some.

main.go
```diff
@@ -23,63 +23,68 @@ import (
...
+// TODO(slackcontext) Use closure to pass in the ctx, which includes the sessionID
 func recoveryFunc() {
 	// in the event of a panic, log panic
 	if r := recover(); r != nil {
 		var clients = shared.NewClientFactory(shared.SetVersion(version.Raw()))
+
 		var ctx = context.Background()
-		ctx = config.SetContextSessionID(ctx, uuid.New().String())
+		ctx = slackcontext.SetSessionID(ctx, uuid.New().String())
+
```
mwbrooks, line 87: In a follow-up PR, I'd like to clean this up so that we can use the existing `SessionID` rather than generating a new one when there is an unrecoverable error.

cmd/root_test.go
zimeg: @mwbrooks This is wonderful with the setup in `root`!

https://github.com/slackapi/slack-cli/blob/aae77de3b02a46739b7d50af095c7d81ee8bc532/internal/slackcontext/slackcontext_mock.go#L23-L34

cmd/root.go
zimeg: I'm super curious about this 👀

It's not super clear what the bug is but I'm not doubting that this might be causing problems... Were you noticing something strange during command completion?

cmd/root.go
```diff
@@ -261,43 +262,35 @@ func InitConfig(clients *shared.ClientFactory, rootCmd *cobra.Command) error {
...
 		clients.Config.ProjectID = projectID
-
 		// Used by OpenTracing
-		if span := opentracing.SpanFromContext(ctx); span != nil {
-			span.SetTag("project_id", clients.Config.ProjectID)
-			ctx = opentracing.ContextWithSpan(ctx, span)
-		}
-
+		ctx = slackcontext.SetProjectID(ctx, projectID)
+		rootCmd.SetContext(ctx)
```
zimeg, line 286: 👾 ✨

cmd/root_test.go
```diff
@@ -38,7 +41,7 @@ func TestRootCommand(t *testing.T) {
 	clientsMock := shared.NewClientsMock()
 	testutil.MockCmdIO(clientsMock.IO, cmd)
 
-	err := cmd.Execute()
+	err := cmd.ExecuteContext(ctx)
```
zimeg, line 44: TIL! Is this the same as `Execute`, but sets the `ctx` on the command?

Edit: Just saw the note in the PR description - this is neat! 🙏 📚

cmd/root.go
mwbrooks: I only noticed the "bug" 🐛 while reading the code and didn't test it.

From my understanding, the commenter thought `ctx.Done()` is called when the execution is completed, but it's actually called if `ctx.Cancel()` is called. I don't see any references to `ctx.Cancel()` so this switch case is never actually met. Regardless, if it was met, it falls through to the successful completion switch case. So functionally, everything would work fine.

So, the main bug is just the comment that's misleading. I'll make a note to follow-up on this and check if we should be cancelling the context at some point 🤔

cmd/root.go
mwbrooks: I don't _think_ we need to do this, because the `rootCmd.ExecuteContext(ctx)` will execute the command with our context. However, if someone called `rootCmd.Context()` before executing the Cobra command, they'll at least get the correct context.

cmd/root_test.go
mwbrooks: You got it! 

* `cmd.Execute()` will use a `context.Background()` (empty context)
* `cmd.ExecuteContext(ctx)` will use `ctx`

Whenever someone called `cmd.Context()` they'll get whatever context has been set. We want to ensure that they always get our context.

PR #9, opened 2025-04-04
docs(readme): add the Slack CLI logo to the README.md

mwbrooks:
### Summary

This pull request adds the Slack CLI logo to the title of the README.md file to match [slackapi/bolt-js](https://github.com/slackapi/bolt-js) and [slackapi/bolt-python](https://github.com/slackapi/bolt-python).

### Reviewers

* **[Preview of README.md](https://github.com/slackapi/slack-cli/blob/e2ccc7c309758e9a1f43a65f17cf0b3c5afdbcf8/README.md)**
* **Updated `.gitignore` to no longer ignore the `docs/*`**
  * We were ignoring the `docs/*` because the command `docgen` uses `docs/` as the default directory
  * If we start to store our public docs in this repo, we'll need to use the `docs/` directory as well
  * So, I felt it's okay to stop ignoring this directory
  * Thoughts?
* **Logo is offset in Chrome on macOS**
  * Do you notice that the logo is offset vertically a little bit?
  * If I add `margin-bottom: -0.1em;` then it fixes it on my system, but this could be different on each system
  * I'm leaning toward just keeping it as being a little offset
  * Thoughts?

### Requirements

### DIFF
--- a/.licenserc.yml
+++ b/.licenserc.yml
@@ -26,6 +26,7 @@ header:
     - '**/*.json'
     - '**/*.md'
     - '**/*.tmpl'
+    - '**/*.svg'
     - 'dotgitignore'
     - 'go.sum'
     - 'go.mod'


README.md
```diff
@@ -1,4 +1,4 @@
-# Slack CLI
+# Slack CLI <img src="docs/static/img/slack-cli-logo.svg" alt="Slack CLI logo" style="height: 1em; max-width: 100%" />
```
WilliamBergamin, line 1: Just personnel preference but also sharing some of my past explorations, the title can be centered by using html tags as well,  [bolt-python](https://github.com/slackapi/bolt-python/blob/main/README.md), [python-sdk](https://github.com/slackapi/python-slack-sdk/blob/main/README.md), [python-hooks](https://github.com/slackapi/python-slack-hooks/blob/main/README.md))
```suggestion
<h1 align="center">Slack CLI <img src="docs/static/img/slack-cli-logo.svg" alt="Slack CLI logo" style="height: 1em; max-width: 100%" /></h1>
```

mwbrooks: The center alignment is a nice touch of the Python projects! 👌🏻 

I've tried it with this README.md and it doesn't look quite as slick. Below, I've centered the title, status badges, and quote description.

What are your thoughts on this style @WilliamBergamin @lukegalbraithrussell? My gut says all 3 should be centered otherwise none. The quote description looks funny, but it looks better if I switch it to italics.

**Centered Title:**

image

**Centered Title & Status:**

image

**Centered Title, Status, and Description:**

image

**Centered Title, Status, and Description (Italics):**

image

docs/static/img/slack-cli-logo.svg
zimeg, line 1: ❤️

README.md
zimeg: 📝 Adding onto this to prefer using `#` for title headings and following sections for standard parsings that some readers might use:

https://github.com/artempyanykh/marksman/blob/main/docs/features.md#titles-from-headings

zimeg: If this is centered, I think the last option is best with perhaps swapped lines for the status and description?

I'm not sure of the markup needed for this though.

.gitignore
```diff
@@ -1,6 +1,5 @@
 # Project build files
 bin/
-docs/
```
zimeg, line 3: Removing this makes sense to me with the other documentation changes in progress! IIRC additional written pages - not autogen - might soon find this path.

We might want to consider either committing the autogen files or ignoring these here though. I lean towards ignoring since the build process for docs uses this command itself! Otherwise, we might want to fail CI if these files don't match whats expected...

No blocker for this PR though! These files are find to keep unstaged after `docgen` for now 📚 ✨

mwbrooks: I agree that we may want to change the default path for `docgen` and/or `.gitignore` the files.

Since `docgen` is a hidden command, I think we can safely change the default directory with a `patch` or `minor`. Let's hold off until we move the public docs over, then we can decide what makes the most sense!

README.md
mwbrooks: @zimeg Thanks for sharing those links! 🙇🏻 

Since using HTML can lead to parsing problems for some readers AND popular Golang projects use standard markdown, let's align with the Golang community and continue to use a `#`.

I think we could consider adding a header image that uses the Slack CLI Logo and text, to follow in the foot steps of other popular Golang projects:
* https://github.com/spf13/cobra?tab=readme-ov-file
* https://github.com/charmbracelet/lipgloss?tab=readme-ov-file

zimeg:
> While I expected images to align, this might be something that's even changed later within the browser!

📝 Huh this happened sooner than I had guessed... https://developer.mozilla.org/en-US/blog/h1-element-styles/

> **Default styles for h1 elements are changing**

PR #14, opened 2025-04-04
test: fix collaborator command test to run standalone

mwbrooks:
### Summary

This pull request fixes a flaky test in `cmd/collaborators.go`. 

Previously, the test would only pass when other `package collaborators` tests ran first. After looking into the test, I realized that the test was outdated. Since the `collaborators` command executes `collaborators list` under the hood, I copied the table test from `collaborators list`.

### Requirements

### DIFF
--- a/cmd/collaborators/collaborators_test.go
+++ b/cmd/collaborators/collaborators_test.go
@@ -15,68 +15,91 @@
 package collaborators
 
 import (
-	"context"
+	"fmt"
 	"testing"
 
+	"github.com/slackapi/slack-cli/internal/hooks"
+	"github.com/slackapi/slack-cli/internal/prompts"
 	"github.com/slackapi/slack-cli/internal/shared"
 	"github.com/slackapi/slack-cli/internal/shared/types"
-	"github.com/slackapi/slack-cli/test/testutil"
-	"github.com/stretchr/testify/assert"
+	"github.com/slackapi/slack-cli/internal/slacktrace"
 	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
 )
 
 func TestCollaboratorsCommand(t *testing.T) {
-	// Create mocks
-	clientsMock := shared.NewClientsMock()
-	clientsMock.AddDefaultMocks()
-
-	clientsMock.ApiInterface.On("ListCollaborators", mock.Anything, mock.Anything, mock.Anything, mock.Anything).Return([]types.SlackUser{}, nil)
-
-	// Create clients that is mocked for testing
-	clients := shared.NewClientFactory(clientsMock.MockClientFactory())
-	mockAuths := []types.SlackAuth{}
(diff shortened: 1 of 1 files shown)


cmd/collaborators/collaborators_test.go
```diff
@@ -15,68 +15,91 @@
...
-
-	// Create the command
-	cmd := NewCommand(clients)
-	testutil.MockCmdIO(clients.IO, cmd)
-
-	// Execute test
-	err := cmd.Execute()
-	if err != nil {
-		assert.Fail(t, "cmd.Execute had unexpected error")
+	tests := map[string]struct {
```
mwbrooks, line 31: This test is a copy and paste of `cmd/collaborators/list_test.go` with one small change that I mention as a GitHub reviewer comment

cmd/collaborators/collaborators_test.go
```diff
@@ -15,68 +15,91 @@
...
+			appSelectMock.On("TeamAppSelectPrompt").Return(prompts.SelectedApp{App: tt.App, Auth: types.SlackAuth{}}, nil)
+			clientsMock := shared.NewClientsMock()
+			clientsMock.AddDefaultMocks()
+			clientsMock.ApiInterface.On("ListCollaborators", mock.Anything, mock.Anything, mock.Anything).
+				Return(tt.Collaborators, nil)
+			clients := shared.NewClientFactory(clientsMock.MockClientFactory(), func(clients *shared.ClientFactory) {
+				clients.SDKConfig = hooks.NewSDKConfigMock()
+			})
+
+			err := NewCommand(clients).Execute()
```
mwbrooks, line 88: Changed this from `NewListCommand(clients).Execute()` to `NewCommand(clinets).Execute()`.

cmd/collaborators/collaborators_test.go
```diff
@@ -15,68 +15,91 @@
...
+				fmt.Sprintf("%d", len(tt.Collaborators)),
+			})
+			for _, collaborator := range tt.Collaborators {
+				clientsMock.IO.AssertCalled(t, "PrintTrace", mock.Anything, slacktrace.CollaboratorListCollaborator, []string{
+					collaborator.ID,
+					collaborator.UserName,
+					collaborator.Email,
+					string(collaborator.PermissionType),
+				})
+			}
```
zimeg, line 102: 👾 I like that we're checking the stable value used in tests here instead of `stdout` although I do wonder if we should make note of this to follow back up on?

I don't expect these outputs to change so often, but confirming certain values **are** output is somewhat nice in unit tests and IMO not so much of a hassle to update when changing related code.

No blocker at all but avoiding regression is a constant on mind ✨

mwbrooks: Good point! I can follow-up with another PR that updates this test for both `collaborators` and `collaborators list` 👌🏻

PR #24, opened 2025-04-09
test: mock context in TableTestCommand.ExpectedAsserts

mwbrooks:
### Summary

This pull request updates `test/testutil/commandtests.go` to pass the mocked context into `ExpectedAsserts`. 

It's a lot of repetitive lines changes with the occasional 💎 gem hidden. Don't worry though, I'll leave GitHub comments to mark the non-repetitive areas! 🚩 

The motivation is to use the mocked context throughout the table test's lifecycle. Right now, some `ExpectedAsserts: func(...)` implementations require a context and use `context.Background()`. This works today, but it would fail a function required a value from `ctx`.

### Requirements


cmd/app/add_test.go
```diff
@@ -352,7 +352,7 @@ func TestAppAddCommand(t *testing.T) {
 				mockProjectConfig.On("Cache").Return(mockProjectCache)
 				cm.Config.ProjectConfig = mockProjectConfig
 			},
-			ExpectedAsserts: func(t *testing.T, cm *shared.ClientsMock) {
+			ExpectedAsserts: func(t *testing.T, ctx context.Context, cm *shared.ClientsMock) {
```
mwbrooks, line 355: note: Every reference to `ExpectedAsserts: func(...)` had to be updated to include `ctx`. 🔁

cmd/app/link_test.go
```diff
@@ -95,15 +95,15 @@ func Test_Apps_Link(t *testing.T) {
...
+			ExpectedAsserts: func(t *testing.T, ctx context.Context, cm *shared.ClientsMock) {
 				expectedApp := types.App{
 					AppID:        mockLinkAppID1,
 					TeamDomain:   mockLinkSlackAuth1.TeamDomain,
 					TeamID:       mockLinkSlackAuth1.TeamID,
 					EnterpriseID: mockLinkSlackAuth1.EnterpriseID,
 				}
 				actualApp, err := cm.AppClient.GetDeployed(
-					context.Background(),
+					ctx,
```
mwbrooks, line 106: note: The main motivation of this PR is to replace `context.Background()` references with the mocked context `ctx`. Here's one example.

cmd/collaborators/update_test.go
```diff
@@ -39,7 +39,7 @@ func TestUpdateCommand(t *testing.T) {
 				appSelectMock.On("TeamAppSelectPrompt").Return(prompts.SelectedApp{App: types.App{AppID: "A123"}, Auth: types.SlackAuth{}}, nil)
 				// Set experiment flag
 				clientsMock.Config.ExperimentsFlag = append(clientsMock.Config.ExperimentsFlag, "read-only-collaborators")
-				clientsMock.Config.LoadExperiments(context.Background(), clientsMock.IO.PrintDebug)
+				clientsMock.Config.LoadExperiments(ctx, clientsMock.IO.PrintDebug)
```
mwbrooks, line 42: note: Here's another example where we're replacing a `context.Background()` with the new, mocked context `ctx`

cmd/externalauth/add_secret_test.go
```diff
@@ -264,10 +264,10 @@ func TestExternalAuthAddClientSecretCommand(t *testing.T) {
 				clientsMock.AddDefaultMocks()
 				// TODO this can probably be replaced by a helper that sets up an apps.json file in
 				// the right place on the afero memfs instance
-				err := clients.AppClient().SaveDeployed(context.Background(), fakeApp)
+				err := clients.AppClient().SaveDeployed(ctx, fakeApp)
```
mwbrooks, line 267: note: Moar switch-a-roos

internal/pkg/platform/activity_test.go
```diff
@@ -147,37 +148,37 @@ func TestPlatformActivity_StreamingLogs(t *testing.T) {
...
 				cm.ApiInterface.AssertNumberOfCalls(t, "Activity", 1)
 			},
 		},
 	} {
 		t.Run(name, func(t *testing.T) {
-			ctx := context.Background()
-			ctx = context.WithValue(ctx, config.CONTEXT_TOKEN, "sometoken")
 			// Create mocks
+			ctxMock := slackcontext.MockContext(context.Background())
+			ctxMock = context.WithValue(ctxMock, config.CONTEXT_TOKEN, "sometoken")
```
mwbrooks, line 160: note: One of the few unique changes, where the activity test modifies a value in the context.

internal/pkg/platform/activity_test.go
```diff
@@ -147,37 +148,37 @@ func TestPlatformActivity_StreamingLogs(t *testing.T) {
...
 			clientsMock := shared.NewClientsMock()
 			// Create clients that is mocked for testing
 			clients := shared.NewClientFactory(clientsMock.MockClientFactory())
 			// Setup custom per-test mocks (higher priority than default mocks)
 			if tt.Setup != nil {
-				ctx = tt.Setup(t, ctx, clientsMock)
+				ctxMock = tt.Setup(t, ctxMock, clientsMock)
```
mwbrooks, line 166: thought: I noticed that this activity test appears to have copy & pasted the `TableTestCommand` implementation, then modified it to work for an `internal/pkg/` test. Probably something to revisit in the future because it could be simplified to match our other non-command table tests.

test/testutil/commandtests.go
```diff
@@ -105,7 +105,7 @@ func TableTestCommand(t *testing.T, commandTests CommandTests, newCommandFunc fu
 
 			// Assert mocks or other custom assertions
 			if tt.ExpectedAsserts != nil {
-				tt.ExpectedAsserts(t, clientsMock)
+				tt.ExpectedAsserts(t, ctxMock, clientsMock)
```
mwbrooks, line 108: note: Here's the nuts-and-bolts of this PR. We just pass the mock context into `ExpectedAsserts` so that the same context can be used everywhere. 🌟

mwbrooks (commented):
Notes for the patient reviewers!

cmd/app/add_test.go
zimeg: Another tedious change - thanks for rolling up the sleeves - but this matches actual command execution so much better 🙏 ✨

cmd/app/link_test.go
zimeg: Replacing these instances with the shared mock `ctx` IMO is a huge win for all of our tests 😭

internal/pkg/platform/activity_test.go
zimeg: 🎁 Super intersting find @mwbrooks! I agree we should revisit these tests at some point since it's also surprising to find `context` changes here.

Untangling that might mean leaning more into passing `app` and `auth` as arguments but I don't recall the underhappenings of this package right now! It's no blocker to the changes otherwise either!

zimeg (approved):
@mwbrooks Thank you so much for calling out the gems of this PR! It's all great but this made reviewing much easier for me 📝 ✨ 

I'm somewhat upset to find a `+0.03%` improvement to test coverage but that makes sense. I was hoping for more because these changes are so meaningful to better matching commands at runtime.

These are exciting improvements to the health of the code that I'm excited to merge and include in future tests - thanks again for making these changes 🤖

mwbrooks:
> I'm somewhat upset to find a `+0.03%` improvement to test coverage but that makes sense. I was hoping for more because these changes are so meaningful to better matching commands at runtime.

Thanks for the review @zimeg and I'm glad to hear that the notes made reviewing smoother! I agree, it's tedious work with little upside. I think the main benefits are prevent errors down the road, when the values in context matter. Also, doing a full sweep now helps set better copy & paste standards going forward. 🫠

PR #26, opened 2025-04-10
chore: group documentation in release notes using the 'docs' label

zimeg:
### Summary

This PR updates the `MAINTAINERS_GUIDE.md` to reference the "docs" tag that upstream workflows expect when creating automagic PRs 👾 

This matches labels used in other repos such as:
- https://github.com/slackapi/bolt-js/labels
- https://github.com/slackapi/bolt-python/labels

### Requirements


lukegalbraithrussell (approved):
love to keep it casual

mwbrooks (changes requested):
📝 @zimeg Woo, thanks for keeping it chill. The shorter label is better and it's aligned with Bolt! Can we also update `release.yml` from `documentation` to `docs` so our releases detect the new label?

zimeg:
@mwbrooks Incredible callout. I'll update this now and re-review a few more things before re-requesting such review. Perhaps this was too casual a change 😉

.github/release.yml
```diff
@@ -7,6 +7,9 @@ changelog:
     - title: 🐛 Bug Fixes
       labels:
         - bug
+    - title: 📚 Documentation
+      labels:
+        - docs
```
zimeg, line 12: With the changes of #25 we should highlight changes to documentation between releases IMO!

Even if these changes are sometimes "unversioned" it can still be so interesting to follow along with these updates 👾

zimeg (commented):
📝 A note on additional changes to the meaning of this PR!

.github/release.yml
mwbrooks: ++++ totally agree! Thanks for adding this!

mwbrooks (approved):
✅ LGTM! Thanks a bunch for keeping things consistent with other repos and sprucing up our release notes 📖 ✨

zimeg:
For now I will merge this PR since followups elsewhere might still be waiting but we can revisit these notes if needed 🫡

PR #27, opened 2025-04-10
docs: fix sync issue by updating README.md logo to use GitHub URL

lukegalbraithrussell:
### Summary

This makes it an external link so we no longer need to host the image, which is creating an issue with the docs syncing. 

### Requirements


mwbrooks (approved):
✅ Thanks for fixing the sync issue. It's too bad that we can't keep it generic, but that's alright. If we update the logo, we may need to update the URL as well. 🔗

zimeg:
🤔 Do we want to follow up and remove the `.svg` from source code too? I think we can checkout one of these earlier commits as needed but I'm not sure that we're referencing it elsewhere

PR #16, opened 2025-04-07
test: assert output for collaborator list command

mwbrooks:
### Summary

This pull request is a follow-up on PR https://github.com/slackapi/slack-cli/pull/14#discussion_r2029512849 that adds tests to assert the stdout output for `collaborator` and `collaborator list`.

### Requirements

* [x] I've read and understood the [Contributing Guidelines](/blob/main/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).

PR #28, opened 2025-04-10
test: use slackcontext.MockContext(...) instead of context.Background()

mwbrooks:
### Summary

This pull request is part of the series that updates our tests to use `slackcontext.MockContext(...)` instead of `context.Background()`.

The motivation is to continue to have consistent and reliable unit tests. Right now, many of our test use empty context such as `context.Background()`, which can cause unexpected errors if a function brings to use values from the context. Now that we can mock "guaranteed" context values, this PR combs through to replace `.Background()` references.

_You might want a fresh cup of coffee ☕ to sip while scrolling through this one!_

### Requirements

### DIFF
--- a/cmd/app/link_test.go
+++ b/cmd/app/link_test.go
@@ -24,6 +24,7 @@ import (
 	"github.com/slackapi/slack-cli/internal/iostreams"
 	"github.com/slackapi/slack-cli/internal/shared"
 	"github.com/slackapi/slack-cli/internal/shared/types"
+	"github.com/slackapi/slack-cli/internal/slackcontext"
 	"github.com/slackapi/slack-cli/internal/slackdeps"
 	"github.com/slackapi/slack-cli/internal/slackerror"
 	"github.com/slackapi/slack-cli/test/testutil"
@@ -569,10 +570,8 @@ func Test_Apps_LinkAppHeaderSection(t *testing.T) {
 	}
 	for name, tt := range tests {
 		t.Run(name, func(t *testing.T) {
-			// Setup parameters for test
-			ctx := context.Background()
-
 			// Create mocks
+			ctx := slackcontext.MockContext(t.Context())
 			clientsMock := shared.NewClientsMock()
 			clientsMock.AddDefaultMocks()
 
--- a/cmd/app/uninstall_test.go
+++ b/cmd/app/uninstall_test.go
@@ -48,7 +48,7 @@ func TestAppsUninstall(t *testing.T) {
 	testutil.TableTestCommand(t, testutil.CommandTests{
 		"Successfully uninstall": {
 			Setup: func(t *testing.T, ctx context.Context, clientsMock *shared.ClientsMock, clients *shared.ClientFactory) {
-				prepareCommonUninstallMocks(clients, clientsMock)
+				prepareCommonUninstallMocks(ctx, clients, clientsMock)
 				clientsMock.ApiInterface.On("UninstallApp", mock.Anything, mock.Anything, fakeAppID, fakeAppTeamID).
 					Return(nil).Once()
 			},
@@ -58,7 +58,7 @@ func TestAppsUninstall(t *testing.T) {
 		},
 		"Successfully uninstall with a get-manifest hook error": {
 			Setup: func(t *testing.T, ctx context.Context, clientsMock *shared.ClientsMock, clients *shared.ClientFactory) {
-				prepareCommonUninstallMocks(clients, clientsMock)
+				prepareCommonUninstallMocks(ctx, clients, clientsMock)
 				clientsMock.ApiInterface.On("UninstallApp", mock.Anything, mock.Anything, fakeAppID, fakeAppTeamID).
 					Return(nil).Once()
 				manifestMock := &app.ManifestMockObject{}
@@ -73,7 +73,7 @@ func TestAppsUninstall(t *testing.T) {
 		"Fail to uninstall due to API error": {
 			ExpectedError: slackerror.New("something went wrong"),
 			Setup: func(t *testing.T, ctx context.Context, clientsMock *shared.ClientsMock, clients *shared.ClientFactory) {
-				prepareCommonUninstallMocks(clients, clientsMock)
+				prepareCommonUninstallMocks(ctx, clients, clientsMock)
 				clientsMock.ApiInterface.On("UninstallApp", mock.Anything, mock.Anything, fakeAppID, fakeAppTeamID).
 					Return(slackerror.New("something went wrong")).Once()
 			},
@@ -82,7 +82,7 @@ func TestAppsUninstall(t *testing.T) {
 			CmdArgs:       []string{},
 			ExpectedError: slackerror.New(slackerror.ErrCredentialsNotFound),
 			Setup: func(t *testing.T, ctx context.Context, cm *shared.ClientsMock, cf *shared.ClientFactory) {
-				prepareCommonUninstallMocks(cf, cm)
+				prepareCommonUninstallMocks(ctx, cf, cm)
 				appSelectMock := prompts.NewAppSelectMock()
 				uninstallAppSelectPromptFunc = appSelectMock.AppSelectPrompt
 				appSelectMock.On("AppSelectPrompt").Return(prompts.SelectedApp{App: fakeApp}, nil)
@@ -95,7 +95,7 @@ func TestAppsUninstall(t *testing.T) {
 	})
 }
 
-func prepareCommonUninstallMocks(clients *shared.ClientFactory, clientsMock *shared.ClientsMock) *shared.ClientFactory {
+func prepareCommonUninstallMocks(ctx context.Context, clients *shared.ClientFactory, clientsMock *shared.ClientsMock) *shared.ClientFactory {
 
 	// Mock App Selection
 	appSelectMock := prompts.NewAppSelectMock()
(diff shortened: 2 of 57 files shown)


cmd/app/link_test.go
```diff
@@ -569,10 +570,8 @@ func Test_Apps_LinkAppHeaderSection(t *testing.T) {
 	}
 	for name, tt := range tests {
 		t.Run(name, func(t *testing.T) {
-			// Setup parameters for test
-			ctx := context.Background()
-
 			// Create mocks
+			ctx := slackcontext.MockContext(t.Context())
 			clientsMock := shared.NewClientsMock()
```
mwbrooks, line 575: note: This is the standard, repetitive change that you'll see. We are swapping `context.Background()` for `slackcontext.MockContext(t.Context)`. 👯

cmd/app/uninstall_test.go
```diff
@@ -48,7 +48,7 @@ func TestAppsUninstall(t *testing.T) {
 	testutil.TableTestCommand(t, testutil.CommandTests{
 		"Successfully uninstall": {
 			Setup: func(t *testing.T, ctx context.Context, clientsMock *shared.ClientsMock, clients *shared.ClientFactory) {
-				prepareCommonUninstallMocks(clients, clientsMock)
+				prepareCommonUninstallMocks(ctx, clients, clientsMock)
```
mwbrooks, line 51: note: This was missed in the PR that updated `.Setup: func(t, ctx, ...)` to accept the mocked context.

cmd/collaborators/collaborators_test.go
```diff
@@ -132,3 +134,30 @@ func TestCollaboratorsCommand(t *testing.T) {
 		})
 	}
 }
+
+func TestCollaboratorsCommand_PrintSuccess(t *testing.T) {
```
mwbrooks: thought: I'm a little baffled where this test came from 🤔 I didn't write it, so I assume I updated it. But I cannot find the original copy of it! Either way, it tests `printSuccess` and I think we want to keep it. I'm assuming it was introduced by a rebase (kinda regretting doing a rebase instead of a merge).

cmd/function/access_test.go
```diff
@@ -326,6 +327,7 @@ func TestFunctionDistributionCommand_PermissionsFile(t *testing.T) {
 
 	for name, tt := range tests {
 		t.Run(name, func(t *testing.T) {
+			ctx := slackcontext.MockContext(t.Context())
```
mwbrooks, line 330: thought: It's generally best to declare the `ctx` at the top of the function, so that the same context is used throughout.

cmd/triggers/generate_test.go
```diff
@@ -54,7 +55,7 @@ func Test_TriggerGenerate_accept_prompt(t *testing.T) {
 	}
 
 	t.Run(tt.name, func(t *testing.T) {
-		ctx, clientsMock := prepareMocks(tt.triggersListResponse, tt.globResponse, tt.triggersCreateResponse, nil /*trigger create error*/)
+		ctx, clientsMock := prepareMocks(t, tt.triggersListResponse, tt.globResponse, tt.triggersCreateResponse, nil /*trigger create error*/)
```
mwbrooks, line 58: note: We need `t *testing.T` in order to called `slackcontext.MockContext(t.Context())`

internal/pkg/manifest/validate_test.go
```diff
@@ -24,13 +24,14 @@ import (
...
 	"github.com/slackapi/slack-cli/internal/shared/types"
+	"github.com/slackapi/slack-cli/internal/slackcontext"
 	"github.com/slackapi/slack-cli/internal/slackerror"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
 )
 
 func Test_ManifestValidate_GetManifestLocal_Error(t *testing.T) {
-	ctx, clients, _, log, appMock, authMock := setupCommonMocks()
+	ctx, clients, _, log, appMock, authMock := setupCommonMocks(t)
```
mwbrooks, line 34: note: It's not obvious that this change adds a `t *testing.T` argument to `setupCommonMocks(t)`. Same reason as before, so that we can call `slackcontext.MockContext(t.Context())`.

internal/prompts/app_select_test.go
```diff
@@ -573,9 +575,8 @@ func Test_FilterAuthsByToken_Flags(t *testing.T) {
 //
 
 func TestPrompt_AppSelectPrompt_SelectedAuthExpired_UserReAuthenticates(t *testing.T) {
-
```
mwbrooks, line 576: vent: Personal pet-peeve is newlines at the top of a function block in Golang 😆

mwbrooks (commented):
Comments to (hopefully) make reviewing this a little more productive and pleasant 😃

cmd/collaborators/collaborators_test.go
zimeg: This does seem to happen with branches rebased but a quick thanks for making reviews much more pleasant with this 🙏

I do think we can delete these tests to make later updates a bit more clear. IIRC #16 added tests that cover these cases instead of a separate unit here- 🔍

cmd/app/link_test.go
zimeg: Setting up `ctx` with other mocks is an awesome change!

zimeg: Using the `testing` context is a neat update also! I haven't explore enough of that package, but I imagine more useful methods might be used elsewhere too.

cmd/doctor/check_test.go
```diff
@@ -34,10 +34,10 @@ import (
 
 func TestDoctorCheckOS(t *testing.T) {
 	t.Run("returns the operating system version", func(t *testing.T) {
+		ctx := slackcontext.MockContext(t.Context())
 		clientsMock := shared.NewClientsMock()
 		clientsMock.AddDefaultMocks()
 		clients := shared.NewClientFactory(clientsMock.MockClientFactory())
-		ctx := context.Background()
```
zimeg, line 40: 👏 Creating the `ctx` before other mocks is a nice change.

cmd/function/access_test.go
zimeg: Edit: I should've read on! I agree so much that this is a good change.

cmd/triggers/create_test.go
```diff
@@ -546,13 +547,13 @@ func TestTriggersCreateCommand_promptShouldInstallAndRetry(t *testing.T) {
 
 	testcases := []struct {
 		name         string
-		prepareMocks func(*app.AppMock, *shared.ClientsMock)
+		prepareMocks func(*testing.T, context.Context, *shared.ClientsMock, *app.AppMock)
```
zimeg, line 550: 🤖 Love to see these patterns!

internal/pkg/manifest/validate_test.go
```diff
@@ -231,8 +232,9 @@ func Test_HandleConnectorApprovalRequired(t *testing.T) {
 }
 
 // Setup
-func setupCommonMocks() (ctx context.Context, clients *shared.ClientFactory, clientsMock *shared.ClientsMock, log *logger.Logger, mockApp types.App, mockAuth types.SlackAuth) {
+func setupCommonMocks(t *testing.T) (ctx context.Context, clients *shared.ClientFactory, clientsMock *shared.ClientsMock, log *logger.Logger, mockApp types.App, mockAuth types.SlackAuth) {
```
zimeg, line 235: :books: No blocker but sometimes arguments and returns of this amount on a single line are no fun to parse as the reader.

No need to change it and this is opinion too! Also I've tried using linters to tackle this with linters that `max` a line length but handling cases where this is needed is not always convenient TBH.

internal/pkg/manifest/validate_test.go
```diff
@@ -24,13 +24,14 @@ import (
 	"github.com/slackapi/slack-cli/internal/logger"
 	"github.com/slackapi/slack-cli/internal/shared"
 	"github.com/slackapi/slack-cli/internal/shared/types"
+	"github.com/slackapi/slack-cli/internal/slackcontext"
```
zimeg, line 27: 🔍 Other `import` updates also remove `context` and I was wondering if we're using that package in other ways here?

zimeg: Edit: Ahhhh of course! It's in the setup of common mocks 👁️‍🗨️

internal/prompts/app_select_test.go
zimeg: ➕ I find these make jumping around sections sometimes a hassle- Thanks for reducing our linecounts with this change!

zimeg (approved):
@mwbrooks

> Now that we can mock "guaranteed" context values, this PR combs through to replace `.Background()` references.

How great! And thanks for keeping these changes in one PR. The review was energized without chaos. It's so much appreciated.

I left one note on added tests that I think can be removed in these changes, but these changes add a lot of reason to the tests that exist already!

🧪 ✨ 👾

cmd/app/link_test.go
mwbrooks: I like using the `t.Context()` as well. A few benefits are:

1. It makes it easy to search for `context.Background()` and `context.TODO()` to prevent those from slipping into production code.
2. `t.Context()` is a cancel context and it's cancelled just before the test cleanup. So, any goroutines running and listening for `<- ctx.Done()` will have a chance to shutdown.

cmd/collaborators/collaborators_test.go
mwbrooks: Ohhhh, great memory! I'll remove this test then!

mwbrooks: Commit f1d5fa8 removes this test. I've confirmed that we still have test coverage after it's been removed. Thanks @zimeg!

internal/pkg/manifest/validate_test.go
mwbrooks: Yea, that's a fair point @zimeg! If we can use our linter to enforce a max line length before declaring the function as multi-line, then I think that's best!

PR #29, opened 2025-04-10
ci: adds workflow for syncing docs out to docusaurus repo

lukegalbraithrussell:
### Summary

This adds a workflow that takes edits to the md docs in this repo and pushes it to the docusaurus build. This will be using a reusable workflow soon, and i'll replace this then.

### Requirements


.github/workflows/sync-docs-from-cli-repo.yml
```diff
@@ -0,0 +1,74 @@
...
+    branches:
+      - main
+    paths:
+      - "docs/**" 
+  pull_request:
+    branches:
+      - main
+    paths:
+      - "docs/**" 
+  workflow_dispatch:
```
lukegalbraithrussell, line 14: the `if` statement below means it only creates the PR on push and dispatch. but it'll run everything else as a test for PRs made

.github/workflows/sync-docs-from-cli-repo.yml
```diff
@@ -0,0 +1,74 @@
...
+          commit-message: "Sync docs from ${{ github.event.repository.name }} to docs site repo"
+          base: "main"
+          branch: "docs-sync-${{ github.event.repository.name }}-${{ github.sha }}"
+          labels: docs
+          path: "./docs_repo"
+
+      - name: Output the pull request link
+        if: ${{ steps.site-pr.outputs.pull-request-url }}
+        run: |
+          echo "Pull request created: ${{ steps.site-pr.outputs.pull-request-url }}" >> $GITHUB_STEP_SUMMARY
```
zimeg, line 74:
```suggestion
```

Quick note since this is useful to have, but IIRC this step was added for collecting links of various PRs made to multiple **separate** repos at a time - often for sample apps.

Here we target a single source of truth and this might not be as needed, but of course no need to change it! This is a note 👾

zimeg (approved):
Sweet PR for keeping all things up to date 📚 ✨ 

All of this looks good to me and the cases this `runs` for are solid too. I think this is good to merge but left one comment on the last step that is no blocker.

I look forward to the reusable workflow as well but am also so excited for a syncing workflow. It's neat stuff 🤖

PR #33, opened 2025-04-11
refactor: update .Context() references for consistency

mwbrooks:
### Summary

> 🎶 _We all live in a yellow context, a yellow context, a yellow context_ 🛥️ 

This pull request combs through our code to update references to `.Context()` for more consistency. A few common changes are:

* Pulling out context from `someFunc(cmd.Context(), ...)` to `someFunc(ctx, ...)`
  * Verifying that `cmd.Context()` is our mocked context, which can happen from:
    * `cmd.SetContext(ctxMock)` before Cobra execution
    * `cmd.ExecuteContext(ctxMock)` after Cobra execution
* Catching a few missed places where `.Setup` provides the mocked context

### Reviewers

**Production code is modified in this PR, but I don't think we'll see any regressions or issues.**

### Requirements


cmd/app/app.go
```diff
@@ -46,8 +46,9 @@ func NewCommand(clients *shared.ClientFactory) *cobra.Command {
 			return runListCommand(cmd, clients)
 		},
 		PostRunE: func(cmd *cobra.Command, args []string) error {
+			ctx := cmd.Context()
 			if cmd.CalledAs() == "workspace" {
-				clients.IO.PrintInfo(cmd.Context(), false, fmt.Sprintf(
+				clients.IO.PrintInfo(ctx, false, fmt.Sprintf(
```
mwbrooks, line 51: note: I think this is a better pattern and I remember @zimeg has also asked that we reduce the amount of arguments that are functions calls. The main benefit is that other functions in the code block can use the same context `ctx` ensuring more consistency through the code.

cmd/app/link_test.go
```diff
@@ -61,7 +61,7 @@ func Test_Apps_Link(t *testing.T) {
 					mockLinkSlackAuth1,
 				}, nil)
 				cm.AddDefaultMocks()
-				setupAppLinkCommandMocks(t, cm, cf)
+				setupAppLinkCommandMocks(t, ctx, cm, cf)
```
mwbrooks, line 64: note: This catches a `.Setup(t *testing.T, ctx context.Context, ...)` that provides the ctx. We can now use it instead of `setupAppLinkCommandMocks(...)` creating a non-mocked context.

cmd/collaborators/update.go
```diff
@@ -74,9 +74,8 @@ func NewUpdateCommand(clients *shared.ClientFactory) *cobra.Command {
 
 // runUpdateCommand will execute the update command
 func runUpdateCommand(cmd *cobra.Command, clients *shared.ClientFactory, args []string) error {
-	var span opentracing.Span
 	ctx := cmd.Context()
-	span, ctx = opentracing.StartSpanFromContext(ctx, "cmd.Collaborators.Update")
+	span, _ := opentracing.StartSpanFromContext(ctx, "cmd.Collaborators.Update")
```
mwbrooks, line 78: thought: This is up for debate. 🧠  OpenTracing span's return a new context that contains the span (and all of our values). Right now, we inconsistently replace the current `ctx` with the span's `ctx`. I've leaned toward not replacing our `ctx` because:

* We never access the span through the `ctx` AFAIK
  * Instead, we often call `span.Context()...` to get it and modify it for API calls
* We are moving toward a rule of thumb where the context should not be modified after a Cobra Execution

cmd/root.go
```diff
@@ -204,25 +206,25 @@ func Init() (*cobra.Command, *shared.ClientFactory) {
...
+			clients.IO.PrintError(ctx, err.Error())
 			clients.Os.Exit(int(iostreams.ExitError))
 		}
 	})
 	// Since we use the *E cobra lifecycle methods, OnFinalize is one of the few ways we can ensure something _always_ runs at the end of any command invocation, regardless if an error is raised or not during execution.
 	cobra.OnFinalize(func() {
-		cleanup(rootCmd.Context(), clients)
+		ctx := rootCmd.Context()
+		cleanup(ctx, clients)
 	})
```
mwbrooks: note: This is a unique area of our code w.r.t context. 

The `cobra.OnInitialize(...)` and `cobra.OnFinalize(...)` are lifecycle functions called after `rootCmd.ExecuteContext(...)`. I think it would be a nice addition for Cobra to pass the `cmd cobra.Command` into these functions, but it doesn't. 

Since the context may change dramatically before-execution and after-execution, we can get the most up-to-date context with `rootCmd.Context()`. _I think...I haven't tested what happens when you set sub-command sets the context, whether it rolls up to the root command._

mwbrooks (commented):
🦮 Comments to guide the reviewer safely toward the end!

cmd/collaborators/update.go
zimeg: @mwbrooks Thanks for calling this out. How we use `span` in `context` is interesting to me.

We might consider follow ups for this, but IIRC spans are most useful for collecting [traces](https://opentelemetry.io/docs/concepts/signals/traces/) that measure the beginning and end of function calls? I'm not sure if these can be gathered now, but we might find tools to [visualize](https://github.com/CtrlSpice/otel-desktop-viewer) these useful in debugging slow processes 🔍

Not updating this `ctx` might cause strangeness in the trace callstack and make for confusing debugs if we explore this, but I have not tried it at all!

No blocker for this PR since we might consider this a standalone task altogether, but I'm also curious if this isn't seeming like the right use of `context` to you 🤔

zimeg: Making this more consistent overall is a nice change in this PR too!

cmd/app/app.go
zimeg: I noticed too you're updating it to be a first declaration in a function, which I'm so much a fan of too 🙏 ✨ 

In later comments you note that we often avoid modifying `ctx` as well and I think this makes it more clear that it's meant to be more of a function argument and not a variable to change 😉

cmd/root.go
zimeg: 🤔 This is so interesting too!

I'm still thinking about the `span` example and updating `context` throughout the function callstack and it might make sense that values set in "lower" levels of `context` aren't shared "above"? At least if the `ctx` is not returned...

> `context` package that makes it easy to pass request-scoped values

Function calls might be similar to "requests" in this context 🔗 https://go.dev/blog/context

I agree that this is a [confusing](https://go.dev/blog/context-and-structs) area overall though.

It's not clear to me how [immutable](https://pkg.go.dev/context#WithValue) values and pointers pair sometimes... But I am finding that we're setting the `context` as expected and I'm not finding that `cobra` modifies this:

https://github.com/slackapi/slack-cli/blob/b76e933310a8e676f6c1dc07002b0c384c018922/cmd/root.go#L371

To me this seems like a fine change, but I'm also curious if we can pass `ctx` to `Init` instead?

FWIW the current implementation seems to work when gathering context values from within a command `RunE` handler 👾

internal/config/project_test.go
```diff
@@ -619,7 +619,7 @@ func Test_Config_CreateProjectConfigJSONFile(t *testing.T) {
 	}
 	for name, tt := range tests {
 		t.Run(name, func(t *testing.T) {
-			ctx := t.Context()
+			ctx := slackcontext.MockContext(t.Context())
```
zimeg, line 622: 👁️‍🗨️

internal/slacktrace/slacktrace.go
```diff
@@ -35,7 +35,7 @@ package slacktrace
 // To add trace output to a command:
 //
 // use,
-// clients.IO.PrintTrace(cmd.Context(), slacktrace.<NAMEDCONST>)
+// clients.IO.PrintTrace(ctx, slacktrace.<NAMEDCONST>)
```
zimeg, line 38: 😌 This is so useful for encouraging shared styles

zimeg (approved):
@mwbrooks LGTM and swimmingly so  🫧 🐟

This PR has super good improvements to code health and asks a few great questions. Thanks a ton for making comments about these.

I responded with a few ideas about how we might want to use `ctx` throughout command lifecycles and with spans but these might both be explorations for follow up PRs. But no blocking notes within 🎶 🚢 💨

cmd/collaborators/update.go
mwbrooks: Yea, you bring up a good point about the `span`'s forming a track callback. I'd expect that we'd want to update the context in that case.

Since we don't lean too heavily on the traces, I'll keep this PR as-is. The intention is to make everything consistent (not updating the context). Then, like you said, we can have a follow-up PR that updates everything together.

cmd/root.go
mwbrooks: Good points! Given how we use `ctx` in the `OnInitialize` and `OnFinalize`, I think we could safely use the `ctx` passed to `Init`. The upside is that the code is cleaner and easier to read.

Happy to make that change now before merging this PR!

mwbrooks: Commit 9f25031 passes `ctx` into `Init(ctx)` and uses it for `OnInitialize` and `OnFinalize`. All the magic is removed! 🪄 🙅🏻

mwbrooks:
Thanks for the thorough review @zimeg and suggesting the change to `Init(ctx)` 🙇🏻 Merging! :shipit:

PR #47, opened 2025-04-17
refactor: remove context.Background() references from triggers command

mwbrooks:
### Summary

_Context delivery, get it while it's hot!_ 💁🍕 

This pull request updates the `trigger` command to no longer use `context.Background()`. Not too exciting, but I've separated this into a single PR because its changes span a few files

### Requirements

### DIFF
--- a/cmd/triggers/create.go
+++ b/cmd/triggers/create.go
@@ -116,7 +116,7 @@ func runCreateCommand(clients *shared.ClientFactory, cmd *cobra.Command) error {
 		app = _app
 	}
 
-	err = validateCreateCmdFlags(clients, &createFlags)
+	err = validateCreateCmdFlags(ctx, clients, &createFlags)
 	if err != nil {
 		return err
 	}
--- a/cmd/triggers/generate.go
+++ b/cmd/triggers/generate.go
@@ -78,7 +78,7 @@ func TriggerGenerate(ctx context.Context, clients *shared.ClientFactory, app typ
 		},
 	})))
 
-	triggerFilePaths, err := getFullyQualifiedTriggerFilePaths(clients, triggerPaths)
+	triggerFilePaths, err := getFullyQualifiedTriggerFilePaths(ctx, clients, triggerPaths)
 	if err != nil {
 		return nil, err
 	}
@@ -171,8 +171,7 @@ func getTriggerPaths(sdkCLIConfig *hooks.SDKCLIConfig) []string {
 
 // getFullyQualifiedTriggerFilePaths returns an array of file paths that
 // are validated
-func getFullyQualifiedTriggerFilePaths(clients *shared.ClientFactory, triggerPaths []string) ([]string, error) {
-	ctx := context.Background()
+func getFullyQualifiedTriggerFilePaths(ctx context.Context, clients *shared.ClientFactory, triggerPaths []string) ([]string, error) {
 	projectDir, err := clients.Os.Getwd()
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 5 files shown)


cmd/triggers/create.go
```diff
@@ -116,7 +116,7 @@ func runCreateCommand(clients *shared.ClientFactory, cmd *cobra.Command) error {
 		app = _app
 	}
 
-	err = validateCreateCmdFlags(clients, &createFlags)
+	err = validateCreateCmdFlags(ctx, clients, &createFlags)
```
mwbrooks, line 119: note: For the most part, this PR updates 2 functions to accept `ctx context.Context`. So, you'll see a lot of changes that just add `ctx` as an argument.

cmd/triggers/generate.go
```diff
@@ -171,8 +171,7 @@ func getTriggerPaths(sdkCLIConfig *hooks.SDKCLIConfig) []string {
 
 // getFullyQualifiedTriggerFilePaths returns an array of file paths that
 // are validated
-func getFullyQualifiedTriggerFilePaths(clients *shared.ClientFactory, triggerPaths []string) ([]string, error) {
-	ctx := context.Background()
```
mwbrooks, line 175: note: This is 1st key change, where we remove `ctx := context.Background()` and replace wit with a `ctx context.Context` function argument.

cmd/triggers/generate.go
```diff
@@ -290,16 +289,15 @@ func validateCreateCmdFlags(clients *shared.ClientFactory, createFlags *createCm
 	return nil
 }
 
-func maybeSetTriggerDefFlag(clients *shared.ClientFactory, createFlags *createCmdFlags) error {
-	ctx := context.Background()
```
mwbrooks, line 294: note: This is the 2nd key change, where we also replace `ctx := context.Background()` with the `ctx context.Context` function argument.

mwbrooks (commented):
A few comments to help guide the reviewer 🥾

cmd/triggers/generate_test.go
```diff
@@ -372,6 +372,7 @@ func Test_TriggerGenerate_MismatchedFlags(t *testing.T) {
 
 	for name, tt := range tests {
 		t.Run(name, func(t *testing.T) {
+			ctx := slackcontext.MockContext(t.Context())
```
zimeg, line 375: 🔍 This is good.

cmd/triggers/generate.go
zimeg: 🗣️ This is a ramble, but I'm wondering if we want to explore using `ctx` to timeout certain prompts?

IIRC confusion has appeared when inputs aren't obvious, and a timeout with error might be clear that something was expected. No blocker for this PR though!

zimeg (approved):
@mwbrooks LGTM and it's nice to find related changes in one set of changes.

All of the updates and even a line removed are solid. I did leave one comment unrelated to these chagnes that `ctx` has me excited to consider. These improvements I think are making changes like that even possible - cheers aplenty! 🎉

cmd/triggers/generate.go
mwbrooks: We can talk about it more, but it may be a fine line between knowing when someone is taking their time thinking vs unaware of the prompt and waiting.

cmd/triggers/generate_test.go
mwbrooks: I'm really enjoying the `slackcontext.MockContext()` function. It's been very handy to consistently sprinkle a mocked context into our tests.

mwbrooks:
Thanks for the quick review @zimeg! I imagine these context PRs are getting a little stale to read through 😆

PR #56, opened 2025-04-21
feat: 'app link' supports run-on-slack apps with local manifest source

mwbrooks:
### Summary

This pull request updates the **`app link`** command to **work with run-on-Slack apps** that have **a local manifest source**.

Perhaps, this as the beginning of loosening up the `app link` command to not always require a remote manifest source. 

This work came about from the GitHub Issue https://github.com/slackapi/deno-slack-sdk/issues/451. In the issue, I wasn't able to suggest using the `app link` command because it would force the Deno SDK app to have a remote manifest. This seemed silly, since all Deno SDK apps have a local manifest. Since developer can now see their Deno SDK apps in https://api.slack.com/apps, we should allow them to link the app in their project.

### Preview

https://github.com/user-attachments/assets/a304ac0c-29c8-4050-bcc3-b5c803ff5fc8

https://github.com/user-attachments/assets/a9aa8a44-6b1c-4b62-92ab-9a9486376784

### Reviewers

```bash
# Test Deno SDK
$ lack create asdf/ -t slack-samples/deno-starter-template
$ cd asdf/
$ lack app link
# → Confirm NO manifest source prompt
# → CTRL+C (or, extra mile if you want to add an App ID and confirm it's still a local manifest)
# Update manifest source to be remote
$ vim .slack/config.json
# → Change "local" to "remote"
$ lack app link
# → Confirm NO manifest source prompt
# → CTRL+C
$ cd ../
$ rm -rf asdf/

# Test Bolt for JavaScript
$ lack create asdf/ -t slack-samples/bolt-js-custom-function-template
$ cd asdf/
$ lack app link
# → Confirm manifest source prompt
# → CTRL+C
# Update manifest source to be remote
$ vim .slack/config.json
# → Change "local" to "remote"
$ lack app link
# → Confirm NO manifest source prompt
# → CTRL+C
$ cd ../
$ rm -rf asdf/
```

### Requirements

### DIFF
--- a/cmd/app/link.go
+++ b/cmd/app/link.go
@@ -115,7 +115,7 @@ func LinkCommandRunE(ctx context.Context, clients *shared.ClientFactory, app *ty
 // explaining how to link apps, in case the user declines.
 func LinkAppHeaderSection(ctx context.Context, clients *shared.ClientFactory, shouldConfirm bool) {
 	var secondaryText = []string{
-		"Add an existing app created on app settings",
+		"Add an existing app from app settings",
 		"Find your existing apps at: " + style.Underline("https://api.slack.com/apps"),
 	}
 
@@ -156,10 +156,12 @@ func LinkExistingApp(ctx context.Context, clients *shared.ClientFactory, app *ty
 		}
 	}
 
-	// Confirm to update manifest source to remote
+	// Confirm to update manifest source to remote.
+	// - Update the manifest source to remote when its a GBP project with a local manifest.
+	// - Do not update manifest source for ROSI projects, because they can only be local manifests.
 	manifestSource, err := clients.Config.ProjectConfig.GetManifestSource(ctx)
 
-	if !manifestSource.Equals(config.MANIFEST_SOURCE_REMOTE) || err != nil {
+	if err != nil || (!manifestSource.Equals(config.MANIFEST_SOURCE_REMOTE) && cmdutil.IsSlackHostedProject(ctx, clients) != nil) {
 		// When undefined, the default is config.MANIFEST_SOURCE_LOCAL
 		if !manifestSource.Exists() {
 			manifestSource = config.MANIFEST_SOURCE_LOCAL
--- a/cmd/app/link_test.go
+++ b/cmd/app/link_test.go
@@ -20,6 +20,7 @@ import (
 	"testing"
 
 	"github.com/slackapi/slack-cli/internal/api"
+	"github.com/slackapi/slack-cli/internal/app"
 	"github.com/slackapi/slack-cli/internal/config"
 	"github.com/slackapi/slack-cli/internal/iostreams"
 	"github.com/slackapi/slack-cli/internal/shared"
@@ -430,7 +431,7 @@ func Test_Apps_Link(t *testing.T) {
 			CmdArgs:       []string{},
 			ExpectedError: slackerror.New(slackerror.ErrAppNotFound),
 		},
-		"accept manifest source prompt and saves information about the provided deployed app": {
+		"accepting manifest source prompt should save information about the provided deployed app": {
 			Setup: func(t *testing.T, ctx context.Context, cm *shared.ClientsMock, cf *shared.ClientFactory) {
 				cm.AuthInterface.On("Auths", mock.Anything).Return([]types.SlackAuth{
 					mockLinkSlackAuth2,
@@ -503,7 +504,7 @@ func Test_Apps_Link(t *testing.T) {
 				)
 			},
 		},
-		"decline manifest source prompt should not link app": {
+		"declining manifest source prompt should not link app": {
 			Setup: func(t *testing.T, ctx context.Context, cm *shared.ClientsMock, cf *shared.ClientFactory) {
 				cm.AddDefaultMocks()
 				setupAppLinkCommandMocks(t, ctx, cm, cf)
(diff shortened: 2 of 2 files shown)


cmd/app/link.go
```diff
@@ -115,7 +115,7 @@ func LinkCommandRunE(ctx context.Context, clients *shared.ClientFactory, app *ty
 // explaining how to link apps, in case the user declines.
 func LinkAppHeaderSection(ctx context.Context, clients *shared.ClientFactory, shouldConfirm bool) {
 	var secondaryText = []string{
-		"Add an existing app created on app settings",
+		"Add an existing app from app settings",
```
mwbrooks, line 118: note: App Settings now displays Slack CLI created apps as read-only, un-clickable. So, I figured we should update this text to not be restricted to "created on app settings" because the app may be created by the CLI.

cmd/app/link.go
```diff
@@ -156,10 +156,12 @@ func LinkExistingApp(ctx context.Context, clients *shared.ClientFactory, app *ty
...
 	}
 
-	// Confirm to update manifest source to remote
+	// Confirm to update manifest source to remote.
+	// - Update the manifest source to remote when its a GBP project with a local manifest.
+	// - Do not update manifest source for ROSI projects, because they can only be local manifests.
 	manifestSource, err := clients.Config.ProjectConfig.GetManifestSource(ctx)
 
-	if !manifestSource.Equals(config.MANIFEST_SOURCE_REMOTE) || err != nil {
+	if err != nil || (!manifestSource.Equals(config.MANIFEST_SOURCE_REMOTE) && cmdutil.IsSlackHostedProject(ctx, clients) != nil) {
```
mwbrooks: note: This is the main change. We only want to display the manifest source update prompt when the manifest is not already remote AND it's not a ROSI app.

cmd/app/link_test.go
```diff
@@ -430,7 +431,7 @@ func Test_Apps_Link(t *testing.T) {
 			CmdArgs:       []string{},
 			ExpectedError: slackerror.New(slackerror.ErrAppNotFound),
 		},
-		"accept manifest source prompt and saves information about the provided deployed app": {
+		"accepting manifest source prompt should save information about the provided deployed app": {
```
mwbrooks, line 434: note: Just re-wording after reading over the test cases for clarity.

cmd/app/link_test.go
```diff
@@ -537,6 +538,159 @@ func Test_Apps_Link(t *testing.T) {
 				require.Len(t, apps, 0)
 			},
 		},
+		"manifest source prompt should not display for Run-on-Slack apps with local manifest source": {
```
mwbrooks, line 541: note: New test case to assert that the manifest update prompt is NOT displayed for ROSI apps with a local manifest. I don't check for remote manifests, because that's handled in an existing test-case.

cmd/app/link_test.go
```diff
@@ -537,6 +538,159 @@ func Test_Apps_Link(t *testing.T) {
...
+				assert.Equal(t, expectedApp, actualApp)
+				// Assert manifest confirmation prompt was not displayed
+				cm.IO.AssertNotCalled(t, "ConfirmPrompt",
+					mock.Anything,
+					LinkAppManifestSourceConfirmPromptText,
+					mock.Anything,
+				)
+			},
+		},
+		"manifest source prompt should display for GBP apps with local manifest source": {
```
mwbrooks, line 618: note: Added a test-case to assert that the manifest update prompt is still displayed for GBP apps with a local manifest. This case might be redundant, but I felt high confidence after adding it.

cmd/app/link_test.go
```diff
@@ -609,4 +763,9 @@ func setupAppLinkCommandMocks(t *testing.T, ctx context.Context, cm *shared.Clie
 	if err := cm.Config.ProjectConfig.SetManifestSource(ctx, config.MANIFEST_SOURCE_REMOTE); err != nil {
 		require.FailNow(t, fmt.Sprintf("Failed to set the manifest source in the memory-based file system: %s", err))
 	}
+
+	// Mock manifest
+	manifestMock := &app.ManifestMockObject{}
+	manifestMock.On("GetManifestLocal", mock.Anything, mock.Anything).Return(types.SlackYaml{}, nil)
+	cf.AppClient().Manifest = manifestMock
```
mwbrooks: note: Just a little reflection on how we don't actually have a nice way to mock the Manifest client atm. We have a `ManifestMockObject` but it's not automatically setup in the clients mock.

mwbrooks (commented):
✏️ A few notes for the reviewer, so the test-cases don't look overwhelming!

cmd/app/link_test.go
zimeg:
```
: {
```

cmd/app/link.go
zimeg: Nit: I am perhaps low on brainpower right now, but this is not so obvious to me...

At the expense of an extra line or two, what do you think about a separate variable for:

```go
isSlackHostedProject := cmdutil.IsSlackHostedProject(ctx, clients) == nil
```

cmd/app/link.go
zimeg: ⭐ What an eye for good readings!

cmd/app/link_test.go
zimeg: Thanks for adding it! I had a good feeling when running the `dev` build and reading code, but this will be so helpful in guarding against unexpected updates to these prompts.

The `assert` prompting at the end is very very good for this 🏆 ✨

cmd/app/link_test.go
```diff
@@ -609,4 +763,9 @@ func setupAppLinkCommandMocks(t *testing.T, ctx context.Context, cm *shared.Clie
 	if err := cm.Config.ProjectConfig.SetManifestSource(ctx, config.MANIFEST_SOURCE_REMOTE); err != nil {
 		require.FailNow(t, fmt.Sprintf("Failed to set the manifest source in the memory-based file system: %s", err))
 	}
+
+	// Mock manifest
+	manifestMock := &app.ManifestMockObject{}
+	manifestMock.On("GetManifestLocal", mock.Anything, mock.Anything, mock.Anything).Return(types.SlackYaml{}, nil)
+	cf.AppClient().Manifest = manifestMock
```
zimeg, line 770: 📝 Perhaps personal preference, but I'm curious what you think about:

Having standalone and shared "setup"..."mock" functions I find to be confusing, even if convenient. IMO inlining mocks is super helpful in table tests to keep each unit separate.

No blocker at all for this PR, but does this seem like a longterm idea we should nudge towards?

zimeg (approved):
@mwbrooks Thanks tons for tackling the initial issue with careful resolution and also finding a fix for future encounters! 💌

I got a bit stumped on logic within, but all testing is working well so marking this as **approved** 🚢 💨

In following comments I left a note on the confusion I had and another thought about testing. Please feel free to ignore and merge or respond if that feels right 👾

cmd/app/link.go
mwbrooks: You and I were both low on brain power. I originally separated it to a variable in order to make sense of the conditional. Then I combined it afterwards. I'm happy to separate it out and I may do the same with the remote manifest conditional because the `!` and `.Equals` take a moment to make sense.

mwbrooks: Commt 30e69a7 takes your suggestion to heart and it's so much better!

```golang
isManifestSourceRemote := manifestSource.Equals(config.MANIFEST_SOURCE_REMOTE)
isSlackHostedProject := cmdutil.IsSlackHostedProject(ctx, clients) == nil

if err != nil || (!isManifestSourceRemote && !isSlackHostedProject) {
```

cmd/app/link_test.go
mwbrooks:
```
: }
```

cmd/app/link_test.go
mwbrooks: 🧠 I agree with you. I also find the `.DefaultMocks()` method confusing. Sometimes it can be very convenient because I just need a standard happy path mock, but often tests don't require it. The same can be true for the "setup" helpers, where they become overloaded with defaults that can actually drift some of the tests away from their target.

It would be nice to settle on some practice that keeps the tests short(er) while explicit.

mwbrooks:
Thanks for the quick review @zimeg! I totally agree with your "nit" comment to simplify the conditional logic. Thanks a bunch for calling that out and commit https://github.com/slackapi/slack-cli/commit/30e69a79f90cf43bd913cc046054ce22a443f91b tightened it up! 🔩

PR #60, opened 2025-04-23
fix: edge-case where cancelled context causes process to hang

mwbrooks:
### Summary

This pull request fixes a bug in our root command's execution (`.ExecuteContext(...)`) that could cause the Slack CLI to hang indefinitely. However, AFAIK, it's very unlikely for the Slack CLI to hit the edge case to cause the indefinite hang.

The core problem is that `<-ctx.Done()` was not handled, so any cancelled context would cause the Slack CLI to hang waiting for a channel completion or interrupt signal event.

### CHANGELOG

> Fixes a rare edge-case that causes the Slack CLI to hang without exiting the process. Now, the Slack CLI will exit with the correct error code and allow sub-processes to cleanup.

### Reviewers

We don't have tests for this area of the code. I started to add tests, but it requires more refactoring that led to a messy PR. So, I'll save that work for a follow-up PR.

If you want to manually test, then you will need to make a code change that cancels the context causing the Slack CLI to handle the context cancelled channel event.

**cmd/root.go:L373-L374:**
```diff
	clients.IO.PrintDebug(ctx, "Got second process interrupt signal, exiting the process")
	os.Exit(int(iostreams.ExitCancel))
}()
+ cancel()
// The cleanup() method in the root command will invoke via `defer` from within Execute.
if err := rootCmd.ExecuteContext(ctx); err != nil {
```

**Steps for the Reviewer:**
```bash
# Apply the code changes above
$ vim cmd/root.go

# Build the CLI
$ make build
# → It will FAIL because the Slack CLI exits with a non-zero error code from the version command

$ lack version --verbose
[2025-04-22 20:53:02] Got a cancelled context, sending interrupt signal
[2025-04-22 20:53:02] Got second process interrupt signal, exiting the process
# → Confirm the above verbose output
```

### Requirements

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -332,15 +332,18 @@ func ExecuteContext(ctx context.Context, rootCmd *cobra.Command, clients *shared
 	completedChan := make(chan bool, 1)      // completed is used for signalling an end to command
 	exitChan := make(chan bool, 1)           // exit blocks the command from exiting until completed
 	interruptChan := make(chan os.Signal, 1) // interrupt catches signals to avoid abrupt exits
+
 	signal.Notify(interruptChan, os.Interrupt, syscall.SIGTERM)
 	defer func() {
 		signal.Stop(interruptChan)
 		cancel()
 	}()
+
+	// Wait for either an interrupt signal (ctrl+c) or an explicit call to ctx.cancel()
 	go func() {
-		// Wait for either an interrupt signal (ctrl+c), or an explicit call to ctx.cancel()
 		select {
-		case <-interruptChan: // first interrupt signal, cancel context
+		// Received interrupt signal, so cancel the context and cleanup
+		case <-interruptChan:
 			clients.IO.PrintInfo(ctx, false, "\n") // flush the CTRL + C character
 			clients.IO.PrintDebug(ctx, "Got process interrupt signal, cancelling context")
 			cancel()
@@ -355,11 +358,12 @@ func ExecuteContext(ctx context.Context, rootCmd *cobra.Command, clients *shared
 				clients.IO.PrintDebug(ctx, "Exiting with cancel exit code.")
 				os.Exit(int(iostreams.ExitCancel))
 			}()
+		// Received cancelled context, so send an interrupt signal
 		case <-ctx.Done():
-			// No interrupt signal sent, command executed to completion
-			// FIXME - `.Done()` channel is triggered by `cancel()` and not a successfully completion
+			clients.IO.PrintDebug(ctx, "Got a cancelled context, sending interrupt signal")
+			interruptChan <- os.Interrupt
+		// Received completed execution, so exit the process successfully
 		case <-completedChan:
-			// No canceled context, but command has completed execution
 			exitChan <- true
 		}
 		// If we get a second interrupt, no matter what exit the process
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -329,9 +329,9 @@ func InitConfig(ctx context.Context, clients *shared.ClientFactory, rootCmd *cob
 func ExecuteContext(ctx context.Context, rootCmd *cobra.Command, clients *shared.ClientFactory) {
 	// Derive a cancel context that is cancelled when the main execution is interrupted or cleaned up.
 	// Sub-commands can register for the cleanup wait group with clients.CleanupWaitGroup.Add(1)
-	// and listen for <-ctx.Done() to be notified when the main execution is terminated and have
-	// a chance to cleanup. This is useful for long running processes and background goroutines,
-	// the activity and upgrade commands.
+	// and listen for <-ctx.Done() to be notified when the main execution is interrupted, in order
+	// to have a chance to cleanup. This is useful for long running processes and background goroutines,
+	// such as the activity and upgrade commands.
 	ctx, cancel := context.WithCancel(ctx)
 
 	completedChan := make(chan bool, 1)      // completed is used for signalling an end to command
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -327,6 +327,11 @@ func InitConfig(ctx context.Context, clients *shared.ClientFactory, rootCmd *cob
 // It listens for process interrupts and sends to IOStreams' GetInterruptChannel() for use in
 // in communicating process interrupts elsewhere in the code.
 func ExecuteContext(ctx context.Context, rootCmd *cobra.Command, clients *shared.ClientFactory) {
+	// Derive a cancel context that is cancelled when the main execution is interrupted or cleaned up.
+	// Sub-commands can register for the cleanup wait group with clients.CleanupWaitGroup.Add(1)
+	// and listen for <-ctx.Done() to be notified when the main execution is terminated and have
+	// a chance to cleanup. This is useful for long running processes and background goroutines,
+	// the activity and upgrade commands.
 	ctx, cancel := context.WithCancel(ctx)
 
 	completedChan := make(chan bool, 1)      // completed is used for signalling an end to command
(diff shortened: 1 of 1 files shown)


cmd/root.go
```diff
@@ -332,15 +332,18 @@ func ExecuteContext(ctx context.Context, rootCmd *cobra.Command, clients *shared
...
+
 	signal.Notify(interruptChan, os.Interrupt, syscall.SIGTERM)
 	defer func() {
 		signal.Stop(interruptChan)
 		cancel()
 	}()
+
+	// Wait for either an interrupt signal (ctrl+c) or an explicit call to ctx.cancel()
 	go func() {
-		// Wait for either an interrupt signal (ctrl+c), or an explicit call to ctx.cancel()
```
mwbrooks, line 341: note: I moved this comment to be above the block that it described. I did this for a few of the case statements below as well. 📝

cmd/root.go
```diff
@@ -355,11 +358,12 @@ func ExecuteContext(ctx context.Context, rootCmd *cobra.Command, clients *shared
 				clients.IO.PrintDebug(ctx, "Exiting with cancel exit code.")
 				os.Exit(int(iostreams.ExitCancel))
 			}()
+		// Received cancelled context, so send an interrupt signal
 		case <-ctx.Done():
-			// No interrupt signal sent, command executed to completion
-			// FIXME - `.Done()` channel is triggered by `cancel()` and not a successfully completion
+			clients.IO.PrintDebug(ctx, "Got a cancelled context, sending interrupt signal")
+			interruptChan <- os.Interrupt
```
mwbrooks: note: 🍞 🧈 Here's the bread & butter of the PR. 

The `<-ctx.Done()` channel is triggered when the context is cancelled with `cancel()`. Previously, this was not handled. So unless the `<-completedChan` or `<-interruptChan` were triggered, the CLI would hang - although this was very unlikely to happen.

Now, when the context is cancelled, we send an `<-interruptChan`. This seems reasonable, since the context will be cancelled for an unexpected reason, so the interrupt signal matches well (I think).

FWIW, the only time that the context was cancelled was:

1. Part of the cleanup from handling the interrupt signal
2. Deferred function when `ExecuteContext(...)` returns

In both of these cases, either `completedChan` or `interruptChan` should have been triggered.

mwbrooks (commented):
🧠 ⚡ 🧠 Leaving some knowledge sharing for the reviewers!

cmd/root.go
zimeg:
```suggestion
			clients.IO.PrintDebug(ctx, "Got a cancelled context, exiting the process")
			exitChan <- true
```

🤔 I'm not sure if we want to treat a canceled context as interrupts here since this will exit with an unexpected error code:

```
$ echo $?
130
```

zimeg:
📝 Another interesting finding was the [`signal.NotifyContext`](https://pkg.go.dev/os/signal#NotifyContext) function that we might consider using, though this doesn't allow an **forced** exit with two interrupts so easily...

zimeg:
📝 Sharing another article as I close out a few tabs for now:

https://pace.dev/blog/2020/02/17/repond-to-ctrl-c-interrupt-signals-gracefully-with-context-in-golang-by-mat-ryer.html

mwbrooks:
> I've been going in circles on this `root` logic and I'm not super confident around the use of `cancel` here...

Thanks for the review and insightful perspective @zimeg 🙇🏻 

For those reading along, @zimeg and I had a pairing session to dig through and unravel what's going on with the `root` logic. 

**Main Takeaways:**

**The 1st takeaway** that we discovered is that the `select` statement that listens for the various channels can only handle 1 case statement. It sounds silly, but we're use to seeing this pattern in a loop and it's how we were both thinking about the flow of the code. Since it's not in a loop, the `select` statement can only handle 1 case statement before breaking out.
* When the `case <- ctx.Done()` is handled, it's impossible for the `case <-interruptChan` to be handle.
  * So, the original fix in this PR didn't actually work as expected.
  * The `interruptChan <- os.Interrupt` actually triggered the 2nd `<- interruptChan` that was listening for a double CTRL+C press.
* Likewise, the various `cancel()` calls were never handled by the `select` because:
  * First `cancel()` call happens when the function is completed.
  * Second `cancel()` call happens when the `select` handles `case <- interruptChan` and since the select isn't in a loop, it won't have a handle to handle another case statement.

**The 2nd takeaway** is that the cancelled context is critical for sub-commands that have registered for the cleanup wait group and need to be notified when the main execution is being terminated (typically, because CTRL+C was pressed). So, we cannot remove the cancelled context logic.

**Update based on what we've learned:**

From our learnings, commit 3de7493:

1. Documents the cancelled context and explains why it's important and how it's used by sub-commands
2. Removed the `case <-ctx.Done()` because the main execution doesn't need to handle a cancelled context.
    * The cancelled context is valuable for sub-commands.
    * The main execution needs to handle the interrupt signals and completed channel, which in turn may cancel the context when necessary.

😅

cmd/root.go
```diff
@@ -327,20 +327,28 @@ func InitConfig(ctx context.Context, clients *shared.ClientFactory, rootCmd *cob
 // It listens for process interrupts and sends to IOStreams' GetInterruptChannel() for use in
 // in communicating process interrupts elsewhere in the code.
 func ExecuteContext(ctx context.Context, rootCmd *cobra.Command, clients *shared.ClientFactory) {
+	// Derive a cancel context that is cancelled when the main execution is interrupted or cleaned up.
+	// Sub-commands can register for the cleanup wait group with clients.CleanupWaitGroup.Add(1)
+	// and listen for <-ctx.Done() to be notified when the main execution is terminated and have
+	// a chance to cleanup. This is useful for long running processes and background goroutines,
+	// the activity and upgrade commands.
 	ctx, cancel := context.WithCancel(ctx)
```
mwbrooks: note: Big comment explaining why we derive the cancel context and how sub-commands can use it.

cmd/root.go
```diff
@@ -355,18 +363,17 @@ func ExecuteContext(ctx context.Context, rootCmd *cobra.Command, clients *shared
 				clients.IO.PrintDebug(ctx, "Exiting with cancel exit code.")
 				os.Exit(int(iostreams.ExitCancel))
 			}()
-		case <-ctx.Done():
-			// No interrupt signal sent, command executed to completion
-			// FIXME - `.Done()` channel is triggered by `cancel()` and not a successfully completion
```
mwbrooks, line 360: note: 🪓 Removed the `case <-ctx.Done():` so that a cancelled context doesn't negate the the `select` statement. We want to allow the `select` statement to handle the interrupt or completed signals/channels.

cmd/root.go
```diff
@@ -327,20 +327,28 @@ func InitConfig(ctx context.Context, clients *shared.ClientFactory, rootCmd *cob
 // It listens for process interrupts and sends to IOStreams' GetInterruptChannel() for use in
 // in communicating process interrupts elsewhere in the code.
 func ExecuteContext(ctx context.Context, rootCmd *cobra.Command, clients *shared.ClientFactory) {
+	// Derive a cancel context that is cancelled when the main execution is interrupted or cleaned up.
+	// Sub-commands can register for the cleanup wait group with clients.CleanupWaitGroup.Add(1)
+	// and listen for <-ctx.Done() to be notified when the main execution is terminated and have
```
zimeg: Awesome language choice here 📚 ✨ 

I was going to suggest "interrupted" instead of "terminated" without realizing that the deferred `cancel` below is so useful for whatever goroutine cleanup might exist!

zimeg (approved):
LGTM! And I'm so glad to've paired on this 🎁 

This was confusing to untangle at first, but removing this case - and adding a comment above - will make future investigations so much more clear.

Finding that recent changes make a canceled context from interrupts meaningful to subcommands is wonderful too. Big cheers to your efforts here 🏆 ✨ 

Before merging we might want to update the PR title, although I do not have a good suggestion...

After merging I'm excited to continue testing this in development. Changes to `root` without tests has caused some stress before. But I realize too that you're ahead of this with #61 😉

cmd/root.go
mwbrooks: I think "interrupted" is a good choice and more general than "terminated." Commit 68ecf29 updates the wording to be interrupted!

zimeg: Ooh I'm alright with either! Interrupted might be confused with an actual `^C` but the surrounding wording I think is helpful so no worries with either choice 🤓

Edit: With the first sentence I find this LGTM as is- apologies for potential stress this comment might've caused!

mwbrooks:
Thanks for the second eyes @zimeg! This PR ended up in a MUCH better place because you took the time to think through the intention and question the implementation! 🙇🏻 

I've updated the PR title and also the **CHANGELOG** section in the PR Description.

📚 Side note: I hope the **CHANGELOG** section helps the Education team more easily write docs.slack.com/changelog entires for complex PRs.

PR #61, opened 2025-04-23
test: add success and error tests to cmd.ExecuteContext

mwbrooks:
### Summary

This pull request adds 2 test-cases to `cmd.ExecuteContext`:

* Command successfully executes
* Command returns an error

We do not test the following test-cases yet because they're a little tricky:

* Interrupt Signal (CTRL+C)
* Cancelled context

### Requirements

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -356,7 +356,7 @@ func ExecuteContext(ctx context.Context, rootCmd *cobra.Command, clients *shared
 				clients.IO.PrintDebug(ctx, "... root cleanup waitgroup done.")
 				cleanup(ctx, clients)
 				clients.IO.PrintDebug(ctx, "Exiting with cancel exit code.")
-				os.Exit(int(iostreams.ExitCancel))
+				clients.Os.Exit(int(iostreams.ExitCancel))
 			}()
 		// Received cancelled context, so send an interrupt signal
 		case <-ctx.Done():
@@ -369,7 +369,7 @@ func ExecuteContext(ctx context.Context, rootCmd *cobra.Command, clients *shared
 		// If we get a second interrupt, no matter what exit the process
 		<-interruptChan
 		clients.IO.PrintDebug(ctx, "Got second process interrupt signal, exiting the process")
-		os.Exit(int(iostreams.ExitCancel))
+		clients.Os.Exit(int(iostreams.ExitCancel))
 	}()
 	// The cleanup() method in the root command will invoke via `defer` from within Execute.
 	if err := rootCmd.ExecuteContext(ctx); err != nil {
@@ -388,7 +388,7 @@ func ExecuteContext(ctx context.Context, rootCmd *cobra.Command, clients *shared
 			}
 			clients.IO.PrintError(ctx, errMsg)
 		}
-		defer os.Exit(int(clients.IO.GetExitCode()))
+		defer clients.Os.Exit(int(clients.IO.GetExitCode()))
 		completedChan <- true
 	} else {
 		completedChan <- true
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -332,15 +332,18 @@ func ExecuteContext(ctx context.Context, rootCmd *cobra.Command, clients *shared
 	completedChan := make(chan bool, 1)      // completed is used for signalling an end to command
 	exitChan := make(chan bool, 1)           // exit blocks the command from exiting until completed
 	interruptChan := make(chan os.Signal, 1) // interrupt catches signals to avoid abrupt exits
+
 	signal.Notify(interruptChan, os.Interrupt, syscall.SIGTERM)
 	defer func() {
 		signal.Stop(interruptChan)
 		cancel()
 	}()
+
+	// Wait for either an interrupt signal (ctrl+c) or an explicit call to ctx.cancel()
 	go func() {
-		// Wait for either an interrupt signal (ctrl+c), or an explicit call to ctx.cancel()
 		select {
-		case <-interruptChan: // first interrupt signal, cancel context
+		// Received interrupt signal, so cancel the context and cleanup
+		case <-interruptChan:
 			clients.IO.PrintInfo(ctx, false, "\n") // flush the CTRL + C character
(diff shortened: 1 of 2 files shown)


cmd/root.go
```diff
@@ -356,7 +356,7 @@ func ExecuteContext(ctx context.Context, rootCmd *cobra.Command, clients *shared
 				clients.IO.PrintDebug(ctx, "... root cleanup waitgroup done.")
 				cleanup(ctx, clients)
 				clients.IO.PrintDebug(ctx, "Exiting with cancel exit code.")
-				os.Exit(int(iostreams.ExitCancel))
+				clients.Os.Exit(int(iostreams.ExitCancel))
```
mwbrooks, line 364: note: Switched a few of these `os.Exit` to our mock-able `clients.Os.Exit`. This is necessary for the tests to pass, otherwise the entire process is terminated.

cmd/root_test.go
```diff
@@ -76,6 +78,65 @@ func TestRootCommand(t *testing.T) {
 	}
 }
 
+func TestExecuteContext(t *testing.T) {
```
mwbrooks, line 81: note: I didn't use the table test format because I think the future Interrupt and Cancelled Context tests will need a custom format. However, I can refactor this to the table test format if the reviewers prefer.

cmd/root_test.go
```diff
@@ -76,6 +78,65 @@ func TestRootCommand(t *testing.T) {
...
+	t.Run("successful execution", func(t *testing.T) {
```
mwbrooks: note: Command completes successfully test-case

cmd/root_test.go
```diff
@@ -76,6 +78,65 @@ func TestRootCommand(t *testing.T) {
...
+		}
+
+		// Execute the command
+		ExecuteContext(ctx, cmd, clients)
+
+		// Assert exit code is success
+		require.Equal(t, iostreams.ExitOK, clients.IO.GetExitCode())
+	})
+
+	t.Run("handles command error", func(t *testing.T) {
```
mwbrooks: note: Command returns an error test-case.

mwbrooks (commented):
A few notes to guide the reviewer and help make a few decisions:

cmd/root_test.go
zimeg: Small preference for table tests to avoid duplicated mock setups, but I can also imagine interrupt handling needing a special case...

No blocker for this PR but it might be a nice change for now?

cmd/root_test.go
```diff
@@ -76,6 +78,65 @@ func TestRootCommand(t *testing.T) {
...
+		testutil.MockCmdIO(clients.IO, cmd)
+
+		// Create a test command that returns an error
+		expectedErr := fmt.Errorf("command failed")
+		cmd.RunE = func(cmd *cobra.Command, args []string) error {
+			return expectedErr
+		}
+
+		// Execute the command
+		ExecuteContext(ctx, cmd, clients)
```
zimeg: TBH I was surprised to find we're not returning an `err` here and instead exit within the `root` command instead of `main.go` but it's also not clear if changing this makes sense...

zimeg: 📝 The same blog referenced in #60 has an interesting article on a minimal `main` with additional testing benefits:

https://pace.dev/blog/2020/02/12/why-you-shouldnt-use-func-main-in-golang-by-mat-ryer.html

cmd/root.go
zimeg: Amazing change! It's super nice to have these interfaces abstract the actual calls for these mocking purposes 🎁 ✨

cmd/root_test.go
```diff
@@ -76,6 +78,65 @@ func TestRootCommand(t *testing.T) {
...
+
+		// Assert exit code is error
```
zimeg: 🤔 Would it be useful or too much to also check that the correct error and code are included in logstashes?

```go
	clientsMock.EventTracker.AssertCalled(t, "FlushToLogstash", mock.Anything, mock.Anything, mock.Anything, iostreams.ExitError)
```

Genuine question since this doesn't seem like critical behavior, but it might guard against regression 👾

cmd/root_test.go
mwbrooks: No problem, I think it's fair to optimize the tests and reduce the duplicate code. Commit [5e57be4](https://github.com/slackapi/slack-cli/pull/61/commits/5e57be43e50db68ac0503773da63beb2cc7e3e66) rewrites the tests as a table test (and it's much shorter!)

cmd/root_test.go
mwbrooks: I had the same thought that `cmd.ExecuteContext` should return an error. I imagine we didn't intentional choose to not return an error, so we change improve this in the future.

I also ❤️ love that article. While dealing with #60, I had a feeling that we may want to radically simplify `main.go` and have it call `os.Exit`. Since `main()` is not testable, we want to keep it's footprint small and since `os.Exit` is problematic to test, it makes sense to isolate it as well. This would allow us to deeply test `cmd/root.go`.

cmd/root_test.go
mwbrooks: 🤔 I think it's a very good idea to add unit tests around Logstash. Our support team relies on the log data to help customers and soon we'll be doing the same on GitHub Issues. So, having confidence that we're flushing quality data to the logs is important 👍🏻 

I tried to add a test but it looks like our EventTracker Mock isn't wired up to Clients Mock and isn't completely implemented 🙁 I've finished wiring it up but it's a decent amount of code, so I'll share it in a separate PR. In the meantime, Commit [f641e1b](https://github.com/slackapi/slack-cli/pull/61/commits/f641e1b82e2ed7c0ad3c5058f3eb16251fc00f7a) add a TODO.

I've already got the EventTracker Mock updates stashed, so I'll publish a PR soon! 🤞🏻

mwbrooks:
Thanks a bunch for the review @zimeg! 🎉 I've refactored the tests into a table test format and added a TODO to test the Logstash Flush. I'll fast-follow to turn that `TODO → PR`!

PR #64, opened 2025-04-25
test: assert cmd.ExecuteContext flushes exit code to logstash

mwbrooks:
### Summary

This pull request is a follow-up for https://github.com/slackapi/slack-cli/pull/61#discussion_r2059054925 that asserts `cmd.ExecuteContext` flushes the correct exit code to Logstash.

In order for this to work, we needed to fix the implementation of the `EventTrackerMock` by adding a few missing functions, writing it up to `ClientsMock`, and adding default mocks to allow our existing tests to pass.

I think we're at the point where we can test Logstash flushes throughout the code! 🎉 

### Requirements

### DIFF
--- a/cmd/root_test.go
+++ b/cmd/root_test.go
@@ -104,6 +104,7 @@ func TestExecuteContext(t *testing.T) {
 			// Mock clients
 			clientsMock := shared.NewClientsMock()
 			clientsMock.AddDefaultMocks()
+			clientsMock.EventTracker.On("FlushToLogstash", mock.Anything, mock.Anything, mock.Anything, mock.Anything).Return(nil)
 			clients := shared.NewClientFactory(clientsMock.MockClientFactory())
 
 			// Mock command
--- a/cmd/root_test.go
+++ b/cmd/root_test.go
@@ -27,6 +27,7 @@ import (
 	"github.com/slackapi/slack-cli/test/testutil"
 	"github.com/spf13/cobra"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
 	"github.com/stretchr/testify/require"
 )
 
@@ -118,8 +119,8 @@ func TestExecuteContext(t *testing.T) {
 			output := clientsMock.GetCombinedOutput()
 
 			// Assertions
-			// TODO: Assert that the event tracker was called with the correct exit code
 			require.Equal(t, tt.expectedExitCode, clients.IO.GetExitCode())
+			clientsMock.EventTracker.AssertCalled(t, "FlushToLogstash", mock.Anything, mock.Anything, mock.Anything, tt.expectedExitCode)
 
 			for _, expectedOutput := range tt.expectedOutputs {
 				require.Contains(t, output, expectedOutput)
--- a/internal/iostreams/iostreams_mock.go
+++ b/internal/iostreams/iostreams_mock.go
@@ -118,5 +118,3 @@ func (m *IOStreamsMock) InitLogFile(ctx context.Context) error {
 func (m *IOStreamsMock) FinishLogFile(ctx context.Context) {}
 
 func (m *IOStreamsMock) FlushToLogFile(ctx context.Context, prefix, errStr string) error { return nil }
-
-func (m *IOStreamsMock) FlushToLogstash(ctx context.Context) error { return nil }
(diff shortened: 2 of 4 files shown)


cmd/root_test.go
```diff
@@ -118,8 +119,8 @@ func TestExecuteContext(t *testing.T) {
 			output := clientsMock.GetCombinedOutput()
 
 			// Assertions
-			// TODO: Assert that the event tracker was called with the correct exit code
 			require.Equal(t, tt.expectedExitCode, clients.IO.GetExitCode())
+			clientsMock.EventTracker.AssertCalled(t, "FlushToLogstash", mock.Anything, mock.Anything, mock.Anything, tt.expectedExitCode)
```
mwbrooks, line 124: note: This is the most important part of the PR. We now assert that the correct exit code is flushed to Logstash.

internal/tracking/tracking_mock.go
```diff
@@ -28,13 +28,32 @@ type EventTrackerMock struct {
 
 func (m *EventTrackerMock) AddDefaultMocks() {
 	m.On("FlushToLogstash", mock.Anything, mock.Anything, mock.Anything, mock.Anything).Return(nil)
```
zimeg:
```suggestion
```

🪓 Ignore this suggestion but I'm now wondering if we can be confident that we're setting up all mock implementations within tests for this method, so a default is no longer needed?

internal/iostreams/iostreams_mock.go
```diff
@@ -118,5 +118,3 @@ func (m *IOStreamsMock) InitLogFile(ctx context.Context) error {
 func (m *IOStreamsMock) FinishLogFile(ctx context.Context) {}
 
 func (m *IOStreamsMock) FlushToLogFile(ctx context.Context, prefix, errStr string) error { return nil }
-
-func (m *IOStreamsMock) FlushToLogstash(ctx context.Context) error { return nil }
```
zimeg, line 122: 🪓 Removed this implementations since this method no longer exists within `iostreams`!

zimeg (approved):
@mwbrooks LGTM! 🧪 ✨

I didn't mean for additional refactor needed to test exit codes here, but it's nice to see the improvements to the testing suite.

Commit e35caa89c4a7129da85558eed79d1e27957e4fa7 removed perhaps past flushes from `iostreams` instead of the event tracker since I don't believe that's used now, but please feel free to revert that as needed. I could not suggest it so quick otherwise 😅

I'm also bummed to see coverage drop from happenchance of the actual test run - IIRC ordering random lists is difficult to make deterministic - but the actual tests are improved so much here! 🚢 💨

internal/tracking/tracking_mock.go
mwbrooks: Sure, I'm okay with that. We should think about how we want to approach the `.DefaultMocks()`. I think it saves us a lot of scaffolded code, but sometimes comes off as magical and leads to brittle tests that dependent on the default mocks. 🤔 

However, since `FlushToLogstash` is only used on `root.go`, I think it's a safe one to always mock explicitly.

Commit 5128747 removes the default mock and declares it explicitly in the `root_test.go` table test.

internal/iostreams/iostreams_mock.go
mwbrooks: Ah great catch! I noticed that while refactoring, but it slipped from my mind.

mwbrooks:
> I didn't mean for additional refactor needed to test exit codes here, but it's nice to see the improvements to the testing suite.

Ha! No worries, this is a healthy refactor, so I'm glad jumped on the opportunity 🤾🏻 

> Commit [e35caa8](https://github.com/slackapi/slack-cli/commit/e35caa89c4a7129da85558eed79d1e27957e4fa7) removed perhaps past flushes from `iostreams` instead of the event tracker since I don't believe that's used now, but please feel free to revert that as needed. I could not suggest it so quick otherwise 😅

Totally cool and this is a very good catch! Thanks for pushing up the commit!

> I'm also bummed to see coverage drop from happenchance of the actual test run - IIRC ordering random lists is difficult to make deterministic - but the actual tests are improved so much here! 🚢 💨

Yea, it's unfortunate, but overall we're inching toward that 70% mark. There's still some low hanging fruit 🍒 that we can use to increase the coverage more as well!

PR #65, opened 2025-04-25
Fix example command references in enterprise docs

amarinelli:
### Summary

Small update to fix example command references which were not using double hyphen for org workspace grant flag.

### Requirements

### DIFF
--- a/docs/guides/using-slack-cli-on-an-enterprise-grid-organization.md
+++ b/docs/guides/using-slack-cli-on-an-enterprise-grid-organization.md
@@ -93,13 +93,13 @@ You can also skip the prompt and proactively grant the desired access by appendi
 For running on a local development server:
 
 ```
-slack run -org-workspace-grant <team_id>
+slack run --org-workspace-grant <team_id>
 ```
 
 For deploying to production:
 
 ```
-slack deploy -org-workspace-grant <team_id>
+slack deploy --org-workspace-grant <team_id>
 ```
 
 ## Adjusting access to specific workspaces {#adjusting-access}


zimeg:
@amarinelli Super sharp eye! Thanks for catching this and sharing a fix so fast 👁️‍🗨️ ✨ 

I am noticing the `docs` workflow is not sharing secrets with branches of a fork though! 🤔 

@lukegalbraithrussell I'm wondering if we'll want to run this workflow in an [`environment`](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-deployments/managing-environments-for-deployment) to build the site with **tokens**? IMO it is important to have "successful" checks before a merge.

The `slackapi/slack-github-action` uses this approach to run [tests](https://github.com/slackapi/slack-github-action/blob/aef4dbb5739df4e2f43fe325742203d16310cc70/.github/workflows/test.yml#L13) with IRL Slack tokens as a possible reference!

mwbrooks (commented):
🙌🏻 Yay, thanks for the fix @amarinelli!

💯 Agree with @zimeg that we want to see our GitHub Checks to passing before merging these PRs. This seems like a good PR to work out the kinks.

amarinelli:
Thanks, all! I noticed the failure and was reading through contributor/maintainer docs to see if that was on me to build the docs or similar. Let me know if there is any actions I need to take now or next time.

lukegalbraithrussell:
@amarinelli Yeah this is a docs ci edge case I didn't think about. Nothing for you to do rn -- I'm looking into a solution today going off @zimeg's suggestions

PR #66, opened 2025-04-25
feat: update 'feedback' command to include feedback about the Slack CLI

mwbrooks:
### CHANGELOG

> Updated `feedback` command to display a prompt for Slack CLI feedback. Now developers can ask questions, submit issues, or suggest features for the Slack CLI through GitHub Issues.

### Summary

This pull request updates the `feedback` command to display a prompt with 2 choices:

1. Slack CLI
2. Slack Platform

While expanding `feedback` to display multiple options, there are a lot of improvements that we can make. However, I've tried to scope this PR to only adding the 2nd option and tweaking things to look & feel alright. 

**The main changes are:**

* **Added option for "Slack CLI Feedback"**
* **Renamed the original option to "Slack Platform Feedback"**
    * Previously, it was "Help make the Slack platform better"
* **Added a `Never` frequency**
    * So that the Slack CLI Feedback is never, randomly asked
    * The Slack Platform Feedback continues to be randomly asked after some commands
* **Added a `SkipQueryParams` field (optional)**
    * So that the the `system_id`, `project_id`, and other query string params are not included in the https://github.com/slackapi/slack-cli/issues URL
    * Default is `false`, so the Slack Platform Feedback URL continues to have the query string params
* **Updated `slacktrace.FeedbackMessage` to include the feedback option chosen**
    * e.g. `SLACK_TRACE_FEEDBACK_MESSAGE=slack-cli-feedback`
* **Generalized "survey" wording to be "feedback" or "options"**
    * This isn't perfect, but it's a step in the right direction
    * It just felt weird to see things like "Open survey in browser?" when a user was opening GitHub Issues

### Preview

https://github.com/user-attachments/assets/8f707c8f-8e2c-4bd2-b98c-afad45afc450

### Reviewers

```bash
# Feedback Prompt
$ lack feedback
# → Select Slack CLI
# → Confirm opening in browser without qs params

$ lack feedback
# → Confirm completed timestamp on Slack CLI
# CTRL+C

# Feedback Flag
$ lack feedback --name slack-cli-feedback
# → Confirm skips prompt and asks to open browser
# CTRL+C

# Feedback Trace
$ SLACK_TEST_TRACE=1 lack feedback --verbose
# → Select something
# → Confirm Trace with selection
# CTRL+C

# Post-Command Feedback Prompt
$ lack doctor
# → Confirm feedback message after command completes
```

### Requirements

### DIFF
--- a/cmd/feedback/feedback.go
+++ b/cmd/feedback/feedback.go
@@ -90,7 +90,7 @@ var SurveyStore = map[string]SlackSurvey{
 			clients.IO.PrintInfo(ctx, false, fmt.Sprintf(
 				"%s\n%s\n",
 				style.Secondary("You can send us a message at "+style.Highlight(email)),
-				style.Secondary("Or, survey your experiences at "+style.Highlight("https://docs.slack.dev/developer-support")),
+				style.Secondary("Or, share your experiences at "+style.Highlight("https://docs.slack.dev/developer-support")),
 			))
 		},
 		Trace: slacktrace.FeedbackMessage,
@@ -206,7 +206,8 @@ func NewFeedbackCommand(clients *shared.ClientFactory) *cobra.Command {
 		Short:   "Share feedback about your experience or project",
 		Long:    "Help us make the Slack Platform better with your feedback",
 		Example: style.ExampleCommandsf([]style.ExampleCommand{
-			{Command: "feedback", Meaning: "Open a feedback survey in your browser"},
+			{Command: "feedback", Meaning: "Choose to give feedback on part of the Slack Platform"},
+			{Command: "feedback --name slack-cli-feedback", Meaning: "Give feedback on the Slack CLI"},
 		}),
 		PreRun: func(cmd *cobra.Command, args []string) {
 			clients.Config.SetFlags(cmd)
@@ -225,7 +226,7 @@ func NewFeedbackCommand(clients *shared.ClientFactory) *cobra.Command {
 	}
 	sort.Strings(surveyNames)
 	nameFlagDescription := style.Sectionf(style.TextSection{
-		Text:      "name of the survey:",
+		Text:      "name of the feedback:",
 		Secondary: surveyNames,
 	})
 	cmd.Flags().StringVar(&surveyNameFlag, "name", "", nameFlagDescription)
@@ -238,7 +239,7 @@ func NewFeedbackCommand(clients *shared.ClientFactory) *cobra.Command {
 // runFeedbackCommand will open the user's browser to the feedback survey webpage.
 func runFeedbackCommand(ctx context.Context, clients *shared.ClientFactory, cmd *cobra.Command) error {
 	if len(SurveyStore) == 0 {
-		clients.IO.PrintInfo(ctx, false, "No surveys currently available; please try again later")
+		clients.IO.PrintInfo(ctx, false, "No feedback options currently available; please try again later")
 		return nil
 	}
 
@@ -246,14 +247,14 @@ func runFeedbackCommand(ctx context.Context, clients *shared.ClientFactory, cmd 
 
 	if _, ok := SurveyStore[surveyNameFlag]; !ok && surveyNameFlag != "" {
 		return slackerror.New("invalid_survey_name").
-			WithMessage("Invalid survey name provided: %s", surveyNameFlag).
-			WithRemediation("View survey options with %s", style.Commandf("feedback --help", false))
+			WithMessage("Invalid feedback name provided: %s", surveyNameFlag).
+			WithRemediation("View the feedback options with %s", style.Commandf("feedback --help", false))
 	}
 
 	if surveyNameFlag == "" && noPromptFlag {
 		return slackerror.New("survey_name_required").
-			WithMessage("Please provide a survey name or remove the --no-prompt flag").
-			WithRemediation("View survey options with %s", style.Commandf("feedback --help", false))
+			WithMessage("Please provide a feedback name or remove the --no-prompt flag").
+			WithRemediation("View feedback options with %s", style.Commandf("feedback --help", false))
 	}
 
 	clients.IO.PrintInfo(ctx, false, style.Sectionf(style.TextSection{
@@ -339,7 +340,7 @@ func executeSurvey(ctx context.Context, clients *shared.ClientFactory, s SlackSu
 	var err error
 	var ok bool
 	if !noPromptFlag {
-		ok, err = clients.IO.ConfirmPrompt(ctx, "Open survey in browser?", true)
+		ok, err = clients.IO.ConfirmPrompt(ctx, "Open in browser?", true)
 		if err != nil {
 			return err
 		}
@@ -353,7 +354,7 @@ func executeSurvey(ctx context.Context, clients *shared.ClientFactory, s SlackSu
 	if ok { // Open survey in browser
 		clients.Browser().OpenURL(u)
 	} else { // Print survey URL
-		clients.IO.PrintInfo(ctx, false, fmt.Sprint("Survey URL: \n", style.Secondary(u)))
+		clients.IO.PrintInfo(ctx, false, fmt.Sprint("Feedback URL: \n", style.Secondary(u)))
 	}
 
 	// Record completion
--- a/cmd/feedback/feedback.go
+++ b/cmd/feedback/feedback.go
@@ -63,13 +63,13 @@ const (
 	Always  Frequency = iota // We always want to ask
 	Once                     // Ask user to complete the survey only once
 	Monthly                  // Ask user to complete the survey once a month
+	Never                    // Do not ask the user to complete the survey
 )
 
 // Supported survey names
 const (
-	PlatformSurvey = "platform-improvements"
-	// TODO: uncomment when the survey is ready for release
-	//ProjectInfoSurvey = "project-info"
+	PlatformSurvey   = "platform-improvements"
+	SlackCLIFeedback = "slack-cli-feedback"
 )
 
 type SurveyConfigInterface interface {
@@ -109,6 +109,37 @@ var SurveyStore = map[string]SlackSurvey{
 			return clients.Config.SystemConfig
 		},
 	},
+	// SlackCLIFeedback asks for Slack CLI feedback using GitHub Issues
+	SlackCLIFeedback: {
+		Name:              SlackCLIFeedback,
+		PromptDisplayText: "Slack CLI",
+		PromptDescription: "Questions, issues, and feature requests about the Slack CLI",
+		URL: url.URL{
+			RawPath: "https://github.com/slackapi/slack-cli/issues",
+		},
+		Info: func(ctx context.Context, clients *shared.ClientFactory) {
+			clients.IO.PrintInfo(ctx, false, fmt.Sprintf(
+				"%s\n%s\n",
+				style.Secondary("Ask questions, submit issues, or suggest features for the SLack CLI:"),
+				style.Secondary(style.Highlight("https://github.com/slackapi/slack-cli/issues")),
+			))
+		},
+		Trace: slacktrace.FeedbackMessage,
+		Ask: func(ctx context.Context, clients *shared.ClientFactory) (bool, error) {
+			clients.IO.PrintInfo(ctx, false, style.Sectionf(style.TextSection{
+				Emoji: "love_letter",
+				Text:  "We would love to know how things are going",
+				Secondary: []string{
+					"Share your experience with " + style.Commandf("feedback --name slack-cli-feedback", false),
+				},
+			}))
+			return false, nil
+		},
+		Frequency: Never,
+		Config: func(clients *shared.ClientFactory) SurveyConfigInterface {
+			return clients.Config.SystemConfig
+		},
+	},
 }
 
 // SetAskedAtTimestamp writes a timestamp for when the survey was last asked to the project or system level config
@@ -134,6 +165,10 @@ func (s SlackSurvey) SetCompletedAtTimestamp(ctx context.Context, clients *share
 
 // ShouldAsk returns true if we should ask the user the complete the survey
 func (s SlackSurvey) ShouldAsk(cfg config.SurveyConfig) (bool, error) {
+	if s.Frequency == Never {
+		return false, nil
+	}
+
 	if cfg.AskedAt == "" { // survey has never been asked before
 		return true, nil
 	}
--- a/cmd/feedback/feedback.go
+++ b/cmd/feedback/feedback.go
@@ -104,7 +104,7 @@ var SurveyStore = map[string]SlackSurvey{
 				Emoji: "love_letter",
 				Text:  "We would love to know how things are going",
 				Secondary: []string{
-					"Share your experience with " + style.Commandf("feedback --name slack-cli-feedback", false),
+					"Share your experience with " + style.Commandf(fmt.Sprintf("feedback --name %s", SlackCLIFeedback), false),
 				},
 			}))
 			return false, nil
@@ -210,7 +210,7 @@ func NewFeedbackCommand(clients *shared.ClientFactory) *cobra.Command {
 		Long:    "Help us make the Slack Platform better with your feedback",
 		Example: style.ExampleCommandsf([]style.ExampleCommand{
 			{Command: "feedback", Meaning: "Choose to give feedback on part of the Slack Platform"},
-			{Command: "feedback --name slack-cli-feedback", Meaning: "Give feedback on the Slack CLI"},
+			{Command: fmt.Sprintf("feedback --name %s", SlackCLIFeedback), Meaning: "Give feedback on the Slack CLI"},
 		}),
 		PreRun: func(cmd *cobra.Command, args []string) {
 			clients.Config.SetFlags(cmd)
--- a/cmd/feedback/feedback.go
+++ b/cmd/feedback/feedback.go
@@ -40,7 +40,7 @@ type SlackSurvey struct {
 	PromptDisplayText string
 	// PromptDescription is displayed beneath the `feedback` command prompt option
 	PromptDescription string
-	// SkipQueryParams is a flag to skip adding query params to the survey URL
+	// SkipQueryParams is a flag to skip adding query params to the survey URL (optional, default false)
 	SkipQueryParams bool
 	// URL is the survey URL
 	URL url.URL
--- a/cmd/feedback/feedback.go
+++ b/cmd/feedback/feedback.go
@@ -68,8 +68,8 @@ const (
 
 // Supported survey names
 const (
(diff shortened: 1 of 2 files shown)


cmd/feedback/feedback.go
```diff
@@ -40,6 +40,8 @@ type SlackSurvey struct {
 	PromptDisplayText string
 	// PromptDescription is displayed beneath the `feedback` command prompt option
 	PromptDescription string
+	// SkipQueryParams is a flag to skip adding query params to the survey URL (optional, default false)
+	SkipQueryParams bool
```
mwbrooks, line 44: note: Added a new, optional param to skip the query string params. This produces a clean URL for the GitHub Issues, which don't use our query string data.

cmd/feedback/feedback.go
```diff
@@ -63,13 +65,13 @@ const (
 	Always  Frequency = iota // We always want to ask
 	Once                     // Ask user to complete the survey only once
 	Monthly                  // Ask user to complete the survey once a month
+	Never                    // Do not ask the user to complete the survey
```
mwbrooks, line 68: note: Added a new frequency option to `Never` ask the user to complete a survey.

cmd/feedback/feedback.go
```diff
@@ -63,13 +65,13 @@ const (
...
 )
 
 // Supported survey names
 const (
-	PlatformSurvey = "platform-improvements"
-	// TODO: uncomment when the survey is ready for release
-	//ProjectInfoSurvey = "project-info"
+	SlackCLIFeedback      = "slack-cli-feedback"
+	SlackPlatformFeedback = "platform-improvements"
```
mwbrooks: note: We can't rename the `platform-improvements` ID without some migration logic, but how does the `slack-cli-feedback` ID look? We'll follow this pattern for the Bolt Frameworks, so let's make sure we're happy with it. 🕹️

cmd/feedback/feedback.go
```diff
@@ -80,17 +82,49 @@ type SurveyConfigInterface interface {
 // SurveyStore stores all available surveys.
 // New surveys should be added here.
 var SurveyStore = map[string]SlackSurvey{
-	// PlatformSurvey asks for general developer experience feedback
-	PlatformSurvey: {
-		Name:              PlatformSurvey,
-		PromptDisplayText: "Help make the Slack platform better",
-		PromptDescription: "Tell us about your experience as a Slack developer",
+	// SlackCLIFeedback asks for Slack CLI feedback using GitHub Issues
+	SlackCLIFeedback: {
```
mwbrooks, line 86: note: Here's a screenshot of the prompt. How do the title and descriptions look?
image

cmd/feedback/feedback.go
```diff
@@ -80,17 +82,49 @@ type SurveyConfigInterface interface {
...
+			RawPath: "https://github.com/slackapi/slack-cli/issues",
+		},
+		Info: func(ctx context.Context, clients *shared.ClientFactory) {
+			clients.IO.PrintInfo(ctx, false, fmt.Sprintf(
+				"%s\n%s\n",
+				style.Secondary("Ask questions, submit issues, or suggest features for the SLack CLI:"),
+				style.Secondary(style.Highlight("https://github.com/slackapi/slack-cli/issues")),
+			))
+		},
+		Trace: slacktrace.FeedbackMessage,
```
mwbrooks, line 101: note: In the future, we could make the trace optional or removed entirely. I think it's better to have 1 trace for all feedback options (`slacktrace.FeedbackMessage`) then include the selected feedback name as trace data.

cmd/feedback/feedback.go
```diff
@@ -80,17 +82,49 @@ type SurveyConfigInterface interface {
...
+		Ask: func(ctx context.Context, clients *shared.ClientFactory) (bool, error) {
+			clients.IO.PrintInfo(ctx, false, style.Sectionf(style.TextSection{
+				Emoji: "love_letter",
+				Text:  "We would love to know how things are going",
+				Secondary: []string{
+					"Share your experience with " + style.Commandf("feedback --name slack-cli-feedback", false),
+				},
+			}))
+			return false, nil
+		},
```
mwbrooks: note: Since we `Never` randomly ask about this survey, this function is never actually executed. But it's a required field.

cmd/feedback/feedback.go
```diff
@@ -80,17 +82,49 @@ type SurveyConfigInterface interface {
...
+		Frequency: Never,
+		Config: func(clients *shared.ClientFactory) SurveyConfigInterface {
+			return clients.Config.SystemConfig
+		},
+	},
+	// SlackPlatformFeedback asks for general developer experience feedback
+	SlackPlatformFeedback: {
+		Name:              SlackPlatformFeedback,
+		PromptDisplayText: "Slack Platform",
+		PromptDescription: "Developer support for the Slack Platform, Slack API, Block Kit, and more",
```
mwbrooks, line 121: note: Renamed this to be focused on Slack Platform features - API, Block Kit, etc. Is the distinction clear enough?

cmd/feedback/feedback.go
```diff
@@ -80,17 +82,49 @@ type SurveyConfigInterface interface {
...
 		URL:               url.URL{RawPath: "https://docs.slack.dev/developer-support"},
 		Info: func(ctx context.Context, clients *shared.ClientFactory) {
 			clients.IO.PrintInfo(ctx, false, fmt.Sprintf(
 				"%s\n%s\n",
 				style.Secondary("You can send us a message at "+style.Highlight(email)),
-				style.Secondary("Or, survey your experiences at "+style.Highlight("https://docs.slack.dev/developer-support")),
+				style.Secondary("Or, share your experiences at "+style.Highlight("https://docs.slack.dev/developer-support")),
```
mwbrooks, line 127: note: Throughout, I remove the "survey" reference for terms around sharing feedback, since we don't actually link to any surveys now.

mwbrooks (commented):
🤟🏻 Some notes to guide the reviewer through my thoughts! 💭

cmd/feedback/feedback.go
zimeg: QQ: For my own understanding, would Bolt feedback follow this pattern as:

- `bolt-feedback`

Or were you thinking that this might be runtime specific?

cmd/feedback/feedback.go
```diff
@@ -203,22 +242,22 @@ func NewFeedbackCommand(clients *shared.ClientFactory) *cobra.Command {
...
 	if len(SurveyStore) == 0 {
-		clients.IO.PrintInfo(ctx, false, "No surveys currently available; please try again later")
+		clients.IO.PrintInfo(ctx, false, "No feedback options currently available; please try again later")
 		return nil
 	}
 
 	surveyNames, surveyPromptOptions := initSurveyOpts(ctx, clients, SurveyStore)
 
 	if _, ok := SurveyStore[surveyNameFlag]; !ok && surveyNameFlag != "" {
 		return slackerror.New("invalid_survey_name").
```
zimeg, line 252:
```suggestion
		return slackerror.New(slackerror.ErrInvalidSurveyName).
```

I'm not sure if this is a coded error at the moment, but this can help match formats for quick lookups later:

```diff
- 🚫 Please provide a feedback name or remove the --no-prompt flag
+ 🚫 Please provide a feedback name or remove the --no-prompt flag (invalid_survey_name)
```

cmd/feedback/feedback.go
```diff
@@ -203,22 +242,22 @@ func NewFeedbackCommand(clients *shared.ClientFactory) *cobra.Command {
...
+			WithRemediation("View the feedback options with %s", style.Commandf("feedback --help", false))
 	}
 
 	if surveyNameFlag == "" && noPromptFlag {
 		return slackerror.New("survey_name_required").
-			WithMessage("Please provide a survey name or remove the --no-prompt flag").
-			WithRemediation("View survey options with %s", style.Commandf("feedback --help", false))
+			WithMessage("Please provide a feedback name or remove the --no-prompt flag").
+			WithRemediation("View feedback options with %s", style.Commandf("feedback --help", false))
 	}
```
zimeg, line 261: 🤔 This was a confusing message for me and I tried to remediate it with:

```sh
$ slack feedback --no-prompt slack-cli-feedback
```

I realize now the `--help` command demonstrates the `--name` flag, but I was more stumped than I'd like to admit 😉

cmd/feedback/feedback.go
```diff
@@ -299,26 +338,30 @@ func executeSurvey(ctx context.Context, clients *shared.ClientFactory, s SlackSu
 	if s.Info != nil {
 		s.Info(ctx, clients)
 	}
-	clients.IO.PrintTrace(ctx, s.Trace)
+	clients.IO.PrintTrace(ctx, s.Trace, s.Name)
```
zimeg, line 341:
```suggestion
	clients.IO.PrintTrace(ctx, slacktrace.FeedbackMessage, s.Name)
```

🪓 Since the same trace argument is used it might be more clear to hardcode that here? No blocker at all though!

cmd/feedback/feedback.go
zimeg: 🔍 I am so biased to compressed outputs with sections so was expecting something perhaps more as:

```txt
$ slack feedback

💌 Feedback
   Thanks for taking a moment to share your thoughts!

? What type of feedback would you like to give?
> Slack CLI
  Questions, issues, and feature requests about the Slack CLI
  Slack Platform
  Developer support for the Slack Platform, Slack API, Block Kit, and more

```

But perhaps with secondary text for the descriptions? This is more opinion than suggestion, but I'm also curious what you think! I do like the wording here 📚 ✨

cmd/feedback/feedback.go
zimeg: Edit: At a glance I think the current `slack-cli-feedback` is solid too!

Before settling on it I am also wondering if this might be extended more, but I like that we can have different types of issues **after** the "CLI" option is selected, on GitHub Issues.

zimeg (commented):
💌 Super quick notes before I wanted to drop before popping offline for a few minutes but this is super great!

cmd/feedback/feedback.go
mwbrooks: Good question! My thought is that we'd like to add a link for each framework that the CLI supports, so that developers can learn where to ask question. So:

* Bolt for JavaScript (`bolt-js-feedback`)
* Bolt for Python (`bolt-python-feedback`)
* Deno SDK (`deno-slack-sdk-feedback`)
* Slack CLI (`slack-cli-feedback`)

mwbrooks: For the ID, we could also go with (`-github`), which would be more accurate:

* Slack CLI (`slack-cli-github`)

Or, we could drop the post-fix entirely and follow the GitHub repo names:

* Bolt for JavaScript (`bolt-js`)
* Bolt for Python (`bolt-python`)
* Deno SDK (`deno-slack-sdk`)
* Slack CLI (`slack-cli`)

cmd/feedback/feedback.go
mwbrooks: I share you feeling that I'd like to see the prompts condensed. Unfortunately, I don't think we jump into doing it. 😢 

The reason is that this prompt is rendered by `clients.IO.SelectPrompt` and it adds the extra space when a description is provided. The same look & feel exists for `slack samples`. So, I think we would want to apply this change everywhere, consistently. 

I think it's better to wait for the BubbleTea refactor. Thoughts?

cmd/feedback/feedback.go
mwbrooks: Great suggestion! I'd like to keep this PR focused on adding the Slack CLI feedback option, but I agree that we should improve the feedback's implementation a bit!

I've pull this suggestion into PR #68, which adds `slackerror.ErrFeedbackNameInvalid` and `slackerror.ErrFeedbackNameRequired`!

cmd/feedback/feedback.go
mwbrooks: Good point, PR #68 (commit 602085f) updates this error message to be:

> "Please provide a feedback --name flag or remove the --no-prompt flag"

cmd/feedback/feedback.go
mwbrooks: I'm all for axing it! 🪓 

I've pulled this work out to PR #68, which:

* Removes the `Trace` field from the `Survey` struct
* Always uses `slacktrace.FeedbackMessage`
* Adds tests to check that the trace is printed

cmd/feedback/feedback.go
zimeg: Ahha! I didn't think to check this in implementation, but I agree BubbleTea will be a super great improvement for prompts 🫧 ✨

cmd/feedback/feedback.go
zimeg: @mwbrooks The standalone* GitHub repo seems so good! I like that it is simple.

I'm now wondering if expanding the [`doctor`](https://github.com/slackapi/slack-cli/blob/main/.github/cli-sdk-specification.md#-doctor-optional) hook responses to include the upstream paths might be helpful for this?

cmd/feedback/feedback.go
zimeg: Heh I missed this in a quick review earlier! But I so agree! Thanks for opening #68 🙏 ✨

cmd/feedback/feedback.go
zimeg:
```suggestion
	SlackCLIFeedback      = "slack-cli"
	SlackPlatformFeedback = "platform-improvements"
```

🪓 Removing the "-feedback" suffix to leave the specifics to a GitHub issue! This also makes flag options more intuitive IMO:

```diff
- $ slack feedback --name slack-cli-feedback
+ $ slack feedback --name slack-cli
```

cmd/feedback/feedback.go
zimeg: 🤔 No blocker for this PR, but it might be interesting to replace the `Ask` parameter with a "string" for `style.TextSection` inputs instead.

IMO it's somewhat confusing to keep this prompt with the "Never" option otherwise 🔍

cmd/feedback/feedback.go
```diff
@@ -299,26 +338,30 @@ func executeSurvey(ctx context.Context, clients *shared.ClientFactory, s SlackSu
...
+			return err
+		}
 	}
 
 	if ok { // Open survey in browser
-		clients.Browser().OpenURL(u)
+		clients.Browser().OpenURL(url)
 	} else { // Print survey URL
-		clients.IO.PrintInfo(ctx, false, fmt.Sprint("Survey URL: \n", style.Secondary(u)))
+		clients.IO.PrintInfo(ctx, false, fmt.Sprint("Feedback URL: \n", style.Secondary(url)))
```
zimeg, line 364: 📝 No blocker but this might be another section to revisit formatting for:

```sh

💌 Feedback
   Share thoughts or ideas or findings at the following URL:
   https://github.com/slackapi/slack-cli/issues

```

cmd/feedback/feedback.go
```diff
@@ -80,17 +82,49 @@ type SurveyConfigInterface interface {
...
+				style.Secondary(style.Highlight("https://github.com/slackapi/slack-cli/issues")),
+			))
+		},
+		Trace: slacktrace.FeedbackMessage,
+		Ask: func(ctx context.Context, clients *shared.ClientFactory) (bool, error) {
+			clients.IO.PrintInfo(ctx, false, style.Sectionf(style.TextSection{
+				Emoji: "love_letter",
+				Text:  "We would love to know how things are going",
+				Secondary: []string{
+					"Share your experience with " + style.Commandf("feedback --name slack-cli-feedback", false),
```
zimeg:
```suggestion
					"Share your experience with " + style.Commandf("feedback --name slack-cli", false),
```

cmd/feedback/feedback.go
```diff
@@ -171,7 +209,8 @@ func NewFeedbackCommand(clients *shared.ClientFactory) *cobra.Command {
 		Short:   "Share feedback about your experience or project",
 		Long:    "Help us make the Slack Platform better with your feedback",
 		Example: style.ExampleCommandsf([]style.ExampleCommand{
-			{Command: "feedback", Meaning: "Open a feedback survey in your browser"},
+			{Command: "feedback", Meaning: "Choose to give feedback on part of the Slack Platform"},
+			{Command: "feedback --name slack-cli-feedback", Meaning: "Give feedback on the Slack CLI"},
```
zimeg:
```suggestion
			{Command: "feedback --name slack-cli", Meaning: "Give feedback on the Slack CLI"},
```

AFAICT the suggested changes in comments above catch all changes from "slack-cli-feedback" to "slack-cli" if this is a change we're wanting to make!

zimeg (approved):
LGTM! Thanks for making a follow up PR with the unrelated changes noted in prior comments 🤖 ✨ 

I left a few more comments about formatting outputs that might be worth revisiting later, and a few more comments about the feedback `name` that we can settle on before merging. Once these all look good please feel free to merge 🚢

I'm looking forward to feedback and reports to come 💌 ✨

cmd/feedback/feedback.go
mwbrooks: Cool, i like the simple names for `slack-cli`, `bolt-js`, bolt-python`, etc. 

Commit f5b343f updates it to `slack feedback --name slack-cli`.

In PR #68, I'll also explore whether I can switch `platform-improvements` to `slack-platform` with backwards compatibility that we deprecated in v4.0.0. 🤔

cmd/feedback/feedback.go
mwbrooks: This is a great suggestion. I agree, the `Ask` with `Never` is unnecessary. The `style.TextSection` also simplifies the implementation while keeping the prompt more consistent.

I've made a note and we can follow-up in a future PR!

Honestly, I don't see us using the `Ask` in the originally intended way. If we don't leverage it in the future, we could also replace it with a single "ask" for feedback.

cmd/feedback/feedback.go
mwbrooks: This looks so much better than the raw "Feedback URL:" that exists right now. I'll add a note and iterate on this in future PR!

cmd/feedback/feedback.go
mwbrooks: Thanks! Commits [f5b343f](https://github.com/slackapi/slack-cli/pull/66/commits/f5b343fbb3a425e34445a80b0e321e3af5c0679f) and [48802d9](https://github.com/slackapi/slack-cli/pull/66/commits/48802d92f7cc269561cda48e1d5a94e1119dfed6) change the `--name slack-cli` and update the 2 output references!

mwbrooks:
@zimeg Thanks for the review and suggestions on future improvements! I've updated the `--name slack-cli` and I'll handle a couple of our ideas in the follow-up PR #68. I'm feeling confident that we're leaving the `feedback` command in a better place than we started! ⛺

PR #67, opened 2025-04-25
test: fix failing version tests if compilation flags are not provided

mwbrooks:
### Summary

This pull request fixes tests that require a valid semver from failing when using VSCode/Cursor.

Previously, tests in the following packages would fail:

* `cmd/root_test.go`
* `cmd/doctor/doctor_test.go`
* `cmd/version/version_test.go`

If we want to take this work a step further, we could update our `.circleci` pipeline to validate that the `BUILD_VERSION` does not contain `v0.0.0`. However, locally, I think we want the build to remain loose, so that developers can build the CLI even when their local repo doesn't contain git tags.

### Reviewers

Testing:
* All unit tests should continue to pass when using `make test`
* Running file/package tests in one of the above files in VSCode should now pass

### Preview

image

### Requirements

### DIFF
--- a/cmd/root_test.go
+++ b/cmd/root_test.go
@@ -130,7 +130,6 @@ func TestExecuteContext(t *testing.T) {
 	}
 }
 
-// FYI: do not try to run this test in vscode using the run/debug test inline test helper; as the assertions in this test will fail in that context
 func TestVersionFlags(t *testing.T) {
 	ctx := slackcontext.MockContext(t.Context())
 
--- a/internal/pkg/version/version.go
+++ b/internal/pkg/version/version.go
@@ -24,7 +24,7 @@ import (
 const EnvTestVersion = "SLACK_TEST_VERSION"
 
 // Version is set with `git describe` at build time in the Makefile
-var Version = "dev"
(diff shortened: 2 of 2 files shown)


zimeg (approved):
@mwbrooks What a nice catch! 👁️‍🗨️ 

I'm finding this fixes the exact confusion you called out, but for forked PRs. It also fixes most simple of test runs 🧪 ✨ 

```
$ go test ./...
```

>  I think we want the build to remain loose, so that developers can build the CLI even when their local repo doesn't contain git tags.

This is a good callout. I agree that this is not a check that should be within unit tests, and I agree with the following:

> If we want to take this work a step further, we could update our `.circleci` pipeline to validate that the `BUILD_VERSION` does not contain `v0.0.0`.

While we have manual steps to confirm the final compilations are built with the expected version before making a release public, I could imagine this causing headache in our own processes. No blocker here though!

mwbrooks:
Thanks again for the PR Review @zimeg! I was super curious if this helped address the forks without tags, so thanks for confirming that as well! 🙇🏻 Sounds like we can merge this PR and hold off on any extra verification from my build pipeline! 🎉

PR #69, opened 2025-04-28
refactor: enables linter rule ST1003 'poorly chosen identifier'

mwbrooks:
### Summary

This pull request enables the Linter Rule ST1003 "Poorly Chosen Identifier" and updates all identifiers to match the new rule.

For the most part, this changing `Id → ID`, `Http → HTTP`, `SOME_CONSTANT → SomeConstant`, etc.

There is no change to functionality. 🚳 

### Reviewers

I'm so so so so sorry for the HUGE PR. I think this will be a common theme as we enable more of the default linter rules. 😆 

### Requirements

* [x] I've read and understood the [Contributing Guidelines](/blob/main/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).

PR #71, opened 2025-04-30
refactor: follow-ups from ST1003

mwbrooks:
### Summary

This pull request addresses some follow-up tasks from PR #69:

* Update `Cli` references to `CLI`

### Requirements

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -33,7 +33,6 @@ linters:
     staticcheck:
       checks:
         - all
-        - '-ST1003' # disable rule 'Poorly chosen identifier'
         - '-ST1005' # disable rule 'Incorrectly formatted error string'
         - '-ST1006' # disable rule 'Poorly chosen receiver name'
         - '-ST1008' # disable rule 'Function’s error value should be its last return value'
--- a/internal/pkg/apps/install_test.go
+++ b/internal/pkg/apps/install_test.go
@@ -43,13 +43,13 @@ func TestInstall(t *testing.T) {
 
 	tests := map[string]struct {
 		mockApp                 types.App
-		mockApiCreate           api.CreateAppResult
-		mockApiCreateError      error
-		mockApiInstall          api.DeveloperAppInstallResult
-		mockApiInstallState     types.InstallState
-		mockApiInstallError     error
-		mockApiUpdate           api.UpdateAppResult
-		mockApiUpdateError      error
+		mockAPICreate           api.CreateAppResult
+		mockAPICreateError      error
+		mockAPIInstall          api.DeveloperAppInstallResult
+		mockAPIInstallState     types.InstallState
+		mockAPIInstallError     error
+		mockAPIUpdate           api.UpdateAppResult
+		mockAPIUpdateError      error
 		mockAuth                types.SlackAuth
 		mockAuthSession         api.AuthSession
 		mockBoltExperiment      bool
@@ -69,10 +69,10 @@ func TestInstall(t *testing.T) {
 	}{
 		"create a hosted app manifest with expected rosi values": {
 			mockApp: types.App{},
-			mockApiCreate: api.CreateAppResult{
+			mockAPICreate: api.CreateAppResult{
 				AppID: "A001",
 			},
-			mockApiUpdateError: slackerror.New(slackerror.ErrAppAdd),
+			mockAPIUpdateError: slackerror.New(slackerror.ErrAppAdd),
 			mockAuth: types.SlackAuth{
 				EnterpriseID: mockEnterpriseID,
 				TeamID:       mockTeamID,
@@ -87,14 +87,14 @@ func TestInstall(t *testing.T) {
 				UserID:       &mockUserID,
 			},
 			mockBoltExperiment: true,
-			mockManifestSource: config.MANIFEST_SOURCE_LOCAL,
+			mockManifestSource: config.ManifestSourceLocal,
 			mockManifest: types.SlackYaml{
 				AppManifest: types.AppManifest{
 					MetaData: &types.ManifestMetaData{
 						MajorVersion: 2,
 					},
 					Settings: &types.AppSettings{
-						FunctionRuntime: types.SLACK_HOSTED,
+						FunctionRuntime: types.SlackHosted,
 					},
 				},
 			},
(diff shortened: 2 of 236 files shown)


.golangci.yml
```diff
@@ -44,6 +44,10 @@ linters:
         - '-QF1008' # disable rule 'Omit embedded fields from selector'
         - '-QF1011' # disable rule 'Omit redundant type from variable declaration'
         - '-ST1016' # disable rule 'Use consistent method receiver names'
+      # https://golangci-lint.run/usage/linters/#staticcheck
+      # https://staticcheck.dev/docs/configuration/options/#initialisms
+      initialisms: ["ACL", "API", "ASCII", "CLI", "CPU", "CSS", "DNS", "EOF", "GUID", "HTML", "HTTP", "HTTPS", "ID", "IP", "JSON", "QPS", "RAM", "RPC", "SLA", "SMTP", "SQL", "SSH", "TCP", "TLS", "TTL", "UDP", "UI", "GID", "UID", "UUID", "URI", "URL", "UTF8", "VM", "XML", "XMPP", "XSRF", "XSS", "SDK", "SIP", "RTP", "AMQP", "DB", "TS"]
+
```
mwbrooks: question: Opinions here? The above is the latest default list for the `staticcheck` initialisms. I've also added `CLI` and `SDK` to the list.

If we don't feel comfortable maintaining our own custom list, then we can remove this and manually enforce words like `CLI` and `SDK`.

internal/pkg/apps/install_test.go
```diff
@@ -90,7 +90,7 @@ func TestInstall(t *testing.T) {
 			mockManifestSource: config.ManifestSourceLocal,
 			mockManifest: types.SlackYaml{
 				AppManifest: types.AppManifest{
-					MetaData: &types.ManifestMetaData{
+					Metadata: &types.ManifestMetadata{
```
mwbrooks, line 93: ☺️ So much better.

internal/slackerror/errors.go
```diff
@@ -73,11 +73,11 @@ const (
 	ErrConnectorDenied                               = "connector_denied"
 	ErrConnectorNotInstalled                         = "connector_not_installed"
 	ErrChannelNotFound                               = "channel_not_found"
-	ErrCliAutoupdate                                 = "cli_autoupdate_error"
-	ErrCliConfigLocationError                        = "cli_config_location_error"
-	ErrCliConfigInvalid                              = "cli_config_invalid"
-	ErrCliReadError                                  = "cli_read_error"
-	ErrCliUpdateRequired                             = "cli_update_required" // Slack API error code
+	ErrCLIAutoUpdate                                 = "cli_autoupdate_error"
```
mwbrooks, line 76: note: I also capitalized "Update"

.golangci.yml
zimeg: 👀 We might want to reduce this to what we use now and add to it as needed?

I'm not certain how accurate these checks are - making "Ts" error in a testing file put me down too much of a rabbit hole - but AFAICT we'd have alright coverage with this:

```suggestion
      # https://golangci-lint.run/usage/linters/#staticcheck
      # https://staticcheck.dev/docs/configuration/options/#initialisms
      initialisms:
        - "ACL"
        - "API"
        - "CLI"
        - "EOF"
        - "HTML"
        - "HTTP"
        - "HTTPS"
        - "ID"
        - "IP"
        - "JSON"
        - "TCP"
        - "TLS"
        - "TTL"
        - "UI"
        - "UUID"
        - "URI"
        - "URL"
        - "UTF"
        - "SDK"
        - "TS"

```

zimeg: 📝 I also lean into separate lines for faster reading and changes later, but no blocker on this!

internal/pkg/create/sample.go
```diff
@@ -93,7 +93,7 @@ func getSampleReposFromGitHub(client Sampler) ([]GithubRepo, error) {
 		return []GithubRepo{}, err
 	}
 	req.Header.Set("Accept", "application/vnd.github+json")
-	req.Header.Set("X-GitHub-Api-Version", "2022-11-28")
+	req.Header.Set("X-GitHub-API-Version", "2022-11-28")
```
zimeg: 😉 
```suggestion
	req.Header.Set("X-GitHub-Api-Version", "2022-11-28")
```

internal/pkg/create/sample_test.go
```diff
@@ -98,7 +98,7 @@ func TestSamples_GetSampleRepos(t *testing.T) {
 			req, err = http.NewRequest("GET", "https://api.github.com/orgs/slack-samples/repos?type=public", nil)
 			require.NoError(t, err)
 			req.Header.Set("Accept", "application/vnd.github+json")
-			req.Header.Set("X-GitHub-Api-Version", "2022-11-28")
+			req.Header.Set("X-GitHub-API-Version", "2022-11-28")
```
zimeg:
```suggestion
			req.Header.Set("X-GitHub-Api-Version", "2022-11-28")
```

🔍 I'm curious if a change to the `sample.go` file without a change here might fail the test?

zimeg (approved):
😵‍💫 Oof I got caught in configurations to find false negatives or raise errors on **warnings**.

This is not happening all the time AFAICT but I do not want to explore this more right now.

A response to ideas on the list of initialisms is noted below, and also a suggested change that we should make before merging, but for the most part I'm glad to see improvements to this rule.

Continued cleanups continue to be motivating for me. Even with perhaps frustrated findings 👾

.golangci.yml
mwbrooks: 👍🏻 I support shortening the list to only the initialisms that we use. Since the list is quite short to begin with, I don't think we're losing very much but customizing it. I also like using the separated lines. It makes adding a new item easier to review!

Thanks for the first pass at the new list! I searched the source code and I think you caught everything. I'll merge your commit in but I may add `UTF8` and `XML`, since I can see us referencing those at some point in the future.

mwbrooks: Commit e6803e5 adds XML and UTF8 😄

internal/pkg/create/sample.go
mwbrooks: Oof, scary find! 🙇🏻 Thank you! This is another drawback of branches on branches. While we fixed this in the original branch, the change still existed in this sub-branch! Yikes!

Commit 81569f1  reverts these changes.

internal/pkg/create/sample_test.go
mwbrooks: It should fail 🤔 

Commit 81569f1 reverts this change!

PR #75, opened 2025-05-02
refactor: enable linter rule ST1006 for 'poorly chosen receiver names'

mwbrooks:
### Summary

This pull request enables the linter `staticcheck` rule [ST1006: 'Poorly chosen receiver name'](https://staticcheck.dev/docs/checks#ST1006).

🧠 🧠 🧠

> The name of a method’s receiver should be a reflection of its identity; often a one or two letter abbreviation of its type suffices (such as “c” or “cl” for “Client”). Don’t use generic names such as “me”, “this” or “self”, identifiers typical of object-oriented languages that place more emphasis on methods as opposed to functions. 

For this PR, the only change required was to remove the `_` receiver name, which was used to represent an unused variable.

### Requirements

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -35,7 +35,6 @@ linters:
         - all
         - '-ST1003' # disable rule 'Poorly chosen identifier'
         - '-ST1005' # disable rule 'Incorrectly formatted error string'
-        - '-ST1006' # disable rule 'Poorly chosen receiver name'
         - '-ST1008' # disable rule 'Function’s error value should be its last return value'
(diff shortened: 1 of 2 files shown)


zimeg (approved):
Alas so gone be the discarded value `_` 🫡

Without doing a `grep` of all possibilities I am glad to find this rule might be generous with what's allowed. Erroring for `fs` would be frustrating IMO 🔍

mwbrooks:
Thanks for the review @zimeg! 🙇🏻 I didn't know that `_` was poor practice either OR that you can skip the receiver name entirely 🧠 It feels good to have these rules in place. 🚦

PR #77, opened 2025-05-05
refactor: rename clients.APIInterface() to clients.API()

mwbrooks:
### Summary

This pull request renames the function `clients.APIInterface()` to `clients.API()`. 

The motivation is that the name is shorter and more accurate. The `clients.API()` returns an instance of `api.Client` that implements the `api.APIInterface` interface. It doesn't return an interface. Alternatively, we could name it `clients.APIClient()` but I think we're trying to move away from this naming pattern.

There was no change  functionality was changed.

### Requirements

### DIFF
--- a/internal/shared/clients.go
+++ b/internal/shared/clients.go
@@ -97,7 +97,7 @@ func NewClientFactory(options ...func(*ClientFactory)) *ClientFactory {
 		IO: clients.IO,
 	}
 	clients.EventTracker = tracking.NewEventTracker()
-	clients.API = clients.defaultAPIInterfaceFunc
+	clients.API = clients.defaultAPIFunc
 	clients.AppClient = clients.defaultAppClientFunc
 	clients.AuthInterface = clients.defaultAuthInterfaceFunc
 	clients.Browser = clients.defaultBrowserFunc
@@ -129,8 +129,8 @@ func (c *ClientFactory) defaultAPIClientFunc() *api.Client {
 	return api.NewClient(nil, c.Config.APIHostResolved, c.IO)
 }
 
-// defaultAPIInterfaceFunc return a new API Client using the ConfigAPIHost
-func (c *ClientFactory) defaultAPIInterfaceFunc() api.APIInterface {
+// defaultAPIFunc return a new API Client using the ConfigAPIHost
+func (c *ClientFactory) defaultAPIFunc() api.APIInterface {
 	return c.defaultAPIClientFunc()
 }
 
--- a/internal/shared/clients.go
+++ b/internal/shared/clients.go
@@ -44,20 +44,16 @@ import (
 
 // ClientFactory are shared clients and configurations for use across the CLI commands (cmd) and handlers (pkg).
 type ClientFactory struct {
-	// Internal dependencies
-
-	APIInterface func() api.APIInterface
-
-	AppClient    func() *app.Client
-	Config       *config.Config
-	SDKConfig    hooks.SDKCLIConfig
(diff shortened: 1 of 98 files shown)


internal/shared/clients.go
```diff
@@ -44,20 +44,16 @@ import (
...
+	API           func() api.APIInterface
+	AppClient     func() *app.Client
 	AuthInterface func() auth.AuthInterface
+	CLIVersion    string
+	Config        *config.Config
+	EventTracker  tracking.TrackingManager
+	HookExecutor  hooks.HookExecutor
+	IO            iostreams.IOStreamer
+	Runtime       runtime.Runtime
+	SDKConfig     hooks.SDKCLIConfig
```
mwbrooks, line 56: There are 3 changes here:

1. Renamed `APIInterface → API`
2. Alphabetically ordered these fields
3. Removed newline breaks and unnecessary comments so indentation aligns

mwbrooks (commented):
A few comments to help clear up some noisy diffs! 🧘🏻

zimeg (approved):
@mwbrooks Less code is better so often! 🏆 

While we're updating this would it be best to make a similar change to tests?

https://github.com/slackapi/slack-cli/blob/aabed5e7a2a074c083d0263d6153a6ccaead6807/internal/shared/clients_mock.go#L35

internal/shared/clients.go
zimeg: Meant to send this with the review prior, but *thank you so very much*! This is a great PR to make such a change 😌 🎉

internal/shared/clients_mock.go
```diff
@@ -32,7 +32,7 @@ import (
 // ClientsMock defines mocks that will override aspects of clients for testing purposes.
 type ClientsMock struct {
 	mock.Mock
-	APIInterface  *api.APIMock
+	API           *api.APIMock
```
mwbrooks, line 35: note: Update `ClientsMock.APIInterface` to `ClientsMock.API` in commit 54da648

mwbrooks:
> @mwbrooks Less code is better so often! 🏆
> 
> While we're updating this would it be best to make a similar change to tests?
> 
> https://github.com/slackapi/slack-cli/blob/aabed5e7a2a074c083d0263d6153a6ccaead6807/internal/shared/clients_mock.go#L35

@zimeg Great catch! I'm a little disappointed in myself for not catching this - I looked through the test and mocks, but overlooked this change.

Commit 54da648 updates the `ClientsMock.APIInterface` to `ClientsMock.API`. I used a Find/ReplaceAll on `clientsMock.APIInterface`, `cm.APIInterface`, and a few manual changes in the `clients_mock.go`. It's a MASSIVE change list, but an easy scroll. I didn't notice anything amuck, but I'd appreciate it if you could take a look as well! 🙇🏻

mwbrooks:
@zimeg Sorry to keep adding more to this PR, but commit efbba6a798b0980b7e691d550e5813e753977378 also updates a `defaultAPIInterfaceFunc` reference (and removes a TODO comment).

zimeg (approved):
@mwbrooks You are so fast! 👾 ✨ 

All of these changes look great and I'm super excited to be using a more refined comments and `clients` in calling code 👁️‍🗨️

mwbrooks:
Back at you @zimeg! Thanks for taking the time to review these tedious changes! Ever-so-slowly, things look and feel better.

I'll merge this PR and follow-up with one that updates `clients.AuthInterface()` to `clients.Auth()` _(should be a lot smaller)_

PR #86, opened 2025-05-09
refactor: enable linter rule ST1016 for 'consistent method receiver names'

mwbrooks:
### Summary

This pull request enables the linter staticcheck rule [ST1016: 'Use consistent method receiver names'](https://staticcheck.dev/docs/checks/#ST1016).

For each, I choose the receiver name that was used the majority of the time.

### Requirements


internal/prompts/app_select.go
```diff
@@ -129,20 +129,20 @@ func (apps *TeamApps) IsEmpty() bool {
 // Basically, treat as a convenience getter intended for use in the context where
 // you have a team you want to filter against and you don't care whether it's an
 // auth or an app that corresponds. E.g. when you are comparing to --team flags
-func (t *TeamApps) authOrAppTeamDomain() string {
-	if t.Auth.TeamDomain != "" && (t.Auth.TeamID == t.Hosted.App.TeamID || t.Auth.TeamID == t.Local.App.TeamID) {
+func (apps *TeamApps) authOrAppTeamDomain() string {
```
mwbrooks, line 132: note: The `apps` receiver name is used the majority of the time. However, I noticed that [Golang encourages short, 1-or-2 letter receiver names](https://staticcheck.dev/docs/checks/#ST1006). In this case `a`, `t`, or `ta` would be more appropriate. However, this file is quite long and `apps` reads well. So I leaned toward less changes.

.golangci.yml
```diff
@@ -40,7 +40,6 @@ linters:
         - '-QF1004' # disable rule 'Use strings.ReplaceAll instead of strings.Replace'
         - '-QF1008' # disable rule 'Omit embedded fields from selector'
         - '-QF1011' # disable rule 'Omit redundant type from variable declaration'
-        - '-ST1016' # disable rule 'Use consistent method receiver names'
```
zimeg, line 43: 🪓 ✨

internal/iostreams/iostreams_mock.go
```diff
@@ -98,13 +98,13 @@ func (m *IOStreamsMock) IsTTY() bool {
 }
 
 // SetExitCode sets the desired exit code in a thread-safe way
-func (io *IOStreamsMock) SetExitCode(code ExitCode) {
-	atomic.StoreInt32((*int32)(&io.exitCode), int32(code))
+func (m *IOStreamsMock) SetExitCode(code ExitCode) {
```
zimeg, line 101: 🧪 Nice! I like `mock` being `m` if this is what's found elsewhere!

internal/prompts/app_select.go
zimeg: In implementations having a more complete variable name seems better to me. Thanks for linking these docs though. It makes for interesting considerations... :gift:

> The name need not be as descriptive as that of a method argument, as its role is obvious and serves no documentary purpose.

Sometimes I find it a hassle to search for single letter variables FWIW but this is the most interesting consideration. Perhaps I am needing to read more `go` 🙏 ✨

zimeg: Edit: Wow I find the change that follows this in `os.go` makes sense as a single letter. So much to consider.

zimeg (approved):
@mwbrooks Once more a wonderful PR to improve the health of code 🏥 ✨

These are appreciated and it's motivating to find a few of these landing so often - I'm feeling better and better about the code and hadn't even realized these linting rules were broken before the latest `golangci-lint` updates 🔍

internal/prompts/app_select.go
mwbrooks: Yea, I'm with you. Sometimes the single letter just doesn't feel right. For `TeamApps`, maybe rethinking the struct's name (or purpose) will help? But, overall I'm cool with `apps` :)

internal/iostreams/iostreams_mock.go
mwbrooks: Yea, `m` seems to be a common receiver name for mocks. Makes sense to me and keeps it consistent.

mwbrooks:
Thanks for the review @zimeg! While we don't have to take every linting rule to heart, it definitely helps to keep the code more consistent and readable. For the most part, I think we're already following the short receiver names, so I don't think we'll have trouble keeping with it! 🚀

PR #90, opened 2025-05-14
feat: display manifest source in human-friendly format

mwbrooks:
### CHANGELOG

> Display the manifest source as human-friendly output: "project (local)" and "app settings (remote)". You'll see these changes in commands such as `slack create`, `slack init`, and `slack app link <id>`.

### Summary

This pull request updates the `type ManifestSource` with a `manifestSource.Human()` method to display human-friendly strings during output.

The string format is `noun (id)`:
* `project (local)`
* `app settings (remote)`

Since we often need to output the manifest source, the `local` and `remote` IDs make the developer think a lot. I feel that we can be a good host by displaying these IDs in a more applicable way.

#### `slack init` or `slack create`

image

#### `slack app link <id>`

image

### Reviewer

```bash
# Create command
$ lack create asdf/
# → Select "Starter App"
# → Select "Bolt for JavaScript"
# → Confirm the output "project (local)"

$ cd asdf/

# Init command
$ lack init
# → Confirm the output "project (local)"

# Link command
$ lack app link A0123
# → Confirm warning output uses "project (local)" and "app settings (remote)" in meaningful ways
# → Select "N"

# Clean up
$ cd ../
$ rm -rf asdf/
```

### Requirements


cmd/app/link.go
```diff
@@ -173,14 +173,15 @@ func LinkExistingApp(ctx context.Context, clients *shared.ClientFactory, app *ty
...
+				"Linking an existing app requires the app manifest source to be",
+				fmt.Sprintf("managed by %s.", config.ManifestSourceRemote.Human()),
 				" ",
-				fmt.Sprintf(style.Highlight(`Your manifest source is "%s"`), manifestSource.String()),
+				fmt.Sprintf(`App manifest source can be "%s" or "%s":`, config.ManifestSourceLocal.Human(), config.ManifestSourceRemote.Human()),
 				fmt.Sprintf("- %s: uses manifest from your project's source code for all apps", config.ManifestSourceLocal.String()),
 				fmt.Sprintf("- %s: uses manifest from app settings for each app", config.ManifestSourceRemote.String()),
 				" ",
+				fmt.Sprintf(style.Highlight(`Your manifest source is "%s"`), manifestSource.Human()),
+				" ",
```
mwbrooks: note: I tweaked the wording now that we've had some time to reflect on this. Once we default Bolt projects to remote manifest, this warning will no longer be shown, so it's not a long-lived change.

mwbrooks (commented):
Quicks to explain some of my edits ✏️

internal/config/manifest_test.go
```diff
@@ -84,11 +84,11 @@ func Test_Config_ManifestSource_String(t *testing.T) {
 		a        ManifestSource
 		expected string
 	}{
-		"project manifest source is local": {
+		"local manifest source": {
 			a:        ManifestSourceLocal,
 			expected: "local",
 		},
-		"remote manifest source is remote": {
```
zimeg, line 91: 🪓 🪓

cmd/app/link.go
zimeg: 🙏 ✨ This section has been confusing to me before, but I hadn't realized that a defaulted `remote` manifest source will avoid this altogether!

cmd/app/link.go
```diff
@@ -173,14 +173,15 @@ func LinkExistingApp(ctx context.Context, clients *shared.ClientFactory, app *ty
 			Emoji: "warning",
 			Text:  "Warning",
 			Secondary: []string{
-				"Existing apps have manifests configured by app settings",
-				"Linking existing apps requires the manifest source to be " + config.ManifestSourceRemote.String(),
-				fmt.Sprintf(`Manifest source can be "%s" or "%s"`, config.ManifestSourceLocal.String(), config.ManifestSourceRemote.String()),
+				"Linking an existing app requires the app manifest source to be",
+				fmt.Sprintf("managed by %s.", config.ManifestSourceRemote.Human()),
```
zimeg:
```suggestion
				"Linking an existing app requires the app manifest source to be managed by",
				fmt.Sprintf("%s.", config.ManifestSourceRemote.Human()),
```

We might have a few extra characters inline to use for this sentence, but no blocker of course 👾

internal/config/manifest.go
```diff
@@ -35,6 +37,17 @@ func (ms ManifestSource) String() string {
 	return string(ms)
 }
 
+// Human returns the string value as a human-friendly name
+func (ms ManifestSource) Human() string {
+	switch ms {
+	case ManifestSourceLocal:
+		return fmt.Sprintf("project (%s)", ms.String())
```
zimeg: 🤔 This might be confusing, but I have a small favor towards quoting "project" while keeping `(local)` and `(remote)` in parentheses.

To me it'd read better in lines like:

```
App manifest source can be "project" (local) or "app settings" (remote)
```

zimeg: Edit: But perhaps we remove quotes in these lines altogether? Or handle the cases of "project" and "app settings" in configuration, but with warning to use the preferred `local` or `remote`?

I now lean towards removing quotes, but I'd love to know what is most clear to you!

internal/config/manifest.go
```diff
@@ -35,6 +37,17 @@ func (ms ManifestSource) String() string {
...
+	case ManifestSourceRemote:
+		return fmt.Sprintf("app settings (%s)", ms.String())
+	}
+	return ""
```
zimeg:
```suggestion
	return ms.String()
```

Ought not we find a confusing blank output here in some strange case 😉 

Edit: A test for this too would be superb but is also no blocker for me 🤖 ✨

cmd/app/link.go
```diff
@@ -173,14 +173,15 @@ func LinkExistingApp(ctx context.Context, clients *shared.ClientFactory, app *ty
...
 				" ",
-				fmt.Sprintf(style.Highlight(`Your manifest source is "%s"`), manifestSource.String()),
+				fmt.Sprintf(`App manifest source can be "%s" or "%s":`, config.ManifestSourceLocal.Human(), config.ManifestSourceRemote.Human()),
```
zimeg: 🗣️ Another comment suggests removing the quotations with `Human` but I'd be curious how this reads.

cmd/app/link.go
zimeg: Edit: For my own understanding, will it still be shown but instead much less often?

I can imagine changing the manifest source to `local` for automated productions then linking an app later, though this might not be the common case...

cmd/app/link.go
zimeg: Edit: I realize too this makes the message so much more clear! I had stumbled over why rejecting this prompt exited the command, but leading with "required" is so good 🧠

cmd/app/link.go
```diff
@@ -173,14 +173,15 @@ func LinkExistingApp(ctx context.Context, clients *shared.ClientFactory, app *ty
...
 				fmt.Sprintf("- %s: uses manifest from your project's source code for all apps", config.ManifestSourceLocal.String()),
```
zimeg, line 180:
```suggestion
				fmt.Sprintf("- %s: uses manifest from your project's source code on reinstallation", config.ManifestSourceLocal.String()),
```

📚 This might not be a good suggestion, but confusion could arise with when these updates are.

Perhaps this is alright though! Another command might make this difference more obvious in later iteration?

```sh
$ slack manifest update
```

zimeg: Edit: This comment makes me think "on update" is better wording, but again no blocker!

```suggestion
				fmt.Sprintf("- %s: uses manifest from your project's source code on update", config.ManifestSourceLocal.String()),
```

cmd/app/link.go
mwbrooks: Good question! I think the warning becomes less common after `bolt-install` and we could consider changing it to a message instead of a prompt.

The warning should be displayed when a project manages the app manifest (`manifest.source: "local"`) and an existing app is linked (`slack app link`). 

Right now, the warning prompts you to switch to a `remote` manifest source. I think we have an opportunity to make this more flexible. There are some cases where you may want to link an app while keeping your `local` manifest source.

Long story short, we'll want to think more about this as we make decisions around `bolt-install` 😆

cmd/app/link.go
mwbrooks: Great suggestion and this looks good on a regular 80 character terminal!
image

internal/config/manifest.go
mwbrooks: Great eye! Commit 3913b13 handles unknown manifest sources with your suggestion above and adds a test case 👌🏻

cmd/app/link.go
mwbrooks: I agree that this can be confusing. Unfortunately, the project-level manifest is used in many places and not just updates/re-installs. For example, `slack manifest validate` but also many areas of the internal-workings such as checking whether the app is hosted or not (a pre-run for some commands).

Looking ahead, this warning is probably going to be removed or reduced to a single line such as:
```bash
🤖 Some Title
   Reading app manifest from app settings
   Reading app manifest from project
```

So, we probably don't need to overthink the current warning too much.

cmd/app/link.go
mwbrooks: 🤔 I think for this specific line, we want to tell the developer what their current manifest source is set to. I stuck with the `.String()` but we could switch it to the `.Human()`.

internal/config/manifest.go
mwbrooks: Great suggestion. I think the `"project" (local)` format is an improvement! I'll update the tests and commit the changes! 🙇🏻 

image

image

mwbrooks: Commit 4eaea06 now quotes the human-friendly name and wraps the ID in parentheses.

cmd/app/link.go
mwbrooks: I think for now, let's leave it as-is. We already list the options above this text. Originally, the text was meant to explain what the project's current configuration is. Underneath, we then show the value in `config.json` and explain what the value will be changed to.
image

cmd/app/link.go
zimeg: @mwbrooks 🧠 ✨  Of course, how could I forget!

> Unfortunately, the project-level manifest is used in many places

I will celebrate this learning with good a fortune 👾

PR #92, opened 2025-05-14
feat: add 'app settings' command to open app settings webpage for a selected app

zimeg:
> **Changelog**
>
> The link between code of a project and configurations on app settings is now a commanded call with the latest `slack app settings` command.
>
> Bolt developers with a remote manifest source managed on app settings might find this useful for discovering new features or making updates otherwise.

### Summary

This PR adds the `app settings` command to open app settings for a selected app.

### Preview

help

https://github.com/user-attachments/assets/7c49b2c5-93d5-4d90-9fbe-b86415b2cb19

### Reviewers

Hello please feel free to test this with various apps, perhaps a:

- Bolt app **installed**
- Bolt app **uninstalled**
- Bolt project without apps
- Deno app (this errors)
- QA or `dev` instances

```
$ slack app settings
```

For the most adventurous, the `--app` flag can be attempted too.

### Notes

- ROSI apps are not supported in app settings at this time and will cause this command to **error**! The `--force` flag can be used to attempt this opening.

### Requirements


cmd/app/app.go
```diff
@@ -34,6 +34,7 @@ func NewCommand(clients *shared.ClientFactory) *cobra.Command {
 			{Command: "install", Meaning: "Install a production app to a team"},
 			{Command: "link", Meaning: "Link an existing app to the project"},
 			{Command: "list", Meaning: "List all teams with the app installed"},
+			{Command: "app settings", Meaning: "Open app settings in a web browser"},
```
mwbrooks, line 37: note: Foreseeing a future of alphabetically ordered examples prefixed with `app` 🔮

internal/slackerror/errors.go
```diff
@@ -42,6 +42,7 @@ const (
 	ErrAppExists                                     = "app_add_exists"
 	ErrAppFlagRequired                               = "app_flag_required"
 	ErrAppFound                                      = "app_found"
+	ErrAppHosted                                     = "app_hosted"
```
mwbrooks, line 45: note: I can see this being a useful error for us in the future!

cmd/app/settings.go
```diff
@@ -0,0 +1,121 @@
...
+
+func NewSettingsCommand(clients *shared.ClientFactory) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "settings [flags]",
+		Short: "Open app settings for configurations",
+		Long: strings.Join([]string{
+			"Open app settings to configure an application in a web browser.",
+			"",
+			"Discovering new features and customizing an app manifest can be done from this",
+			fmt.Sprintf("intuitive interface for apps with a \"%s\" manifest source.", config.ManifestSourceRemote.String()),
```
mwbrooks: suggestion: Rather than "intuitive" what about "web"? Web implies a UI, which is usually more intuitive, and feels a little less market-y.

```suggestion
			fmt.Sprintf("web interface for apps with a \"%s\" manifest source.", config.ManifestSourceRemote.String()),
```

image

cmd/app/settings.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/slackapi/slack-cli/internal/config"
+	"github.com/slackapi/slack-cli/internal/prompts"
+	"github.com/slackapi/slack-cli/internal/shared"
+	"github.com/slackapi/slack-cli/internal/slackerror"
+	"github.com/slackapi/slack-cli/internal/slacktrace"
+	"github.com/slackapi/slack-cli/internal/style"
+	"github.com/spf13/cobra"
+)
+
+var settingsAppSelectPromptFunc = prompts.AppSelectPrompt
```
mwbrooks, line 32: thought: **[Non-Blocker]** I'd like us to use this new command as a way to nudge out best practices for maintainability and testability. While we use the `xyzFunc` pattern for testability, it always feels like a hack.

I think the better approach would be dependency injection: `NewSettingsCommand(clients, prompts.AppSelectPrompt)`.

But, how do we design this to scale for all commands and `xyzFunc` workarounds?

Maybe we use `clients`? This would remove the global swap-a-roo and isolate it to the instance of clients.

```
// clients.go
clients.AppSelectPrompt = prompts.AppSelectPrompt

// clients_mock.go
clients.AppSelectPrompt = appSelectMock.AppSelectPrompt
```

I think this is out-of-scope of this PR, but perhaps we can follow-up with some approach in a future PR? We can take some time to see if it helps address other `xyzFunc` use-cases as well.

cmd/app/settings.go
```diff
@@ -0,0 +1,121 @@
...
+// appSettingsCommandPreRunE determines if the command can be run in a project
+func appSettingsCommandPreRunE(clients *shared.ClientFactory, cmd *cobra.Command, args []string) error {
+	ctx := cmd.Context()
+	err := cmdutil.IsValidProjectDirectory(clients)
+	if err != nil {
+		return err
+	}
+	if clients.Config.ForceFlag {
+		return nil
+	}
```
mwbrooks, line 77: suggestion: It might be worth adding a comment here to help other maintainers know why we return early when the `--force` flag is provided 🤔 
```suggestion
	// Allow the force flag to ignore hosted apps and try to open app settings
	if clients.Config.ForceFlag {
		return nil
	}
```

cmd/app/settings.go
```diff
@@ -0,0 +1,121 @@
...
+	if err != nil {
+		return err
+	}
+	host := clients.API().Host()
+	parsed, err := url.Parse(host)
+	if err != nil {
+		return err
+	}
+	parsed.Host = "api." + parsed.Host
+	settingsURL := fmt.Sprintf("%s/apps/%s", parsed.String(), app.App.AppID)
```
mwbrooks, line 109: praise: Very nice work 👏🏻

cmd/app/settings.go
```diff
@@ -0,0 +1,121 @@
...
+
+	clients.IO.PrintInfo(ctx, false, "\n%s", style.Sectionf(style.TextSection{
+		Emoji: "house",
+		Text:  "App Settings",
+		Secondary: []string{
+			settingsURL,
+		},
```
mwbrooks, line 116: suggestion: What do you think about adding output explaining the action being taken (opening the web browser)?
```suggestion
		Secondary: []string{
			"Opening default web browser:",
			settingsURL,
		},
```

☝🏻 You might have better wording than that.

cmd/app/settings.go
```diff
@@ -0,0 +1,121 @@
...
+	}))
+	clients.Browser().OpenURL(settingsURL)
+
+	clients.IO.PrintTrace(ctx, slacktrace.AppSettingsSuccess, settingsURL)
```
mwbrooks, line 120: praise: heh, you read my mind to add `settingsURL` 😉

cmd/app/settings_test.go
```diff
@@ -0,0 +1,155 @@
...
+							Settings: &types.AppSettings{FunctionRuntime: types.SlackHosted},
+						},
+					},
+					nil,
+				)
+				cm.AppClient.Manifest = manifestMock
+			},
+			ExpectedError: slackerror.New(slackerror.ErrAppHosted),
+		},
+		"requires an existing application": {
```
mwbrooks, line 112: note: Nice, I tested this manually and I'm happy to see you already added a test!

mwbrooks (approved):
✅ Amazing work @zimeg 🎉 

🧪 Manual testing is looking great and you've got very high test coverage on this command 📈 

📝 I left a few minor suggestions that you can take, leave, or riff with. I've also left a lengthy suggestion to follow-up with a 2nd PR that attempts to introduce a ~best~ better practice for our global `xyzFunc` swaps.

cmd/app/app.go
zimeg: As much as I appreicate alphabetics I was thinking these examples were ordered with expected steps in mind?

To me this tells an interesting story:

1. `install`
2. `link`
3. `list`
4. `settings`
5. `uninstall`
6. `delete`

Which is almost in alphabetics I realize now! So perhaps following "install" with "delete" isn't so wild 👾

cmd/app/settings.go
zimeg: @mwbrooks Good catch as always 🙏 ✨ 

For some reason I thought the "I" in "UI" stood for intuitive but let's let it mean "interface" here. It makes the double "u" more meaningful for web as well 😉

cmd/app/app.go
mwbrooks: Ahhhh, that's alright as well. For now, the order flows well and it's easy to see where `unlink` & `update` would land.

cmd/app/settings.go
zimeg: @mwbrooks Sounds good! I agree the placement is otherwise left to interpretation 📚 

I noticed too a test for this case is missing which doesn't seem ideal to me since this is not an erroring path that we support. Albeit the `--force` flag might make this more opinion, but let's avoid regressions with another test 🔍

cmd/app/app.go
zimeg: When `unlink` lands we should revisit this. Background thoughts are now nudging me towards preferring the more standard order 😳

cmd/app/settings.go
zimeg: 🤔 I might have preference against this since erroring cases might cause confusion that the standalone link saves?

The balance of text also feels lopsided but I don't think this is significant reason against this:

```
🏠 App Settings
   Opening default web browser:
   https://api.slack.com/apps/A08SPJ83H4L
```

zimeg: 📝 It does make me think `--no-browser` might be a flag to later support with the default behavior though!

cmd/app/settings.go
zimeg: @mwbrooks I'm so interested in this as well 🧠 ✨ 

I considered mocking the calls used in `AppSelectPrompt` to avoid this pattern altogether, but found that this adds so much overhead and perhaps mixes concerns...

Using `clients` might be a decent approach! It seems like we're settling on patterns of having identical function signatures within a command:

```go
	return appSettingsCommandRunE(clients, cmd, args)
},
```

But we've discussed avoiding `clients` for sake of more obvious usage of packages and dependencies. I'm wondering if within the command handler `clients` might be alright to keep, but for all other code following it should be avoided?

I'll hold off on making changes here so we can keep these patterns consistent for now, but I want to know what you think and I'm happy to follow up with changes we do decide on 🤖

zimeg:
@mwbrooks I am so excited for this to land in the next release! 🚀 ✨ 

Thanks for helping refine this to a better place as always. I left a few responses about perhaps much larger changes that this PR poked at, but for now I will merge this.

Follow up for particular flag options is also interesting to me, but also might be outside the scope of a first iteration.

cmd/app/settings.go
zimeg: The setups of `PrintTrace` continue to impress me! 👾

I'm hoping soon we can equip all commands with at least a simple "start" and "success" with follow ups in `@slack/cli-test` but for now I'm also so glad we can print a success message with additional detail 📚 ✨

cmd/app/settings.go
mwbrooks: Sounds good to me. Also `--no-browser` would be an appropriate future flag, since `activity --browser` already exists to open the activity in the browser.

cmd/app/settings.go
mwbrooks: Totally agree, good call on not adding more complexity in this PR. Let's discuss it offline to figure out what we want to try out!

cmd/app/settings.go
mwbrooks: It seems like a reasonable PR to add a "start" and "success" to each command and land the equivalent definitions in `@slack/cli-test`.

PR #95, opened 2025-05-15
refactor: enable linter rule QF1008 for 'Omit embedded fields from selector expression'

mwbrooks:
### Summary

This pull request enables the linter staticcheck rule [QF1008: 'Omit embedded fields from selector expression'](https://staticcheck.dev/docs/checks/#QF1008).

This rule ensures that we omit embedded fields from an expression. For example, the `type TriggerUpdateRequest` has the embedded field `TriggerRequest`:

```
type TriggerUpdateRequest struct {
	TriggerRequest
	TriggerID string `json:"trigger_id"`
}
```

And the following two expressions are equal in Golang:

* `triggerUpdateRequest.TriggerRequest.Inputs = something`
* `triggerUpdateRequest.Inputs = something`

Hedging a guess, we only need to use the embedded field when there is a naming conflict. 🤔 

### Requirements


internal/api/app_test.go
```diff
@@ -68,7 +68,7 @@ func TestClient_ExportAppManifest_Ok(t *testing.T) {
 	defer teardown()
 	result, err := c.ExportAppManifest(ctx, "token", "A0123456789")
 	require.NoError(t, err)
-	require.Equal(t, "example", result.Manifest.AppManifest.DisplayInformation.Name)
+	require.Equal(t, "example", result.Manifest.DisplayInformation.Name)
```
zimeg, line 71: This reads sososo much better to me 📚🥹✨

zimeg (approved):
@mwbrooks Thank so much for sharing the reasoning of this change! 

For me this makes it more clear that certain **structs** have fields attached and perhaps this makes for better objects overall? Let's revisit this after some time developing 👾

PR #99, opened 2025-05-19
refactor: enable linter rule QF1004 for 'Use strings.ReplaceAll when possible'

mwbrooks:
### Summary

This pull request enables the linter staticcheck rule [QF1004: 'Use strings.ReplaceAll instead of strings.Replace with n == -1'](https://staticcheck.dev/docs/checks/#QF1004).

### Requirements


internal/pkg/create/create.go
```diff
@@ -524,7 +524,7 @@ func InstallProjectDependencies(
 // In the future, this function can be extended to support other Git hosts, such as GitLab.
 // TODO, @cchensh, we should get prepared for other non-Git hosts and refactor the create pkg
 func generateGitZipFileURL(templateURL string, gitBranch string) string {
-	zipURL := strings.Replace(templateURL, ".git", "", -1) + "/archive/refs/heads/"
+	zipURL := strings.ReplaceAll(templateURL, ".git", "") + "/archive/refs/heads/"
```
zimeg, line 527:
```suggestion
	zipURL := strings.TrimSuffix(templateURL, ".git") + "/archive/refs/heads/"
```

🤔 This might've been an existing bug, but we might want to just trim the [suffix](https://pkg.go.dev/strings#TrimSuffix) here?

In this [example](https://go.dev/play/p/LcRRqB-FfPS) the template URL for the project is changed in an unexpected fashion:

```diff
- https://github.com/slack-samples/example.git-project.git
+ https://github.com/slack-samples/example-project
```

I don't think we have a test that covers these cases, but it might be nice to cover this case with these changes 🙏 ✨

mwbrooks: Wow, this is a great find! I doubt we have a test for this, so I'm leaning toward merging this PR (linter enablement) and follow-up with a new PR that fixes this issue w/ tests (fix: creating projects from a template url that contains .git).

mwbrooks:
Thanks for the quick review @zimeg. Slowly, we're enabling all of the linters 📝 

I'm going to merge this PR and follow-up with a 2nd that fixes the [issue that you noticed](https://github.com/slackapi/slack-cli/pull/99#discussion_r2096242693). I'm doing this so that we have CHANGELOG entries for each (enabling a new linter and fixing an issue). Plus, we can take time to ensure that there are proper tests. ✌🏻

PR #100, opened 2025-05-19
ci: pin action hashes and escape variables with minimum permission

zimeg:
### Summary

This PR uses the wonderful [`zizmor`](https://github.com/zizmorcore/zizmor) tool to audit our own workflows and [`pinact`](https://github.com/suzuki-shunsuke/pinact) for pinned versioning 👾

### Reviewers

A similar audit can be performed with the zizmor tool:

```
$ zizmor .
...
No findings to report. Good job! (1 ignored, 7 suppressed)
```

> The suppressed findings are expected permission blocks at the top-level of a workflow, but we set this for each job.

### Notes

Most changes I hope are repetitive, but I will comment on the more significant ones!

### Requirements

### DIFF
--- a/.github/workflows/delete-pr-build-on-close.yml
+++ b/.github/workflows/delete-pr-build-on-close.yml
@@ -13,7 +13,7 @@ name: Delete pre-release when a branch is deleted
 # See https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-pull_request-workflow-when-a-pull-request-merges
 # See https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request_target
 on:
-  pull_request_target:
+  pull_request_target: # zizmor: ignore[dangerous-triggers]
     types:
       - closed
 
--- a/.github/workflows/dependencies.yml
+++ b/.github/workflows/dependencies.yml
@@ -12,17 +12,18 @@ jobs:
     steps:
       - name: Gather credentials
         id: credentials
-        uses: actions/create-github-app-token@v2.0.6
+        uses: actions/create-github-app-token@df432ceedc7162793a195dd1713ff69aefc7379e # v2.0.6
         with:
           app-id: ${{ secrets.GH_APP_ID_RELEASER }}
           private-key: ${{ secrets.GH_APP_PRIVATE_KEY_RELEASER }}
       - name: Checkout the repo
-        uses: actions/checkout@v4.2.2
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
+          persist-credentials: true
           ref: main
           token: ${{ steps.credentials.outputs.token }}
       - name: Install Golang
-        uses: actions/setup-go@v5.5.0
+        uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5 # v5.5.0
         with:
           go-version: "stable"
       - name: Get the latest version
--- a/.github/workflows/license_check.yml
+++ b/.github/workflows/license_check.yml
@@ -13,15 +13,10 @@ jobs:
     permissions:
       contents: read
     steps:
-      - uses: actions/checkout@v4.2.2
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
-          # Number of commits to fetch. 0 indicates all history for all branches and tags.
-          # Default: 1
-          # TODO - We should not fetch all history.
-          #        But we need to fetch the latest tag to run `make test` and `make build`.
-          #        This is a workaround until we fetch the latest tag using the Makefile.
-          fetch-depth: 0
+          persist-credentials: false
       - name: Check license headers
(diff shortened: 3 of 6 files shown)


.github/workflows/delete-pr-build-on-close.yml
```diff
@@ -13,7 +13,7 @@ name: Delete pre-release when a branch is deleted
 # See https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-pull_request-workflow-when-a-pull-request-merges
 # See https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request_target
 on:
-  pull_request_target:
+  pull_request_target: # zizmor: ignore[dangerous-triggers]
```
zimeg, line 16: Heh the comments above explain the reason we use `pull_request_target` here but this comment makes a hopeful assumption that `zizmor` can be added to pipelines for future audits 🤖 ✨

.github/workflows/license_check.yml
```diff
@@ -13,15 +13,10 @@ jobs:
...
     steps:
-      - uses: actions/checkout@v4.2.2
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
-          # Number of commits to fetch. 0 indicates all history for all branches and tags.
-          # Default: 1
-          # TODO - We should not fetch all history.
-          #        But we need to fetch the latest tag to run `make test` and `make build`.
-          #        This is a workaround until we fetch the latest tag using the Makefile.
-          fetch-depth: 0
```
zimeg, line 23: 🪓 In this file we do not need to checkout all things!

.github/workflows/tests.yml
```diff
@@ -18,30 +18,23 @@ jobs:
...
       contents: read
     steps:
-      - uses: actions/checkout@v4.2.2
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
-          # Number of commits to fetch. 0 indicates all history for all branches and tags.
-          # Default: 1
-          # TODO - We should not fetch all history.
-          #        But we need to fetch the latest tag to run `make test` and `make build`.
-          #        This is a workaround until we fetch the latest tag using the Makefile.
```
zimeg, line 27: 🪓 In this file we **do** need to fetch tags from past commits. This might be a lot of commits, and this `TODO` can be deceiving...

zimeg (commented):
💌 Notes on reasons that certain lines were changed for the most curious and kind reviewers!

.github/workflows/dependencies.yml
```diff
@@ -12,17 +12,18 @@ jobs:
     steps:
       - name: Gather credentials
         id: credentials
-        uses: actions/create-github-app-token@v2.0.6
+        uses: actions/create-github-app-token@df432ceedc7162793a195dd1713ff69aefc7379e # v2.0.6
```
mwbrooks, line 15: note: Whoa, this is intense! I understand from a security point-of-view (pinning to a dash instead of a tag). Thanks for adding the comment!

question: How will dependabot handle this going forward? Will it try to use a tag?

.github/workflows/sync-docs-from-cli-repo.yml
```diff
@@ -41,7 +43,9 @@ jobs:
 
       - name: Update docs in docs site repo
         run: |
-          rsync -av --delete ./docs/ ./docs_repo/content/${{ github.event.repository.name }}/
+          rsync -av --delete ./docs/ "./docs_repo/content/$REPO/"
+        env:
+          REPO: ${{ github.event.repository.name }}
```
mwbrooks, line 48: praise: Clean! ✨

mwbrooks (approved):
✅ LGTM! Thanks for helping to boost our security.

❓ I left one question around dependabot and how we want to handle its PRs going forward, but it's non-blocking to merging this one!

.github/workflows/dependencies.yml
zimeg: @mwbrooks I had the exact same questions! It's nice that we can keep this comment inline for somewhat reasonable understandings - I cannot remember hashes so well... 🔍 

> How will dependabot handle this going forward? Will it try to use a tag?

AFAICT tags from official **releases** will be used in these updates and the comment that follows will match. This should match the current updating events, but instead with commit details!

An [example](https://github.com/slackapi/slack-github-action/pull/443/files) update has shown this to be alright and this [blog](https://github.blog/changelog/2022-10-31-dependabot-now-updates-comments-in-github-actions-workflows-referencing-action-versions/) post shared a few more detail 🎃

.github/workflows/sync-docs-from-cli-repo.yml
zimeg: 📝 Quick note, I find this surprisingly **more** portable too!

I don't believe the handlebar notation is supported in `bash`, but I do know environment variables are 😉

zimeg:
@mwbrooks Thanks for the quick and curious review! 🔏 ✨ 

Explorations with tags in this project has now made me warried of using these as immutable values... But we can guard against that with this change! I will merge this now.

PR #103, opened 2025-05-21
fix: removing .git from inside git template urls

mwbrooks:
### CHANGELOG

> Fixes a bug when creating a project using a git repo template. If the template includes `.git` inside of the URL, then the template will fail to be cloned. For example, `slack create --template https://github.com/slack-samples/example.git.project.git`.

### Summary

This pull request is a follow-up of https://github.com/slackapi/slack-cli/pull/99#discussion_r2096242693 that fixes a bug when creating a project from a git repo template. If the repo name included `.git` inside of the name (and, optionally as the extension), then both `.git` references would be removed from the URL.

h/t @zimeg for noticing this bug! 🐛

### Requirements


internal/pkg/create/create_test.go
```diff
@@ -128,6 +128,24 @@ func Test_generateGitZipFileURL(t *testing.T) {
 				httpClientMock.On("Get", mock.Anything).Return(nil, fmt.Errorf("HTTPClient error"))
 			},
 		},
+		"Returns the zip URL with .git suffix removed": {
+			templateURL: "https://github.com/slack-samples/deno-starter-template.git",
```
zimeg, line 132: 📣 This IMO is the case we ought consider as "correct" for the `--template` flag since it appears on the "HTTPS" tab of the `Code` dropdown from the sample page:

🔗 https://github.com/slack-samples/deno-starter-template
:tv: https://github.com/slack-samples/deno-starter-template.git

FWIW we do not use the `.git` prefix in sample names but might recommend using the entire URL ending with `.git` if this is encountered later... Or recommend changing the sample name altogether, though I understand this can cause hassle 😉

zimeg (approved):
@mwbrooks Such a clean change - great call saving this for a follow up with tests! 🙏 ✨ 

I left rambling thoughts on how this might snag an edge later, but we can explore this more if an issue appears perhaps 🔍

PR #110, opened 2025-05-29
refactor: loop over a range of integers starting from zero in shorthand

zimeg:
### Summary

This PR replaces longform loop expressions with a shorthand range when iterating across integers starting from zero.

No change to functionalities!

### Notes

- 📚 https://go.dev/wiki/Range
- AFAICT the following is not supported in `go`. Perhaps for good reason I believe:

```go
for i := range 1..100 { /* */ }
```

### Requirements


mwbrooks (approved):
✅ Looks good to me! I never knew that we use the `range` syntax with a single integer!

PR #124, opened 2025-06-10
build: sign macos binaries for arm and amd architectures and universal packagings

zimeg:
### Summary

This PR attempts to sign all macOS binaries before packaging these into development and latest bundles for a tarball.

### Notes

A feature build will be tagged in this PR to test this!

### Requirements

* [x] I've read and understood the [Contributing Guidelines](/blob/main/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).

PR #114, opened 2025-06-03
feat: package a development and latest build for darwin architectures

zimeg:
> **Changelog**
>
> Specific builds of the development and latest editions for the different Darwin architectures are now packaged.

### Summary

This PR adds the following archives and tests expected executables exist:

- `slack_cli_dev_macOS_amd64.tar.gz`
- `slack_cli_dev_macOS_arm64.tar.gz`
- `slack_cli_latest_macOS_amd64.tar.gz`
- `slack_cli_latest_macOS_arm64.tar.gz`

### Requirements

* [x] I've read and understood the [Contributing Guidelines](/blob/main/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).

PR #101, opened 2025-05-20
build: package for apple intel and apple silicon machines alongside universal binaries as fallback

zimeg:
> **Changelog**
>
> Developers developing on Apple silicon can now upgrade to a native build supporting the `arm64` architecture. Installing an installation with the install script will also detect the matching build to use.

### Summary

This PR builds `darwin` universal binaries along with specific builds for both `amd64` and now `arm64`! 🍎 ✨

### Reviewers

Check the installation scripts installs:

```sh
$ ./scripts/install.sh -v 3.2.0 lack-dev  # Confirm the current release installs
$ ./scripts/install.sh -v 3.3.0 lack-dev  # Confirm the next version is attempted
...
https://downloads.slack-edge.com/slack-cli/slack_cli_3.3.0_macOS_arm64.tar.gz
#=O#-  #    #
curl: (22) The requested URL returned error: 403
```

Confirm automatic upgrades update:

```sh
$ go build -ldflags="-s -w -X 'github.com/slackapi/slack-cli/internal/pkg/version.Version=v3.0.0'" -o bin/slack
$ lack version
Using lack v3.0.0
$ lack upgrade
...
Architecture: arm64
OS: darwin
Downloading slack_cli_3.1.0_macOS_arm64.zip
Downloading cli from https://downloads.slack-edge.com/slack-cli/slack_cli_3.1.0_macOS_arm64.zip
```

Please note "arm64" might be "amd64" for some setups. An apple computer is required too to test this 👾 

### Notes

- Universal binaries are installed as `macOS_64-bit` for backwards compatibilities with improved support. Rosetta is no long required.
- The previous `macOS_64-bit` installation is moved to `macOS_amd64`.
- A new packaging for the M-series now exists as `macOS_arm64`!

### Requirements

* [x] I've read and understood the [Contributing Guidelines](/blob/main/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).

PR #126, opened 2025-06-12
feat: include target architecture of compiled builds in logstash events

zimeg:
### Summary

This PR adds the target architecture of compiled builds in events sent to logstash for errors following #101 #114 #124.

### Reviewers

Confirm this build includes the target machine's architecture with the following command:

```sh
$ slack version --verbose
...
FlushToLogstash will POST https://dev.slackb.com/events/cli payload: [{"context":{"arch":"arm64","...":"..."}]
``` 

### Requirements


internal/tracking/tracking.go
```diff
@@ -193,6 +193,7 @@ func (e *EventTracker) FlushToLogstash(ctx context.Context, cfg *config.Config,
 			CLIVersion:       versionString,
 			Host:             ioutils.GetHostname(),
 			OS:               runtime.GOOS,
+			Arch:             runtime.GOARCH,
```
zimeg, line 196: 📣 TIL that `GOARCH` is set at **compile** time and not runtime, so this matches the installed build:

🔗 https://pkg.go.dev/runtime

> GOARCH, GOOS, and GOROOT are recorded at compile time and made available by constants or functions in this package, but they do not influence the execution of the run-time system.

zimeg (commented):
🪵 Words of findings and thought in these changes for the kind reviewers!

internal/tracking/tracking.go
mwbrooks: Wow! I was about to point that this is may report the user's system architecture instead of the binary's architecture. Thanks for digging into it and finding out that it uses the binary's arch!

zimeg:
@mwbrooks Thanks for checking the releases! I didn't think to also test with the universal build but found `arm64` on my machine! 📠 

The `amd64` build however outputs what we're both expecting so I will merge this.

But beforehand, a thankful TIL about this most helpful command - I have been using System Preferences:

```
$ xattr -dr com.apple.quarantine ./bin/slack
```

PR #130, opened 2025-06-16
refactor(config): destructify config.GetProjectDirPath(fs, os)

mwbrooks:
### Summary

_Proposal to refactor the `package config` from structs to standalone functions._

This pull request refactors the function `GetProjectDirPath` in `package config` from a struct to a standalone function. This is a small example to showcase the changes involve in the factor and explain the benefits. If we like this direction, then I'll follow-up with large pull requests that make the changes for all of `internal/config/project.go`.

**Before:**
```golang
err := clients.Config.ProjectConfig.GetProjectDirPath()
```

**After:**
```golang
err := config.GetProjectDirPath(fs, os)
```

**Benefits:**
* More idiomatic Golang
* Separate data (`fs`) from behaviour (the function)
* Makes dependencies explicit in the function signature
* Follows Golang's preference for small, focused functions over methods on large structs
* More accurate tests that use a memory-based file system that represents the production environments
* Lighter `clients` struct overall

**Drawbacks:**
* Must update tests to mock the memory-based file system (`fs.MemMapFs`) instead of the methods (`projectConfigMock.On("GetManifestSource", mock.Anything,)...`)
* Challenging to untangle some of the `package config` methods, such as `Cache()` and `SetSurveyConfig`

### Requirements


internal/config/project.go
```diff
@@ -58,7 +58,6 @@ type ProjectConfigManager interface {
 	ReadProjectConfigFile(ctx context.Context) (ProjectConfig, error)
 	WriteProjectConfigFile(ctx context.Context, projectConfig ProjectConfig) (string, error)
 	ProjectConfigJSONFileExists(projectDirPath string) bool
-	GetProjectDirPath() (string, error)
```
mwbrooks, line 61: note: Removing `GetProjectDirPath` from the interface.

internal/config/project.go
```diff
@@ -308,21 +307,21 @@ func (c *ProjectConfig) ProjectConfigJSONFileExists(projectDirPath string) bool
 
 // GetProjectDirPath returns the path to the project directory or an error if not a Slack project
 // TODO(@mbrooks) Standardize the definition of a validate project directory and merge with `cmdutil.ValidProjectDirectoryOrExit`
-func (c *ProjectConfig) GetProjectDirPath() (string, error) {
```
mwbrooks, line 311: note: Changing `GetProjectDirPath` from a method on the struct to a standalone function. Now the dependencies are injected as arguments instead of accessed as private struct fields.

internal/config/project_mock.go
```diff
@@ -83,11 +83,6 @@ func (m *ProjectConfigMock) ProjectConfigJSONFileExists(projectDirPath string) b
 	return args.Bool(0)
 }
 
-func (m *ProjectConfigMock) GetProjectDirPath() (string, error) {
-	args := m.Called()
-	return args.String(0), args.Error(1)
-}
-
```
mwbrooks, line 90: note: Removing the mock. We don't use this mock, but for some future methods, it will require us set the correct data in the memory-based file system and then read it back to confirm it.

internal/slackdeps/os_mock.go
```diff
@@ -25,7 +25,7 @@ import (
 const MockHomeDirectory = "/Users/user.name"
 
 // MockWorkingDirectory is the default returned by Getwd().
-const MockWorkingDirectory = "/Users/user.name/app"
+const MockWorkingDirectory = "/Users/user.name/project"
```
mwbrooks, line 28: note: I changed this to `/project` because it was a little weird to read in test output. We're creating a new project, not a Slack app. So it makes sense to have the directory called project.

internal/slackdeps/os_mock.go
```diff
@@ -50,6 +50,7 @@ func (m *OsMock) AddDefaultMocks() {
 	m.On("UserHomeDir").Return(MockHomeDirectory, nil)
 	m.On("GetExecutionDir").Return(MockHomeDirectory, nil)
 	m.On("SetExecutionDir", mock.Anything)
+	m.On("IsNotExist", nil).Return(false)
```
mwbrooks, line 53: note: In order for tests to pass, we need our mock for `os.IsNotExist(err)` to return false when `err = nil`. In most cases, we don't need a mock here and could use the real `IsNotExist` 🤔

internal/config/project.go
zimeg: Huge unlock in making it clear what's **needed** for these functions to complete! I like that "standard" dependencies are no longer kept within a struct here.

Organizing related functions into packages still seems ideal for separating most concerns. Perhaps a returned `configuration` object has attached methods? But I'm so open to going back and forth until a good balance is found 🙏 ✨

internal/config/project_mock.go
zimeg: @mwbrooks This is an excellent change from testing "mocks" to "fixtures" where we can exercise all of the logic with a certain set of known files! 🧪

internal/config/project_test.go
```diff
@@ -796,6 +794,6 @@ func addProjectMocks(t require.TestingT, fs afero.Fs) {
 	require.NoError(t, err)
 
 	// Fixture: project/.slack/hooks.json file
-	err = afero.WriteFile(fs, GetProjectHooksJSONFilePath(slackdeps.MockWorkingDirectory), []byte("{}\n"), 0600)
+	err = afero.WriteFile(fs, GetProjectHooksJSONFilePath(slackdeps.MockWorkingDirectory), []byte("{}\n"), 0644)
```
zimeg, line 797: 👁️‍🗨️ ✨

internal/slackdeps/os_mock.go
zimeg: 🗣️ Past readings have made me a believer in mock data matching what might be expected in production! Nice change once more.

internal/slackdeps/os_mock.go
zimeg: > In most cases, we don't need a mock here and could use the real `IsNotExist`

Heh I had this exact same though! Now that we're moving towards fixtures that might be a more stable change to avoid mocking strangeness from odd defaults. No blocker for these changes though! 🚢

internal/config/project.go
mwbrooks: Good to know that this is resonating with you as well @zimeg! There's definitely some ergonomics that we'll need to figure out, whether returning a struct or better naming conventions. But at least the general direction sounds positive!

internal/config/project_mock.go
mwbrooks: Oooo, I like how you phrased that. From testing "mocks" to testing "fixtures." The fixtures feel more real, even though they require more setup and the unit tests run deeper down the stack.

internal/slackdeps/os_mock.go
mwbrooks: Great point that as we use more fixtures instead of mocks, we may not need to mock a `IsNotExist` scenario. Instead, we can setup the fixtures to satisfy that edge-case.

PR #132, opened 2025-06-19
fix: 'run' exiting on disconnected api errors when streaming activity logs

### LINKED ISSUE #128: bug: unexpected daemon process
### Description

When using the Slack CLI to run a bolt for python project, the CLI process will sometimes terminate without cleaning up the Bolt Python process. This leaves a daemon type python process running on my machine.

This can be a pain point, I've accidentally created multiple instances of socket mode apps leaving the Slack to dispatch events randomly between them.

### Version

Using lack v3.3.0

### OS Info

ProductName:            macOS
ProductVersion:         15.5
BuildVersion:           24F74
Darwin Kernel Version 24.5.0: Tue Apr 22 19:54:49 PDT 2025; root:xnu-11417.121.6~2/RELEASE_ARM64_T6000

#### Steps to reproduce:

1. `slack create my-test-app -t https://github.com/slack-samples/bolt-python-starter-template`
2. `slack run`
3. Once the app is running, quickly disconnect and reconnect your machine from the internet
4. The Slack CLI process will terminate with the following error
```zsh
🚫 Failed to stream the most recent activity logs (streaming_activity_logs_error)
   HTTP request failed (http_request_failed)
```
5. The python process should reconnect the socket connection
6. Triggering an event to be dispatch to your app like loading the app_home will log messages in the terminal
7. You can use the activity monitor to clean up the demon process

### Expected result:

When the Slack CLI process terminates I expect that child processes it created are cleaned up as well

### Actual result:

Its seems like some child processes created by the CLI are not clean up properly

## Note

This may also be an issue with the hooks implementations

### Requirements

* [x] I've read and understood the [Contributing guidelines](/blob/main/.github/contributing.md) and have done my best effort to follow them.
* [x] I've searched for any related issues and avoided creating a duplicate issue.

zimeg:
> **Changelog**
>
> We fixed a bug where the `run` command exited with an error if the activity logs failed to stream to now attempt retries for those missed logs.

### Summary

This PR avoids exiting on an error when streaming the activity logs for the `run` command to workaround or fix #128! 👾 

### Preview

The following video shows a disconnected and reconnected internet connection and a healing socket connection:

- `0:07`: The app has started and connected
- `0:19`: The internet connection is turned off
- `0:52`: A new web socket attempts to connect
- `1:15`: Activity logs begin to fail retries - **New**
- `1:37`: The internet connection is turned on
- `1:40`: A single socket mode connection is created with streaming logs

https://github.com/user-attachments/assets/a82f38ce-cb6d-4fc7-b0f1-f1d2bdf15136

### Notes

- This change doesn't fix the cause of persisting processes with a cancelled context or exit without error, but we instead avoid both to require interrupts to or process exits from the **app**!
- Following the note above, the "Closing due to inactivity. Au revoir!" might need to include an error instead. Perhaps we save this or canceling the context on exits for a follow up PR? 🤖 

### Requirements


internal/api/activity.go
```diff
@@ -112,13 +112,13 @@ func (c *Client) Activity(ctx context.Context, token string, activityRequest typ
 
 	b, err := c.get(ctx, url, token, "")
 	if err != nil {
-		return ActivityResult{}, errHTTPRequestFailed.WithRootCause(err)
+		return ActivityResult{}, slackerror.New(slackerror.ErrHTTPRequestFailed).WithRootCause(err)
```
zimeg, line 115: 📣 We create a new error here to avoid appending different repeated root causes to this same error:

```
HTTP request failed (http_request_failed)
   Get "https://slack.com/api/apps.activities.list?app_id=A0923PDU39B&limit=100&min_log_level=info&min_date_created=1750312064888018": dial tcp: lookup slack.com: no such host

   Get "https://slack.com/api/apps.activities.list?app_id=A0923PDU39B&limit=100&min_log_level=info&min_date_created=1750312064888018": dial tcp: lookup slack.com: no such host

   Get "https://slack.com/api/apps.activities.list?app_id=A0923PDU39B&limit=100&min_log_level=info&min_date_created=1750312064888018": dial tcp: lookup slack.com: no such host

   Get "https://slack.com/api/apps.activities.list?app_id=A0923PDU39B&limit=100&min_log_level=info&min_date_created=1750312064888018": dial tcp: lookup slack.com: no such host
```

internal/pkg/platform/activity.go
```diff
@@ -88,11 +88,12 @@ func Activity(
...
-	if err != nil {
-		return err
-	}
 
+	// Return an error from the activity API if the logs are not being tailed
 	if !args.TailArg {
+		if err != nil {
+			return err
+		}
 		return nil
```
zimeg, line 97: 🔍 Small preference to make the `nil` return more clear instead of returning `err` in both cases, even if it too is `nil`.

internal/pkg/platform/activity_test.go
```diff
@@ -148,11 +200,33 @@ func TestPlatformActivity_StreamingLogs(t *testing.T) {
 				}()
 				return ctx
 			},
+			ExpectedError: nil,
 			ExpectedAsserts: func(t *testing.T, ctx context.Context, cm *shared.ClientsMock) {
 				// with the above polling/canceling setup, expectation is activity called only once.
 				cm.API.AssertNumberOfCalls(t, "Activity", 1)
 			},
 		},
+		"should return nil if TailArg is set and activity request fails while polling": {
```
zimeg, line 209: 🧪 The added `nil` values for `ExpectedError` above are included to make these cases more explicit instead of preferring default behavior.

For this test I think it's useful, but I'm of course open to reverting this!

zimeg (commented):
📝 A few of the thoughts happening with some of these changes for the most kind reviewers!

internal/pkg/platform/activity.go
mwbrooks: Thanks, I prefer this clarify as well!

internal/pkg/platform/activity_test.go
mwbrooks: No need to revert, I like this as well!

mwbrooks (approved):
✅ Nice improvement @zimeg!

🧪 During testing, I noticed that a few times I'd receive a `connection reset by peer` error when re-connecting the wifi network. However, it doesn't always happen.

image

🎥 Below is a video for anyone who wants to see the original error that this PR fixes:

https://github.com/user-attachments/assets/36337a97-f395-49a1-8c37-c1c522cd873c

zimeg:
🔍 A CLI error has since appeared for me after running the app for a while and repeats every few seconds but without exiting which I think is expected:

```
[2025-06-24 18:39:00] Setting up a ticker to poll activity on a 3s interval
...
[INFO]  bolt-app ⚡️ Bolt app is running!
[2025-06-24 22:10:18] HTTP Request: GET https://slack.com/api/apps.activities.list?app_id=A093KR40YEL&limit=100&min_log_level=info&min_date_created=1750815539904263 HTTP/1.1
[2025-06-24 22:10:18] HTTP Request Body:
[2025-06-24 22:10:18] <no body>
[2025-06-24 22:10:18] HTTP Response Status: HTTP/1.1 200 OK
[2025-06-24 22:10:18] HTTP Response Body:
[2025-06-24 22:10:18] <no body>
[2025-06-24 22:10:18] The following error was returned by the apps.activities.list Slack API method
[2025-06-24 22:10:18]
[2025-06-24 22:10:18] 🚫 Your access token has expired (token_expired)
[2025-06-24 22:10:18]
[2025-06-24 22:10:18] 💡 Suggestion
[2025-06-24 22:10:18]    Use the command `lack login` to authenticate again
...
[2025-06-24 22:14:18] HTTP Request: GET https://slack.com/api/apps.activities.list?app_id=A093KR40YEL&limit=100&min_log_level=info&min_date_created=1750815539904263 HTTP/1.1
[2025-06-24 22:14:18] HTTP Request Body:
[2025-06-24 22:14:18] <no body>
[2025-06-24 22:14:18] HTTP Response Status: HTTP/1.1 200 OK
[2025-06-24 22:14:18] HTTP Response Body:
[2025-06-24 22:14:18] <no body>
[2025-06-24 22:14:18] The following error was returned by the apps.activities.list Slack API method
[2025-06-24 22:14:18]
[2025-06-24 22:14:18] 🚫 Your access token has expired (token_expired)
[2025-06-24 22:14:18]
[2025-06-24 22:14:18] 💡 Suggestion
[2025-06-24 22:14:18]    Use the command `lack login` to authenticate again
[WARN]  socket-mode:SlackWebSocket:1 A pong wasn't received from the server before the timeout of 5000ms!
[WARN]  socket-mode:SlackWebSocket:1 A ping wasn't received from the server before the timeout of 30000ms!
[DEBUG]  socket-mode:SlackWebSocket:1 Sending close frame (status=1000).
[DEBUG]  socket-mode:SlackWebSocket:1 Sending close frame (status=1000).
[DEBUG]  socket-mode:SlackWebSocket:1 WebSocket close frame received (code: 1006, reason: )
[DEBUG]  socket-mode:SlackWebSocket:1 Terminating WebSocket (close frame received).
[DEBUG]  socket-mode:SocketModeClient:0 Before trying to reconnect, this client will wait for 5000 milliseconds
```

I'm wondering if an expired token for `apps.activities.list` isn't being rotated or used in new requests?  👾

PR #135, opened 2025-06-19
docs: surround remediation flags with backticks for prompted feedback

lukegalbraithrussell:
### Summary

This error has unescaped `<` in its description which our docs sites does not like.

### Requirements

### DIFF
--- a/internal/slackerror/errors.go
+++ b/internal/slackerror/errors.go
@@ -729,7 +729,7 @@ Otherwise start your app for local development with: %s`,
 		Code:    ErrFeedbackNameRequired,
 		Message: "The name of the feedback is required",
 		Remediation: strings.Join([]string{
-			"Please provide a --name <string> flag or remove the --no-prompt flag",
+			"Please provide a `--name <string>` flag or remove the `--no-prompt` flag",
 			fmt.Sprintf("View feedback options with %s", style.Commandf("feedback --help", false)),
 		}, "\n"),
 	},


internal/slackerror/errors.go
```diff
@@ -729,7 +729,7 @@ Otherwise start your app for local development with: %s`,
 		Code:    ErrFeedbackNameRequired,
 		Message: "The name of the feedback is required",
 		Remediation: strings.Join([]string{
-			"Please provide a --name <string> flag or remove the --no-prompt flag",
+			"Please provide a `--name <string>` flag or remove the `--no-prompt` flag",
```
mwbrooks, line 732: note: It's a new dimension for us to write error messages with markdown docs rendering in mind. We'll also want to ensure this always looks good on the terminal. Back ticks can be a dangerous character in the terminal when copy & pasted, but I agree that it reads well.

question: We may be able to update our `slack docs` command to render `&nbsp;` instead of `<` rather than adding backticks to our error messages.

lukegalbraithrussell: oh i just went ahead and added the backticks because i saw them in other errors -- if it creates issues i can add an exception on our docs-cleaning scripts instead. i was trying to avoid adding hardcoded examples in that for maintenance reasons, but if it actually affects terminal output possibilities then it seems like the better option!

zimeg: 👋 Adding to @mwbrooks suggestion - I lean towards improving `docgen` outputs with expected escaping for these cases!

But I do think this change makes the remediation more clear, FWIW.

zimeg (approved):
@lukegalbraithrussell LGTM! Thanks for finding and following up with this fix! 📚 ✨

internal/slackerror/errors.go
mwbrooks: @lukegalbraithrussell Totally cool! I think we should merge this PR because it immediately fixes the docs experience. We can create a task to improve the `docgen` command if we want to avoid adding back ticks to every occurrence of `<` and `>`. :shipit:

PR #138, opened 2025-06-23
refactor(config): destructify project-level getter and setters in package config

mwbrooks:
### Summary

This pull request continues PR https://github.com/slackapi/slack-cli/pull/130 by de-struct-ifying the following project-level methods in `package config`:

* `config.ProjectConfigJSONFileExists(...)`
* `config.WriteProjectConfigFile(...)`
* `config.ReadProjectConfigFile(...)`
* `config.SetManifestSource(...)`

I've chosen these methods because their changes are relatively straight-forward and repetitive. The remaining methods require wider changes, so I'd like to isolate them to smaller, future PRs.

### Changes

A common refactor looks like this:

[Remove method from the interface](https://github.com/slackapi/slack-cli/commit/31fed260b3a945f5c2b2f56e54a229fa555bd9f5#diff-c8411ba332284a90304ba272d45057adba0bbd1fdd50654244e41af12c9bc419L61):

```diff
- GetProjectDirPath() (string, error)
```

[Update method to be a standalone function with dependencies as arguments](https://github.com/slackapi/slack-cli/commit/31fed260b3a945f5c2b2f56e54a229fa555bd9f5#diff-c8411ba332284a90304ba272d45057adba0bbd1fdd50654244e41af12c9bc419L311):

```diff
- func (c *ProjectConfig) GetProjectDirPath() (string, error) {
+ func GetProjectDirPath(fs afero.Fs, os types.Os) (string, error) {
```

[Update references to use the standalone function](https://github.com/slackapi/slack-cli/commit/31fed260b3a945f5c2b2f56e54a229fa555bd9f5#diff-c8411ba332284a90304ba272d45057adba0bbd1fdd50654244e41af12c9bc419L248):

```diff
- projectDirPath, err := c.GetProjectDirPath()
+ projectDirPath, err := GetProjectDirPath(c.fs, c.os)
```

[Remove method from the project mocks](https://github.com/slackapi/slack-cli/commit/31fed260b3a945f5c2b2f56e54a229fa555bd9f5#diff-e4d3866f77409dfb9008298bc9a2c56b9ef701361760c22fae7edf7002feabe5L86-L90):

```diff
- func (m *ProjectConfigMock) GetProjectDirPath() (string, error) {Add commentMore actions
- 	args := m.Called()
- 	return args.String(0), args.Error(1)
- }
```

[Update tests to use filesystem mock instead of project-config mock](https://github.com/slackapi/slack-cli/pull/138/commits/5ea82cca46f2b1be7c8ed931d9ca2f1c5e32378e#diff-a6e372a33bd96c3d43a02322c5801928e1760834c1df44ace4a4a3e4b5c445ccL60):
```diff
- projectConfigMock := &config.ProjectConfigMock{}
- projectConfigMock.On("GetProjectDirPath").Return(slackdeps.MockWorkingDirectory)
+ slackmock.CreateProject(t, ctx, clientsMock.Fs, clientsMock.Os, slackdeps.MockWorkingDirectory)
```

### Requirements


internal/config/project.go
```diff
@@ -237,24 +233,24 @@ func (c *ProjectConfig) SetSurveyConfig(ctx context.Context, name string, survey
...
 	span, _ = opentracing.StartSpanFromContext(ctx, "ReadProjectConfigFile")
 	defer span.Finish()
 
 	var projectConfig ProjectConfig
 
-	projectDirPath, err := GetProjectDirPath(c.fs, c.os)
+	projectDirPath, err := GetProjectDirPath(fs, os)
 	if err != nil {
 		return projectConfig, err
 	}
```
zimeg, line 246: 👁️‍🗨️ This might be because I'm used to the **before** version of this code, but to me this:

```go
projectConfig, err := clients.Config.ProjectConfig.ReadProjectConfigFile(ctx)
```

Hid the magic of how the project path was found. I realize now we're doing the same thing with arguments shared instead, but it wasn't obvious how `fs` and `os` were used to read the configuration to me...

```go
projectConfig, err := config.ReadProjectConfigFile(ctx, clients.Fs, clients.Os)
```

I do think this is good improvement for testing and using this package overall and it's not so confusing IRL but I wanted to share an initial thought here!

test/slackmock/slackmock.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"github.com/slackapi/slack-cli/internal/config"
+	"github.com/slackapi/slack-cli/internal/shared/types"
+	"github.com/spf13/afero"
+	"github.com/stretchr/testify/require"
+)
+
+// createProject will mock a project's required directory and files.
+// If there is an error, it will fail the test.
+func CreateProject(t require.TestingT, ctx context.Context, fs afero.Fs, os types.Os, projectDirPath string) {
```
zimeg, line 29: 🔍 I'm super curious if adding options of common fixtures for this might be tried later, or if that's something that'll be left to test setups?

The tests above I think are clear, and as more of `clients` is destructured this might read so well:

```go
slackmock.CreateProject(t, ctx, clients.Fs, clients.Os, slackdeps.MockWorkingDirectory)

_, err := config.WriteProjectConfigFile(ctx, clients.Fs, clients.Os, tt.projectConfig)
require.NoError(t, err)
```

internal/config/project.go
mwbrooks: I agree @zimeg, this function has always felt magical to me as well. 

Behind the scenes, it's using the current working directory as the project path, but that's bad practice imo. This kind of logic makes it difficult for us to support running commands in sub-directories and it confusing to read as a maintainer. 

We should definitely improve it in the future!

test/slackmock/slackmock.go
mwbrooks: 🧠 Interesting! What do you have in mind? Are you thinking project fixtures like "project-deno", "project-bolt-js", "project-missing-hooks", "project-missing-dotslack", etc?

zimeg: @mwbrooks Those are all seeming solid to me and appear most in tests IIRC - I do wonder if adding these now or much later makes the most sense in case other patterns appear, but those cases are still so good to have 😌 🎉

mwbrooks: @zimeg Sorry, I meant to respond before I hit merge 🤦🏻 I think we can add these later, but perhaps we should try to experiment with it while we're doing the destructify work, in case it's not as elegant as we hope?

mwbrooks:
Thanks so much for the PR review @zimeg! Keep the honest thoughts coming with this destructify work. The goal is to make the code easier to use and maintain, which reducing things like circular dependencies and passing "god objects." If you ever get a sense that this isn't the right direction - particularly from the testing perspective - then we can stop and re-eval!

PR #139, opened 2025-06-25
feat: include bolt for javascript and python in the 'samples' command

zimeg:
> **Changelog**
>
> The Bolt for JavaScript and Bolt for Python samples referenced in [documentation](https://docs.slack.dev/samples) are now included in prompts of the `samples` command and can be filtered with the `--language` flag.

### Summary

This PR adds Bolt for JavaScript and Bolt for Python sample apps to the `samples` command.

### Preview

https://github.com/user-attachments/assets/e1a87dcf-d2f8-43c2-9fa3-2c7b6fbf0c8c

### Reviewers

Create more apps with these commands:

```sh
$ slack create                      # Select "View more samples" and pick a language then sample!
$ slack samples                     # Browse the samples without a prior prompt
$ slack samples --language bolt-js  # Filter the selection to Bolt for JavaScript
```

### Notes

- The `--language` flag was named to match the current similar prompt of the `create` command, but I'm open to updating this if another option is better!
- The `--language` flag wasn't added to the `create` command to avoid filtering the curated apps. IMO it might be confusing to have with that command too 🤖 

### Requirements


cmd/feedback/feedback.go
```diff
@@ -377,6 +377,7 @@ func chooseSurveyPrompt(ctx context.Context, clients *shared.ClientFactory, surv
 	selection, err := clients.IO.SelectPrompt(ctx, msg, surveyPromptOptions,
 		iostreams.SelectPromptConfig{
 			Flag:     clients.Config.Flags.Lookup("name"),
+			Required: true,
```
zimeg, line 380: 🔍 The `required` option defaulting to **false** will cause this prompt to skip from the changes below!

internal/iostreams/survey.go
```diff
@@ -491,6 +495,9 @@ func (io *IOStreams) retrieveFlagValue(flagset []*pflag.Flag) (*pflag.Flag, erro
 		return nil, nil
 	}
 	for _, opt := range flagset {
+		if opt == nil {
+			continue
+		}
```
zimeg, line 500: 👁️‍🗨️ Here we avoid checking nonexistent flags for a changed value which is useful since the `create` command doesn't provide a "language" flag.

zimeg (commented):
📝 A few notes on changed selections adjacent to prompts for the most kind reviewers!

cmd/project/create_samples.go
```diff
@@ -36,13 +38,48 @@ func PromptSampleSelection(ctx context.Context, clients *shared.ClientFactory, s
...
 
-	projectType := "deno"
-	filteredRepos := filterRepos(sampleRepos, projectType)
+	projectTypes := []string{}
+	selection, err := clients.IO.SelectPrompt(ctx, "Select a language:",
+		[]string{
+			fmt.Sprintf("Bolt for JavaScript %s", style.Secondary("Node.js")),
+			fmt.Sprintf("Bolt for Python %s", style.Secondary("Python")),
+			fmt.Sprintf("Deno Slack SDK %s", style.Secondary("Deno")),
+		},
```
mwbrooks, line 47: praise: Great consideration for detail by choosing to match the styling of the `create` command! 🎨

cmd/project/samples.go
```diff
@@ -45,6 +46,7 @@ func NewSamplesCommand(clients *shared.ClientFactory) *cobra.Command {
 
 	cmd.Flags().StringVarP(&samplesTemplateURLFlag, "template", "t", "", "template URL for your app")
 	cmd.Flags().StringVarP(&samplesGitBranchFlag, "branch", "b", "", "name of git branch to checkout")
+	cmd.Flags().StringVar(&samplesLanguageFlag, "language", "", "framework from the app runtime\n  ex: \"bolt-js\", \"bolt-python\", \"deno\"")
```
mwbrooks: thought: I like the `--language` flag. It's more familiar than `--runtime`, which we went with prior to supporting Bolt.

However, I can't shake that 2 choices are frameworks and 1 choice in a runtime. None are actually languages. 😬 

I like the goal of keeping things consistent and recognizing that the `create` command uses language. The create command is also flawed, because it's listing Frameworks (Bolt JS, Bolt Python, and Deno SDK) and Runtimes (Node.js, Python, and Deno). 

To make matter worse, the hooks return a `runtime` value ("deno", "node", & "python"). This at least returns the actual runtimes, haha. And the `doctor` command displays these values under the title "Runtime".

Zooming out, perhaps `--runtime <deno|node|python>` is the technically correct and aligns the best with our current use of Runtime. While `--language <deno|bolt-js|bolt-python>` is more familiar to developers.

🧠 My opinion is to go with `--language <deno|node|python>`. This is a mix of both, but the reason is 3 fold:

1. If you ask a dev what language they prefer, many will say "Node" if they actually mean JavaScript or TypeScript running outside of the browser.
2. The values align with our Runtime values returned by the hooks, displayed by Doctor, and displayed as secondary by the `create` and `samples` prompts. 
3. Since the values match, we can always support `--runtime` as an flag alias using a `semver:minor` - in the future, if we decide runtime was the better choice. _Escape Hatch._

internal/iostreams/survey.go
mwbrooks: Interesting how the survey is kinda getting in the way. Thanks for improving it!

mwbrooks (approved):
✅ Implementation looks crisp and it works well! This is a very nice improvement!

🧪 Local testing works like a charm!

📝 My suggestion is to use `--language <deno|node|python>` to align the values with our `runtime` values but use a more familiar flag name. We may want to swap "Runtime" with "Language" in the Doctor command and verbose logs as well, eventually.

zimeg:
@mwbrooks These are great suggestions! 🙏 

> My opinion is to go with `--language <deno|node|python>`

I was hoping to match `create` outputs to start but I agree that `node` or `deno` or `python` match the filtering that follows "language" as expected and made this change in https://github.com/slackapi/slack-cli/pull/139/commits/f169934a29d0af3a626551cb995723233567f753.

The "runtime" option is so interesting to me though and wasn't something I considered! In later reworks I'll keep this in mind but for now I also find it to have more of an internal meaning.

> Interesting how the survey is kinda getting in the way.

This was a frustrated finding in an otherwise unrelated change but I hope we can revisit the cause of this soon 👁️‍🗨️ 

I will merge this PR now with the above changes. Thanks so much for the kind review and thoughts shared 🧠 ✨

PR #140, opened 2025-06-25
feat: accept the app name as an argument to the 'samples' command

zimeg:
> **Changelog**
>
> Now samples can be created with a provided app name using the `samples` command and a provided app name with perhaps the `slack samples my-app` example.

### Summary

This PR accepts the app name argument for the `samples` command.

### Preview

https://github.com/user-attachments/assets/5f2beb2f-5404-479d-920e-c3399bb35eeb

### Reviewers

Create even more samples at specified paths or not with these commands:

```sh
$ slack sample my-app
$ slack sample
$ slack sample /tmp/examples
```

### Requirements


cmd/project/samples.go
```diff
@@ -75,9 +78,7 @@ func runSamplesCommand(clients *shared.ClientFactory, cmd *cobra.Command, args [
 
 	// If preferred directory name is passed in as an argument to the `create`
 	// command first, honor that preference and use it to create the project
-	if len(args) > 0 {
-		createCmd.SetArgs([]string{args[0]})
-	}
+	createCmd.SetArgs(args)
```
mwbrooks, line 83: praise: Ah, much more elegant!

mwbrooks (approved):
✅ Looks great and this is a really nice improvement that was overlooked!

🧪 Manual testing works well and test coverage looks good.

❓ Thoughts on this being a feature rather than a bug? Our command's help clearly documented that there were no arguments and the logic prevented extra arguments. To me, this looks like an `enhancement` and `semver:minor` 😄

zimeg:
@mwbrooks Thanks so much - I was confused to find this didn't match the `create` command which is what motivated "fix" and perhaps the comments adjacent sidetracked me...

But I notice now that documentation reflected what was found in releases so I updated this to be a "feat" and will merge it as such! 🚢 💨

PR #111, opened 2025-05-29
feat(bolt-install): support using 'run' command with remote manifests

mwbrooks:
### CHANGELOG

> Update the `slack run` command to create and install new & existing Bolt Framework apps that are configured to app settings as the source of truth (i.e. remote manifest).

### Summary

This pull request updates the `slack run` command to create and install new & existing apps that use app settings as the source of truth (remote manifest).

#### `slack run` creates a new Bolt App with app settings as the source of truth

https://github.com/user-attachments/assets/f245349d-eb9f-43c7-ae39-4ba586c72386

#### `slack run` uses an existing Bolt App with app settings as the source of truth

https://github.com/user-attachments/assets/5719ed20-2536-49be-a4d6-f135af8843fa

### Reviewer

```bash
# Run new ROSI app without experiment
$ lack create asdf -t slack-samples/deno-starter-template
$ cd asdf/
$ lack run
# → Select "create new app"
# → Confirm run starts
# → Press CTRL+C

# Run existing ROSI app without experiment
$ lack run
# → Select existing app
# → Confirm run starts
# → Press CTRL+C

# Clean up ROSI app
$ lack delete -f
$ cd ../
$ rm -rf asdf/

# ---

# Run new ROSI app with experiment
$ lack create asdf -t slack-samples/deno-starter-template --experiment bolt-install
$ cd asdf/
$ lack run --experiment bolt-install
# → Select "create new app"
# → Confirm run starts
# → Press CTRL+C

# Run existing ROSI app with experiment
$ lack run --experiment bolt-install
# → Select existing app
# → Confirm run starts
# → Press CTRL+C

# Clean up ROSI app
$ lack delete -f
$ cd ../
$ rm -rf asdf/

# ---

# Run new Bolt app without experiment
$ lack create asdf -t slack-samples/bolt-js-starter-template
$ cd asdf/
$ lack run
# → Select existing app
# → Confirm run starts
# → Press CTRL+C
# Note: This works because the manifest source is local from the create command without the experiment

# Clean up Bolt app
$ lack delete -f
$ cd ../
$ rm -rf asdf/

# ---

# Run new Bolt app with experiment
$ lack create asdf -t slack-samples/bolt-js-starter-template --experiment bolt-install
$ cd asdf/
$ lack run --experiment bolt-install
# → Select existing app
# → Confirm run starts
# → Press CTRL+C

# Run existing Bolt app with experiment
$ lack run --experiment bolt-install
# → Select existing app
# → Confirm run starts
[...]

PR #154, opened 2025-07-11
feat(bolt-install): support using 'install' command with remote manifests

mwbrooks:
### CHANGELOG

> Update the `slack install` command to create and install new Bolt Framework apps that are configured to use app settings as the source of truth (i.e. remote manifest).

### Summary

This pull request updates the `slack install` command to create and install new apps that use app settings as the source of truth (remote manifest). It's a continuation of PR #111, which added similar support to `slack run`.

Affected commands:
* `slack install`
* `slack deploy`
* `slack trigger create`

### Follow-up PRs

- [ ] [Add deprecation notice for `--app local` and `--app deployed`/](https://github.com/slackapi/slack-cli/pull/154#discussion_r2231755015)
- [ ] Refactor `--app <app-id>` and `--app <env>` validation to `promptIsProduction` and add test coverage.

### Reviewers

_Each use-case includes a collapsed section with manual steps._

https://github.com/user-attachments/assets/5e4050d3-a387-443c-9325-6a86d81dd824

<details>
<summary>Manual Test Steps</summary>

```bash
$ lack create asdf/ -t slack-samples/deno-starter-template
$ cd asdf/

# Test Install
$ lack install
# → Select Local
# → Select a team
$ lack install
# → Select Deploy
# → Select same team

# Test Re-install
$ lack install
# → Select Local
# → Select same team
$ lack install
# → Select Deploy
# → Select same team

# Cleanup
$ lack delete -f
# → Select an app
$ lack delete -f
# → Select an app
$ cd ..
$ rm -rf asdf/
```
</details>

https://github.com/user-attachments/assets/b2a36d9b-e284-4cab-91b4-30e5a09675c8

<details>
<summary>Manual Test Steps</summary>

```bash
$ lack create asdf/ -t slack-samples/deno-starter-template
$ cd asdf/

# Test Deploy
$ lack deploy
# → Select a team

# Test Re-deploy
$ lack deploy
# → Select existing app

# Cleanup
$ lack delete -f
# → Select an app
$ lack delete -f
# → Select an app
$ cd ..
$ rm -rf asdf/
```
</details>

https://github.com/user-attachments/assets/ded6d1ab-973e-4431-8f6a-4ccf495a2e44

<details>
<summary>Manual Test Steps</summary>

```bash
$ lack create asdf/ -t slack-samples/deno-starter-template
$ cd asdf/

# Create Trigger on New App
$ lack trigger create --trigger-def triggers/sample_trigger.ts
# → Select a team

# Create Trigger on Existing App
$ lack trigger create --trigger-def triggers/sample_trigger.ts
# → Select existing app

# Cleanup
$ lack delete -f
# → Select an app
$ lack delete -f
# → Select an app
$ cd ..
$ rm -rf asdf/
```
</details>

https://github.com/user-attachments/assets/a1251462-acf3-4c94-8722-30b8c892d9b7

<details>
<summary>Manual Test Steps</summary>

```bash
$ lack create asdf/ -t slack-samples/bolt-js-starter-template
$ cd asdf/

# Test Install
$ lack install
# → Select Local
# → Select a team
$ lack install
# → Select Deploy
# → Select same team

# Test Re-install
$ lack install
# → Select Local
# → Select same team
$ lack install
# → Select Deploy
# → Select same team

# Cleanup
$ lack delete -f
# → Select an app
$ lack delete -f
# → Select an app
$ cd ..
$ rm -rf asdf/
```
</details>

https://github.com/user-attachments/assets/c31e14f8-afb1-43dd-a36c-b5874bea89ce

<details>
<summary>Manual Test Steps</summary>

```bash
$ lack create asdf/ -t slack-samples/bolt-js-starter-template
$ cd asdf/

# Test No Deploy Script
$ lack deploy
# → Select a team
# → Confirm script tutorial error message

# Create Deploy Script
$ vim .slack/hooks.json
# → Add "deploy": "./deploy.sh"
$ vim deploy.sh
# #!/bin/bash
# 
# echo "Deployed!"
$ chmod +x deploy.sh

# Test Deploy
$ lack deploy
# → Select a team

# Test Re-deploy
$ lack deploy
# → Select existing app

# Cleanup
$ lack delete -f
# → Select an app
$ lack delete -f
# → Select an app
$ cd ..
$ rm -rf asdf/
```
</details>

https://github.com/user-attachments/assets/52acbb1d-87fa-4482-8246-7838a93bd91f

<details>
<summary>Manual Test Steps</summary>

```bash
$ lack create asdf/ -t slack-samples/bolt-js-starter-template
$ cd asdf/

# Test Defaulting to Deployed
$ lack install --team T0123
# → Confirm no app environment prompt
# → Confirm creates a deployed app (no local suffix)

# Test --environment local
$ lack install --team T0123 --environment local
# → Confirm no app environment prompt
# → Confirm creates a local app

# Test --environment deployed
$ lack install --team T0123 --environment deployed
# → Confirm no app environment prompt
# → Confirm creates a deployed app

# Cleanup
$ lack delete -f
# → Select an app
$ lack delete -f
# → Select an app
$ cd ..
$ rm -rf asdf/
```
</details>

#### Scripting & Flags - `slack install --team T0123`

https://github.com/user-attachments/assets/e385aedb-574f-44ec-a27c-1e4818594e7f

<details>
<summary>Manual Test Steps</summary>

```bash
$ lack create asdf/ -t slack-samples/bolt-js-starter-template
$ cd asdf/

# Test Defaulting to Deployed
$ lack install --team T0123
# → Confirm warning displayed
# → Confirm no app environment prompt
# → Confirm creates a deployed app (no local suffix)

# Cleanup
$ lack delete -f
# → Select an app
$ cd ..
$ rm -rf asdf/
```
</details>

#### Scripting & Flags - Flag Mismatch `--app`, `--team`, & `--environment`

https://github.com/user-attachments/assets/bde81ab9-9f0d-46ee-bb47-21a227386151

<details>
<summary>Manual Test Steps</summary>

```bash
$ lack create asdf/ -t slack-samples/bolt-js-starter-template
$ cd asdf/

# Install a valid app
$ lack install
# → Copy App ID

# Test Mismatched --app and --team
$ slack install --app <valid-app-id> --team <invalid-team-id>
# → Confirm error

# Test Mismatched --app and --environment
$ slack install --app <valid-app-id> --environment <invalid-environment>
# → Confirm error

# Cleanup
$ lack delete -f
# → Select an app
$ cd ..
$ rm -rf asdf/
```
</details>

#### Scripting & Flags - Support `--app local` and `--app deployed`

Working example:
https://github.com/user-attachments/assets/dcc19355-4b65-42ff-a115-56cd57610598

Mismatch errors:
https://github.com/user-attachments/assets/dea85d8d-8751-4aac-a571-f3d306b2dec1

<details>
<summary>Manual Test Steps</summary>

```bash
$ lack create asdf/ -t slack-samples/bolt-js-starter-template
$ cd asdf/

# Install a valid app for the local environemnt
$ lack install --app local
# → Confirm environment prompt is skipped
# → Confirm app is local

$ lack install --app deployed
# → Confirm environment prompt is skipped
# → Confirm app is deployed

$ lack install --app local --environment deployed
# → Confirm mismatched error message

$ lack install --app deployed --environment local
# → Confirm mismatched error message

# Cleanup
$ lack delete -f
# → Select an app
$ lack delete -f
# → Select an app
$ cd ..
$ rm -rf asdf/
```
</details>

#### Scripting & Flags - Support `--app A123`

Working example:
https://github.com/user-attachments/assets/42caf961-2200-4db2-b683-c2c43ce16c35

Mismatch errors:

https://github.com/user-attachments/assets/b66d359c-2f4c-43e3-a59d-2657887f9ffd

<details>
<summary>Manual Test Steps</summary>

```bash
$ lack create asdf/ -t slack-samples/bolt-js-starter-template
$ cd asdf/

# Install a local and deployed app
$ lack install --app local
$ lack install --app deployed

# Copy Local App ID
$ lack app list
→ Copy local App ID

# Test Local App ID (replace A123 with your App ID)
$ lack install --app A123
# → Confirm local app is re-installed

# Test Deployed App ID (replace A456 with your App ID)
$ lack install --app A456
# → Confirm deployed app is re-installed

# Test mismatched flags
$ lack install --app A123 --environment local
# → Confirm mismatched error message
$ lack install --app A456 --environment local
# → Confirm mismatched error message

# Cleanup
$ lack delete -f
# → Select an app
$ lack delete -f
# → Select an app
$ cd ..
$ rm -rf asdf/
```
</details>

### Requirements


cmd/app/add.go
```diff
@@ -85,38 +85,36 @@ func preRunAddCommand(ctx context.Context, clients *shared.ClientFactory) error
...
-				style.Commandf("app link", false),
-				style.Commandf("app list", false),
-			).
-			WithDetails(slackerror.ErrorDetails{
-				slackerror.ErrorDetail{
-					Code:    slackerror.ErrProjectConfigManifestSource,
-					Message: "Cannot install apps with manifests sourced from app settings",
-				},
-			})
-	}
```
mwbrooks, line 107: 🪓 Removing this section that prevents the `slack install` command from executing on a remote manifest.

cmd/app/add.go
```diff
@@ -85,38 +85,36 @@ func preRunAddCommand(ctx context.Context, clients *shared.ClientFactory) error
...
+		}
+
+		var appEnvironmentType prompts.AppEnvironmentType
+		if isProductionApp {
+			appEnvironmentType = prompts.ShowHostedOnly
+		} else {
+			appEnvironmentType = prompts.ShowLocalOnly
+		}
+
+		selected, err := teamAppSelectPromptFunc(ctx, clients, appEnvironmentType, prompts.ShowAllApps)
```
mwbrooks: note: New code that prompts the developer to choose an environment (local or deployed). Then displays the proper Team Prompt. We cannot `prompts.ShowAllEnvironments` because there is guard logic to avoid having 2 apps on one team. Unfortunately, the prompt isn't aware of the app's environment, so this was our workaround.

cmd/app/add.go
```diff
@@ -85,38 +85,36 @@ func preRunAddCommand(ctx context.Context, clients *shared.ClientFactory) error
...
 		if err != nil {
 			return ctx, "", types.App{}, err
 		}
 		selection = &selected
+
+		if !isProductionApp {
+			selection.App.IsDev = true
+		}
```
mwbrooks: note: Normally, we'd also update the `UserID` for dev apps, but it's already been set.

cmd/app/add_test.go
```diff
@@ -85,40 +85,27 @@ func TestAppAddCommandPreRun(t *testing.T) {
...
-					style.LinkText("https://api.slack.com/apps"),
-					style.Commandf("app link", false),
-					style.Commandf("app list", false),
-				).
-				WithDetails(slackerror.ErrorDetails{
-					slackerror.ErrorDetail{
-						Code:    slackerror.ErrProjectConfigManifestSource,
-						Message: "Cannot install apps with manifests sourced from app settings",
-					},
-				}),
```
mwbrooks, line 102: 🪓 We no longer want to error on a remote manifest, so this test is removed and replaced with `proceeds if manifest.source is remote with the bolt experiment` (below)

cmd/app/add_test.go
```diff
@@ -130,13 +117,84 @@ func TestAppAddCommandPreRun(t *testing.T) {
 }
 
 func TestAppAddCommand(t *testing.T) {
-
 	testutil.TableTestCommand(t, testutil.CommandTests{
+		"adds a new local app": {
```
mwbrooks, line 142: note: Added test for new local apps to complement the existing `new deployed app` test.

cmd/platform/deploy.go
```diff
@@ -163,16 +163,24 @@ func hasValidDeploymentMethod(
...
 	switch {
+	// When the manifest source is local, we can get the manifest from the local project.
 	case manifestSource.Equals(config.ManifestSourceLocal):
 		manifest, err = clients.AppClient().Manifest.GetManifestLocal(ctx, clients.SDKConfig, clients.HookExecutor)
 		if err != nil {
 			return err
 		}
-	case manifestSource.Equals(config.ManifestSourceRemote):
+	// When the manifest source is remote and the app exists, we can get the manifest from the the API.
+	case manifestSource.Equals(config.ManifestSourceRemote) && app.AppID != "":
```
mwbrooks, line 172: note: We need to differentiate between a remote manifest **with** an app (get manifest from app settings) and a remote manifest **without** an app (get manifest from local file).

docs/reference/experiments.md
```diff
@@ -8,13 +8,14 @@ The following is a list of currently available experiments. We'll remove experim
 
 * `bolt-install`: enables creating, installing, and running Bolt projects that manage their app manifest on app settings (remote manifest).
     * `slack create` and `slack init` now set manifest source to "app settings" (remote) for Bolt JS & Bolt Python projects ([PR#96](https://github.com/slackapi/slack-cli/pull/96)).
+    * `slack run` and `slack install` support creating and installing Bolt Framework apps that have the manifest source set to "app settings (remote)" ([PR#111](https://github.com/slackapi/slack-cli/pull/111), [PR#TODO](https://github.com/slackapi/slack-cli/pull/TODO)).
```
mwbrooks: note: I added the `slack run` in PR #111 but it may have been removed by a rogue docs sync 🥷

internal/pkg/apps/install.go
```diff
@@ -54,8 +54,11 @@ func Install(ctx context.Context, clients *shared.ClientFactory, log *logger.Log
 	if err != nil {
 		return types.App{}, "", err
 	}
-	if !manifestUpdates && !manifestCreates {
-		return app, "", nil
+
+	if !clients.Config.WithExperimentOn(experiment.BoltInstall) {
+		if !manifestUpdates && !manifestCreates {
+			return app, "", nil
+		}
```
mwbrooks, line 61: note: Identical to installing a dev app in PR #111

internal/pkg/apps/install.go
```diff
@@ -81,22 +84,47 @@ func Install(ctx context.Context, clients *shared.ClientFactory, log *logger.Log
...
 
-	slackYaml, err := clients.AppClient().Manifest.GetManifestLocal(ctx, clients.SDKConfig, clients.HookExecutor)
-	if err != nil {
-		return app, "", err
+	// When the BoltInstall experiment is enabled, we need to get the manifest from the local file
+	// if the manifest source is local or if we are creating a new app. After an app is created,
+	// app settings becomes the source of truth for remote manifests, so updates and installs always
+	// get the latest manifest from app settings.
+	// When the BoltInstall experiment is disabled, we get the manifest from the local file because
+	// this is how the original implementation worked.
```
mwbrooks, line 92: note: Identical to installing a dev app in PR #111

internal/pkg/apps/install.go
```diff
@@ -124,6 +152,9 @@ func Install(ctx context.Context, clients *shared.ClientFactory, log *logger.Log
 		app.AppID = result.AppID
 		app.TeamID = *authSession.TeamID
 		app.TeamDomain = auth.TeamDomain
+		// TODO: add enterprise ID and user ID to app? See InstallLocalApp.
+		// app.EnterpriseID = config.GetContextEnterpriseID(ctx)
+		// app.UserID = *authSession.UserID
```
mwbrooks: note: While comparing the dev install and this install, I noticed that we set the `EnterpriseID` and `UserID` in the other one. I've left a note incase we want to do it here as well.

mwbrooks (commented):
🧘🏻 Comments for the kind reviewers!

docs/reference/experiments.md
lukegalbraithrussell:
```suggestion
    * `slack run` and `slack install` support creating and installing Bolt Framework apps that have the manifest source set to "app settings (remote)" ([PR#111](https://github.com/slackapi/slack-cli/pull/111), [PR#TODO](https://github.com/slackapi/slack-cli/pull/TODO)).
```

assuming this #TODO is still needed to be filled out?

mwbrooks: Whew! Thank you, @lukegalbraithrussell! Commit 10966f77e1515a54da6cd218e12444f34e77dced adds the PR number.

mwbrooks: @lukegalbraithrussell Don't sweat it! The docs sync has improved a lot since then, thank to you!

cmd/app/add.go
```diff
@@ -72,6 +73,7 @@ func NewAddCommand(clients *shared.ClientFactory) *cobra.Command {
 	}
 
 	cmd.Flags().StringVar(&addFlags.orgGrantWorkspaceID, cmdutil.OrgGrantWorkspaceFlag, "", cmdutil.OrgGrantWorkspaceDescription())
+	cmd.Flags().StringVarP(&addFlags.environmentFlag, "environment", "E", "", "environment of app (local, deployed)")
```
mwbrooks, line 77: note: The `slack install` command now supports the `--environment <local|deployed>` values. This flag can be used to by-pass the prompt.

cmd/app/add.go
```diff
@@ -85,38 +87,50 @@ func preRunAddCommand(ctx context.Context, clients *shared.ClientFactory) error
...
+				"Defaulting app environment to deployed because team flag is provided. "+
+					"Please use '--environment deployed' to avoid breaking changes in the next major version.",
+			)
+		} else {
+			// Prompt for deployed or local app environment.
+			isProductionApp, err = promptIsProduction(ctx, clients)
+			if err != nil {
+				return ctx, "", types.App{}, err
+			}
+		}
```
mwbrooks, line 114: question: I really don't like this implementation and would appreciate suggestions.

Today, `slack install` always uses a deployed app. So scripts using `slack install --team T0123` always create and install a deployed app .

This PR introduces an app environment prompt, which will block the scripts. To keep backwards compatibility, we default to deployed when no `--environment` is specified.

The `promptIsProduction` supports `clients.Config.Flags.Lookup("environment")`. However, I'm having trouble having this flag listed and right now it always return `nil`. If we can set the flag to deployed, this code can become less nested and more linear to read.

cmd/app/add_test.go
```diff
@@ -356,11 +414,84 @@ func TestAppAddCommand(t *testing.T) {
 				cm.API.AssertCalled(t, "DeveloperAppInstall", mock.Anything, mock.Anything, mockOrgAuth.Token, mock.Anything, mock.Anything, mock.Anything, "T123", mock.Anything)
 			},
 		},
+		// TODO(semver:major): Remove this test when the defaulting to deployed is removed.
+		"adds a new deployed app when team flag is provided and environment flag is not set": {
```
mwbrooks, line 607: note: Test for defaulting to `deployed` when the team flag is specified and there is no `--environment` flag.

cmd/app/add.go
```diff
@@ -85,38 +88,56 @@ func preRunAddCommand(ctx context.Context, clients *shared.ClientFactory) error
...
+			isProductionApp = true
+		case "local":
+			isProductionApp = false
+		default:
+			// Prompt for deployed or local app environment.
+			isProductionApp, err = promptIsProduction(ctx, clients)
+			if err != nil {
+				return ctx, "", types.App{}, err
+			}
+		}
```
mwbrooks: question: I really don't like this implementation and would appreciate suggestions.

Today, `slack install` always uses a deployed app. So scripts using `slack install --team T0123` always create and install a deployed app .

This PR introduces an app environment prompt, which will block the scripts. To keep backwards compatibility, we default to deployed when no `--environment` is specified.

The `promptIsProduction` supports `clients.Config.Flags.Lookup("environment")`. However, I'm having trouble having this flag listed and right now it always return nil. If we can set the flag to deployed, this code can become much simpler.

internal/pkg/apps/install.go
```diff
@@ -257,9 +288,7 @@ func printNonSuccessInstallState(ctx context.Context, clients *shared.ClientFact
 	clients.IO.PrintInfo(ctx, false, status)
 }
 
-func validateManifestForInstall(ctx context.Context, clients *shared.ClientFactory, app types.App, appManifest types.AppManifest) error {
-	var token = config.GetContextToken(ctx)
```
zimeg, line 261: 🪓 So helpful when walking through the code!

internal/pkg/apps/install.go
zimeg: @mwbrooks `EnterpriseID` makes a lot of sense to me!

I'm hesitant for the `UserID` at the moment since we might be using this - alongside the `IsDev` value - to differentiate "deployed" and "local" apps IIRC 🔍 

No blocker for this PR, but the app `UserID` might be something to inspect more later!

lukegalbraithrussell (approved):
docs are good!

cmd/app/add.go
zimeg: @mwbrooks The `environment` then `team` ordering is most interesting to me and makes me think about confusion I might feel in similar prompts! 👁️‍🗨️ ✨ 

Thanks for introducing a flag to avoid breakings with these changes also.

To share this flag across commands we might need to add the following to `PreRunE`:

```go
clients.Config.SetFlags(cmd)
```

And then set the `pflag` with a default value in this function:

```go
err := clients.Config.Flags.Lookup("environment").Value.Set("deployed")
if err != nil {
	return ctx, "", types.App{}, err
}
```

Not a blocker for this PR, but this might be nice to have for shared flags 🏁

cmd/app/add.go
```diff
@@ -85,38 +88,56 @@ func preRunAddCommand(ctx context.Context, clients *shared.ClientFactory) error
...
 // RunAddCommand executes the workspace install command, prints output, and returns any errors.
 func RunAddCommand(ctx context.Context, clients *shared.ClientFactory, selection *prompts.SelectedApp, orgGrantWorkspaceID string) (context.Context, types.InstallState, types.App, error) {
 	if selection == nil {
-		selected, err := teamAppSelectPromptFunc(ctx, clients, prompts.ShowHostedOnly, prompts.ShowAllApps)
+		var isProductionApp bool
+		var err error
+
+		// When team flag is provided, default app environment to deployed if not specified.
+		// TODO(semver:major): Remove defaulting to deployed and require the environment flag to be set.
+		if clients.Config.TeamFlag != "" && addFlags.environmentFlag == "" {
```
zimeg:
```suggestion
		if (clients.Config.TeamFlag != "" || clients.Config.AppFlag != "") && addFlags.environmentFlag == "" {
```

I'm not certain that this is the right logic, but we might want to use a default value if the `--app` flag is provided 🔍

zimeg (changes requested):
@mwbrooks This is looking great so far! 🎉 

I'm requesting changes now with a note on the `--app` flag but will continue a review and testing in the meantime.

Thanks for cleaning adjacent things up with these changes, as always 💌

internal/pkg/apps/install.go
mwbrooks: Thanks for the feedback @zimeg!

I'll test again on an Enterprise Organization. If the app works properly, I'll hold off on making either of these changes. If it has been working until now, I'd be concerned about a regression. We can follow-up with a separate PR that sets `EnterpriseID` or something else to align the 2 methods.

cmd/app/add.go
mwbrooks: Thanks @zimeg! Your suggestion to use `clients.Config.SetFlags(cmd)` is exactly what I was overlooking 🤦🏻! I'll make the updates and see how it looks.

cmd/app/add.go
zimeg: 🗣️ This made me question if `--app` should accept `local` or `deployed` in future versions!

Preferring an `--environment` flag for that choice and leaving app IDs to `--app` might bring more straightforward logic here and in other selections, but this isn't a change we ought make now!

mwbrooks: Flag combinations can be confusing! 😝

I think we want the original combination of:

```golang
if clients.Config.TeamFlag != "" && addFlags.environmentFlag == "" {
```

* When `--team` is provided and `--environment` is not, then we default to `--environment deployed`.
* When `--app` is provided, we resolve the `--environment` from `apps.json` and `apps.dev.json`.
* When there is a mismatched combination, then the `AppSelect` will return an error that the app cannot be found.
  * Example: `--team T0123 --app A0123` where `A0123` doesn't belong to `T0123`

cmd/app/add.go
```diff
@@ -72,51 +74,76 @@ func NewAddCommand(clients *shared.ClientFactory) *cobra.Command {
...
-				style.Commandf("app list", false),
-			).
-			WithDetails(slackerror.ErrorDetails{
-				slackerror.ErrorDetail{
-					Code:    slackerror.ErrProjectConfigManifestSource,
-					Message: "Cannot install apps with manifests sourced from app settings",
-				},
-			})
-	}
+	clients.Config.SetFlags(cmd)
```
mwbrooks, line 91: praise: This was the magic, thanks @zimeg 🪄

cmd/app/add.go
```diff
@@ -72,51 +74,73 @@ func NewAddCommand(clients *shared.ClientFactory) *cobra.Command {
...
 	if selection == nil {
-		selected, err := appSelectPromptFunc(ctx, clients, prompts.ShowHostedOnly, prompts.ShowAllApps)
+		// When team flag is provided, default app environment to deployed if not specified.
+		// TODO(semver:major): Remove defaulting to deployed and require the environment flag to be set.
+		if clients.Config.TeamFlag != "" && addFlags.environmentFlag == "" {
+			err := clients.Config.Flags.Lookup("environment").Value.Set("deployed")
+			if err != nil {
+				return ctx, "", types.App{}, err
+			}
+			clients.Config.Flags.Lookup("environment").Changed = true
```
mwbrooks, line 128: note: We now set the flag value (and mark it as changed), so that the `promptIsProduction(...)` function can be used to read the `--environment` flag and optionally prompt for an app environment.

cmd/app/add.go
```diff
@@ -72,51 +74,73 @@ func NewAddCommand(clients *shared.ClientFactory) *cobra.Command {
...
+				Emoji: "warning",
+				Text:  "Warning: Default App Environment",
+				Secondary: []string{
+					"App environment is set to deployed when only the --team flag is provided.",
+					"The next major version will change this behavior.",
+					"When the --team flag is provided, the --environment flag will be required.",
+					"Add the '--environment deployed' to avoid breaking changes.",
+				},
+			}))
+		}
```
mwbrooks, line 140: note: A warning is now displayed that in the next major release, we'll require the `--environment` flag when the `--team` flag is provided.

Image

The double newline haunts me 😱 but it only appear when the warning is shown... which will be rare.

internal/pkg/apps/install.go
mwbrooks: @zimeg I've removed the `UserID` comment so that we don't accidentally implement it later on. I agree, there could be bits of code relying on it to identify a dev app.

mwbrooks (commented):
📝 Thanks for the thorough review @zimeg and for noticing the edge-cases with the `--app`, `--team`, and `--environment` flag combinations.

✨ A few updates since our discussion:
* Display a better warning when the `--team` flag is used and `--environment deploy` is inferred for backwards compability
* `promptIsProduction(...)` now reads the `--environment` flag properly
* Mismatch flag combinations should be handled, such as `--team T0123 --app A0123` and this is largely enforced by `SelectPrompt(...)`
* Updated the original description with 2 new videos showcasing additional scripting use-cases

I know it's tough to think through these use-cases, but I'd appreciate your eye on whether we're overlooking additional use-cases! 🙇🏻

cmd/app/add.go
zimeg: @mwbrooks I'm such a big fan of making this flag combination **required**! What a good call 🤖 ✨

cmd/app/add.go
```diff
@@ -42,6 +42,7 @@ var appSelectPromptFunc = prompts.AppSelectPrompt
 
 type addCmdFlags struct {
 	orgGrantWorkspaceID string
+	environmentFlag     string
 }
```
zimeg, line 46: 📝 thought: IMHO this pattern is so interesting. We might want to encourage it and follow up with changes that bring the `app` and `team` and `token` flags to specific commands soon?

cmd/app/add.go
```diff
@@ -56,10 +57,11 @@ func NewAddCommand(clients *shared.ClientFactory) *cobra.Command {
 		Example: style.ExampleCommandsf([]style.ExampleCommand{
 			{Command: "app install", Meaning: "Install a production app to a team"},
 			{Command: "app install --team T0123456", Meaning: "Install a production app to a specific team"},
```
zimeg:
```suggestion
			{Command: "app install --team T0123456 --environment deployed", Meaning: "Install a production app to a specific team"},
```

⚠️ suggestion: Wanting to avoid the deprecation message that might follow!

internal/pkg/apps/install.go
```diff
@@ -81,22 +84,47 @@ func Install(ctx context.Context, clients *shared.ClientFactory, log *logger.Log
...
 
-	log.Data["appName"] = slackYaml.DisplayInformation.Name
+	log.Data["appName"] = slackManifest.DisplayInformation.Name
 	log.Data["isUpdate"] = app.AppID != ""
 	log.Data["teamName"] = *authSession.TeamName
 	log.Log("INFO", "app_install_manifest")
 
-	manifest := slackYaml.AppManifest
-	if slackYaml.IsFunctionRuntimeSlackHosted() {
+	manifest := slackManifest.AppManifest
```
zimeg, line 122: 🤩 praise: The `slackManifest` including an app manifest is so much more clear than the prior `slackYaml`!

internal/pkg/apps/install_test.go
```diff
@@ -55,7 +55,8 @@ func TestInstall(t *testing.T) {
 		mockBoltExperiment      bool
 		mockConfirmPrompt       bool
 		mockIsTTY               bool
-		mockManifest            types.SlackYaml
+		mockLocalManifest       types.SlackYaml
+		mockRemoteManifest      types.SlackYaml
```
zimeg: 👁️‍🗨️ quibble: Some strange postfix notation might make this ordering more clear?

```suggestion
		mockManifestAppLocal    types.SlackYaml
		mockManifestAppRemote   types.SlackYaml
```

cmd/app/add.go
zimeg: 🧠 question: I noticed the **warning** below hints at `--team` needing the `--environment` flag and was wondering if you think `--app` should also remove support for "local" or "deployed" in the next `semver:major`?

🧠 praise: Also, thanks for clearing up these default cases and flag setups alike 🤓

cmd/app/add.go
```diff
@@ -72,51 +74,73 @@ func NewAddCommand(clients *shared.ClientFactory) *cobra.Command {
...
+			return ctx, "", types.App{}, err
+		}
+
+		// Set the app environment type based on the prompt.
+		var appEnvironmentType prompts.AppEnvironmentType
+		if isProductionApp {
+			appEnvironmentType = prompts.ShowHostedOnly
+		} else {
+			appEnvironmentType = prompts.ShowLocalOnly
+		}
```
zimeg: 👁️‍🗨️ issue: This prompt might appear if the `app` flag is included for an installed app.

🗣️ suggestion: We might want to guard against this case inline and with a test since I'm not sure how else `promptIsProduction` is used, but returning early might be a solid option too!

zimeg: 💟 praise: I like the environment selection before app selection too!

cmd/app/add_test.go
zimeg: 🪓 thought: Once the `bolt` experiment is removed altogether, we can perhaps remove even more cases!

cmd/platform/deploy_test.go
```diff
@@ -144,6 +145,20 @@ func TestDeployCommand_HasValidDeploymentMethod(t *testing.T) {
 			manifestSource: config.ManifestSourceLocal,
 			expectedError:  slackerror.New(slackerror.ErrSDKHookNotFound),
 		},
+		"succeeds if the app exists and the manifest source is remote": {
```
zimeg, line 148: 😉 thought: These are clever tests. I think it calls out the importance of error codes in assertions. But seems right here!

zimeg (changes requested):
@mwbrooks Thanks so much for keeping the description updated. It makes reviews so much easier 👾

I'm requesting changes here with a side effect of **reinstalling** apps with the `app` flag, though it's not so clear to me if the `environment` flag is also needed here:

```sh
$ slack app install --app A0986U6T03A | cat
```

Other comments include thoughts and a small suggestion or two, but I'm excited for this to land and it feels close! ✨

cmd/app/add.go
mwbrooks: > 🧠 question: I noticed the warning below hints at `--team` needing the `--environment` flag and was wondering if you think --app should also remove support for "local" or "deployed" in the next `semver:major`?

I won't lie, I thought `--app local` and `--app deployed` support was removed in v3.0.0 when we introduced `--environment`. This should definitely be removed in the next `semver:major` in favour of `--environment`. :two-cents:

In the meantime, we'll need to handle that use-case here, I suppose 🤔

cmd/app/add_test.go
mwbrooks: Yes please!

cmd/app/add.go
mwbrooks: 🧠 Hmm... we could consider doing that. We may also want to consider `--environment` to be a global flag, since it should be used in combination with `--team` 🤔

internal/pkg/apps/install_test.go
mwbrooks: Commit 013ad0d refactors the naming to `mockManifestAppLocal` and `mockManifestAppRemote` 🔨

cmd/app/add.go
zimeg: 🪓 praise: Having a shared `environment` flag will be so nice. I agree on the continued support until this!

cmd/platform/deploy.go
zimeg: 👁️‍🗨️ thought: An `app.Exists` method might be nice here-

cmd/app/add.go
zimeg: 👾 issue(non-blocking): Hmm I think we might still want to skip this section if an app ID **is** provided since the environment would be known?

🤖 question: It's not so clear to me if an `install` can target different environments for a known app. This might not be an issue if so, but I'm wondering what you think.

```
$ slack install --team tinyspek --app A099WJ22LPK

⚠️ Warning: Default App Environment
   App environment is set to deployed when only the --team flag is provided.
   The next major version will change this behavior.
   When the --team flag is provided, the --environment flag will be required.
   Add the '--environment deployed' to avoid breaking changes.


🏠 App Install
   Installing "vigilant-shark-290 (local)" app to "Tinyspek"
   Finished in 0.4s
```

📌 note: IMO we should not suggest an app and team flag together once the environment flag becomes more standard.

cmd/app/add.go
```diff
@@ -72,51 +74,105 @@ func NewAddCommand(clients *shared.ClientFactory) *cobra.Command {
...
 // RunAddCommand executes the workspace install command, prints output, and returns any errors.
 func RunAddCommand(ctx context.Context, clients *shared.ClientFactory, selection *prompts.SelectedApp, orgGrantWorkspaceID string) (context.Context, types.InstallState, types.App, error) {
 	if selection == nil {
-		selected, err := appSelectPromptFunc(ctx, clients, prompts.ShowHostedOnly, prompts.ShowAllApps)
-		if err != nil {
-			return ctx, "", types.App{}, err
+		// TODO: Move to the promptIsProduction when the prompt is refactored and tested.
+		// Validate that the --app flag is not an app ID when the --environment flag is set.
+		if types.IsAppID(clients.Config.AppFlag) && addFlags.environmentFlag != "" {
+			return ctx, "", types.App{}, slackerror.New(slackerror.ErrMismatchedFlags).WithRemediation("When '--app <app_id>' is set, please do not set the flag --environment.")
```
zimeg, line 101: 🕊️ praise: This is a pleasant remediation.

zimeg (approved):
@mwbrooks LGTM! And thank you for the follow ups across these comments 🙏 ✨

The demo cases are great and I agree that more changes to prompts can happen in follow up PRs.

One perhaps issue related to the `app` flag **without** an environment flag for a saved "local" app is noted below, but nothing blocking AFAICT. Moving towards the `environment` flag and deprecated "deploy" and "local" options elsewhere will be such a nice change for the code I hope 🗣️ 

Please do feel free to merge this if the case below tests alright to you! 🚢 💨

cmd/app/add.go
mwbrooks: > 👾 issue(non-blocking): Hmm I think we might still want to skip this section if an app ID is provided since the environment would be known?

Thanks, after looking over it, I think you're right too. Commit 0e65092 & e28a02e now skips when the `--app <id>` is provided. If `--app <env>` then the `--environment <env>` is set in the lines above.

Can't wait until we can move this logic out of the commands and to a higher-level. 😅

mwbrooks: > 🤖 question: It's not so clear to me if an `install` can target different environments for a known app. This might not be an issue if so, but I'm wondering what you think.

An App ID can only be associated with one environment (1-to-1 relationship), so we don't want to allow the developer to override that. Right now, we error if the `--app` and `--environment` flags are used together. That seems the safest. A better approach is that we don't error if `--environment` matches the saved `--app`, but atm the error message asks you to remove the `--environment` flag.

PR #165, opened 2025-07-23
feat: output warning of deno removal from installation scripts

zimeg:
> **Changelog**
>
> Starting September 1, 2025 the Slack CLI installation script will no longer include `deno` as part of the installation. We recommend installing `deno` separately and configuring CI systems as needed if `deno` is used for application development, such as for ROSI apps.

### Summary

This PR outputs a warning that `deno` will be removed from installation scripts at a future date: September 1, 2025.

### Preview

warning

### Requirements

* [x] I've read and understood the [Contributing Guidelines](/blob/main/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).

PR #167, opened 2025-07-25
feat: error if the runtime is not found with a missing hook

zimeg:
> **Changelog**
>
> We added a new `runtime_not_found` error code that appears when a hook script cannot be found because the runtime for a project wasn't found during execution.

### Summary

This PR errors with a `runtime_not_found` error if the hook `runtime` doesn't exist when a `sdk_hook_not_found` error is returned.

With this change, we make use of [hook resolution](https://github.com/slackapi/slack-cli/blob/main/.github/cli-sdk-specification.md#-hook-resolution) and assume that if the runtime isn't set, that wasn't returned in the response from `get-hooks`.

In sample applications and with `init` we suggest gathering hook values from the `get-hooks` script and don't override the `runtime` value. IMO this is good for the happy path cases! Additional hook customizations might run into unexpected errors with a customized `runtime` but debugging information remains available in verbose outputs 👾 

### Preview

🔍 Errors when a hook is not found **and** the runtime doesn't exist follow:

runtime

🔍 Errors when a hook is found but errors, without a runtime, continue to error:

hook

### Reviewers

Uninstall various runtimes and attempt to interact with a matching app:

```
$ slack create asdf
$ cd asdf
$ slack manifest
```

### Notes

- [ ] A follow up to **not** error if the `get-hooks` script errors is needed so that other hooks can still be parsed, if defined.

### Requirements


cmd/root.go
```diff
@@ -376,20 +376,23 @@ func ExecuteContext(ctx context.Context, rootCmd *cobra.Command, clients *shared
...
-		}
 		if slackerror.Is(err, slackerror.ErrProcessInterrupted) {
 			clients.IO.SetExitCode(iostreams.ExitCancel)
-			clients.IO.PrintDebug(ctx, errMsg)
+			clients.IO.PrintDebug(ctx, err.Error())
 		} else {
+			if slackerror.Is(err, slackerror.ErrSDKHookNotFound) && clients.SDKConfig.Runtime == "" {
+				err = slackerror.New(slackerror.ErrRuntimeNotFound).
+					WithRootCause(slackerror.ToSlackError(err).WithRemediation(""))
+			}
```
zimeg, line 386: 🔏 note: The `runtime` is unknown to actual hook scripts, so the `runtime_not_found` error can't be determined during hook execution itself...

🤔 question: Does this error handling logic make sense to include at the root command cleanup, or should hook executions be checked for this?

cmd/root.go
```diff
@@ -376,20 +376,23 @@ func ExecuteContext(ctx context.Context, rootCmd *cobra.Command, clients *shared
...
 			switch clients.IO.GetExitCode() {
 			case iostreams.ExitOK:
 				clients.IO.SetExitCode(iostreams.ExitError)
 			}
-			clients.IO.PrintError(ctx, errMsg)
+			clients.IO.PrintError(ctx, err.Error())
+		}
+		clients.EventTracker.SetErrorMessage(err.Error())
+		if slackErr, ok := err.(*slackerror.Error); ok {
+			clients.EventTracker.SetErrorCode(slackErr.Code)
```
zimeg, line 395: 🔍 note: Moved this since we now might update the `err` value!

zimeg:
🗣️ note: This is followup in preparation for changes from #165!

cmd/triggers/create.go
```diff
@@ -307,7 +308,9 @@ func triggerRequestFromFlags(flags createCmdFlags, isDev bool) api.TriggerReques
 
 func triggerRequestViaHook(ctx context.Context, clients *shared.ClientFactory, path string, isDev bool) (api.TriggerRequest, error) {
 	if !clients.SDKConfig.Hooks.GetTrigger.IsAvailable() {
-		return api.TriggerRequest{}, slackerror.New(slackerror.ErrSDKHookGetTriggerNotFound)
```
mwbrooks, line 309: praise: I like that you've chosen to remove this very-specific hook error for the general `ErrSDKHookNotFound` and Remediation combination.

zimeg: thought: Once errors across packages are wrapped I'm optimistic these same details will be clear in output messages 🙏

PR #176, opened 2025-08-06
Docs: updates links to point to `docs.slack.dev/tools*` paths

lukegalbraithrussell:
### Summary

tools pages are being directed to docs.slack.dev/tools moving forward. This readies the links for that

### Requirements


docs/guides/installing-the-slack-cli-for-mac-and-linux.md
```diff
@@ -23,10 +23,10 @@ If you attempt to log in with an older version, you'll receive a `cli_update_req
 **Run the automated installer from your terminal window:**
 
 ```zsh
-curl -fsSL https://downloads.slack-edge.com/slack-cli/install.sh | bash
+curl -fsSL https://downloads.slack-edge.com/tools/slack-cli/install.sh | bash
```
mwbrooks: I don't think we want to change this URL because our download URL is not changing.

mwbrooks (changes requested):
Just dropping a quick change request to avoid this being merged by accident. We'll want to undo the changes to the download URLs.

mwbrooks:
👍🏻 Everything else is looking good! We'll just want to undo that one line and then we can merge away!

docs/guides/installing-the-slack-cli-for-mac-and-linux.md
lukegalbraithrussell: ah yes my bad i thought i undid all these

PR #185, opened 2025-08-11
fix: URL checker uses HEAD instead of GET

mwbrooks:
### CHANGELOG

> Fix a bug to allow new projects to be created quicker by using a HTTP HEAD request to check if the template URL exists.

### Summary

This pull request updates the `URLChecker(http, url)` to check if the URL exists using a `HEAD` request instead of a `GET` request. This should improve performance, since it will not download the zip to only confirm that the URL is valid.

There should be no change to functionality.

### Reviewers

```bash
$ lack create asdf
# → Select any app, since all apps are downloaded from template URLS
# → Confirm that the app is created

# Cleanup
$ rm -rf asdf/
```

### Requirements

### DIFF
--- a/internal/deputil/url.go
+++ b/internal/deputil/url.go
@@ -22,7 +22,7 @@ import (
 
 // URLChecker returns url if its status code is 200, otherwise returns empty string
 func URLChecker(httpClient slackhttp.HTTPClient, url string) string {
-	resp, err := httpClient.Get(url)
+	resp, err := httpClient.Head(url)
 	if err != nil {
 		return ""
 	}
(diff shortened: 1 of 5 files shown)


zimeg (approved):
@mwbrooks Super nice find! I'm so glad we're enhancing the `slackhttp` package with these changes 🎁 ✨ 

LGTM and testing works well without `git` available:

```
$ PATH="" ./bin/slack create -t slack-samples/bolt-js-starter-template
...
📋 Next steps to begin development
   Change into your project with cd jolly-koala-551/
```

```sh
$ PATH="" ./bin/slack create -t slack-samples/bolt-unknown-starter-app
...
🚫 Cannot download Git repository as a .zip archive (git_zip_download_error)
```

📬 suggestion(non-blocking): Updating the error message to `git_repo_not_found` or with similar messaging might be interesting!

🚢 note: Please feel free to merge this when the time is right!

mwbrooks:
Thanks for the quick review @zimeg!

> 📬 suggestion(non-blocking): Updating the error message to git_repo_not_found or with similar messaging might be interesting!

Interesting suggestion! This error is actually coming from the caller of `URLChecker(http, url)` so let's hold off and change that error code in another PR so it appears as a separate CHANGELOG entry. 📝

PR #192, opened 2025-08-13
feat: output a list of samples for a language with the samples command

zimeg:
### Changelog

All available samples can be listed with the `slack samples --list` command. The `--language` flag might be found useful in filtering these outputs.

### Summary

This PR adds a `--list` flag to the `samples` command.

### Preview

📸

samples

### Reviewers

The following filters might be useful in experimentation! 🦠 

```sh
$ slack samples --list
$ slack samples --list --language deno
$ slack samples --list --language java
$ slack samples --list --language node
$ slack samples --list --language bolt-js
$ slack samples --list --language bolt-ts
$ slack samples --list --language python
$ slack samples --list --language go
```

### Requirements


cmd/project/samples.go
```diff
@@ -60,7 +69,18 @@ func runSamplesCommand(clients *shared.ClientFactory, cmd *cobra.Command, args [
 	sampler := api.NewHTTPClient(api.HTTPClientOptions{
 		TotalTimeOut: 60 * time.Second,
 	})
-	selectedSample, err := PromptSampleSelection(ctx, clients, sampler)
+	samples, err := create.GetSampleRepos(sampler)
+	if err != nil {
+		return err
+	}
+	if samplesListFlag || !clients.IO.IsTTY() {
```
zimeg, line 77: 👁️‍🗨️ note: We avoid attempting prompts in a non-interactive environment and instead list samples!

🗣️ note: The changes of #193 make this behavior more clear IMO since flag substitutes will then have matching prompts **or** the `create` command should be used.

cmd/project/samples.go
```diff
@@ -50,6 +58,7 @@ func NewSamplesCommand(clients *shared.ClientFactory) *cobra.Command {
 	cmd.Flags().StringVarP(&samplesTemplateURLFlag, "template", "t", "", "template URL for your app")
 	cmd.Flags().StringVarP(&samplesGitBranchFlag, "branch", "b", "", "name of git branch to checkout")
 	cmd.Flags().StringVar(&samplesLanguageFlag, "language", "", "runtime for the app framework\n  ex: \"deno\", \"node\", \"python\"")
+	cmd.Flags().BoolVar(&samplesListFlag, "list", false, "print recommended samples")
```
mwbrooks: Minor suggestion to clarify the usage:

```suggestion
	cmd.Flags().BoolVar(&samplesListFlag, "list", false, "prints samples without interactivity")
```

cmd/project/samples.go
```diff
@@ -85,3 +105,39 @@ func runSamplesCommand(clients *shared.ClientFactory, cmd *cobra.Command, args [
 	// Execute the `create` command with the set flag
 	return createCmd.ExecuteContext(ctx)
 }
+
+// listSampleSelection outputs available samples matching a language flag filter
+func listSampleSelection(ctx context.Context, clients *shared.ClientFactory, sampleRepos []create.GithubRepo) error {
+	filteredRepos := filterRepos(sampleRepos, samplesLanguageFlag)
+	sortedRepos := sortRepos(filteredRepos)
```
mwbrooks, line 113: question: Are you noticing that the repos aren't sorted?

Image

cmd/project/samples.go
```diff
@@ -85,3 +105,39 @@ func runSamplesCommand(clients *shared.ClientFactory, cmd *cobra.Command, args [
...
+		Secondary: []string{
+			message,
+		},
+	}))
+	emojis := []string{
+		"microscope",
+		"test_tube",
+		"petri_dish",
+		"dna",
+	}
```
mwbrooks: thought: While the emojis are a nice touch, I found them a little confusing. I spent some time trying to understand the meaning behind the emoji. I think it would be better to not include the emojis.

mwbrooks (commented):
📝 Nice feature! I personally enjoy the flat list more than the interactive one.

❓ Curious where this feature idea came from?

🧪 While testing, I noticed that the order isn't sorted on my side. Is that happening on your side too? Also, I left a suggestion to remove the emojis since they don't have a strong meaning.

cmd/project/samples.go
zimeg: @mwbrooks This confused me too but these might find order from the stars? ⭐ 

I wonder if showing the number of stars would be helpful? I might expect this to prefix the sample, but am not sure at all if that'd be right:

```
⭐ 60 - deno-timesheet-approval
   Collect timesheet information from users and store it in a Google Sheet
   https://github.com/slack-samples/deno-timesheet-approval
```

cmd/project/samples.go
zimeg: @mwbrooks Oops! Returning to this now! It's confusing without obvious meaning... I agree! This was changed to use different emojis in https://github.com/slackapi/slack-cli/pull/192/commits/da1d1051a4d2715591c37f2f75150f562a270cf0 that separate "template" apps from "example" samples:

samples

I'm curious what no emojis would look like though? I found this "column" ordering to be perhaps confusing in context with long lines. Perhaps we revisit app descriptions later? Or change these outputs elsewise?

zimeg:
> ❓ Curious where this feature idea came from?

@mwbrooks Thanks for reviewing and sharing suggestions alike! I made some updates in above commits and comments that reorder outputs to make things more clear I hope.

I was hoping for a quick filter of current samples for similar outputs as the pages of [docs.slack.dev/samples](https://docs.slack.dev/samples/) 🤖

Although I notice now we filter moreso by runtime without a "sample" and "template" option. I think this is alright to continue with but am open to all suggestion! Will rerequest a kind review! 🙏 ✨

cmd/project/samples.go
mwbrooks: Thanks for taking the time to look back into this! Since the `samples` command has low usage, I think we can start by including the lengthy descriptions and revisit it later.

mwbrooks (approved):
> I'm curious what no emojis would look like though? 

Personally, I'd lean toward no emojis. I think the nuances between examples and templates is difficult to understand. They're all sample apps and the more important filter is the framework language.

My two cents is to remove the emojis all together or use the same emoji for all samples.

✅ Tossing an approval on this one. Overall, it's moving the `samples` command in a better direction. Thanks @zimeg!

zimeg:
@mwbrooks Once again, thanks for the kind review 🙏 ✨ 

> My two cents is to remove the emojis all together or use the same emoji for all samples.

As a fast change to use one emoji I put together https://github.com/slackapi/slack-cli/pull/192/commits/d2e22def9efe21d0aed75be19f37036e6c4b66bd but I'm hoping we can revisit outputs in future iteration 👾 

Let's merge this now with follow up later! 🚢

PR #193, opened 2025-08-13
feat: deprecate the "branch" and "template" flag for the samples command

zimeg:
### Changelog

We've deprecated the `--branch` and `--template` flag for the `slack samples` command. If those were used prior, we recommend using the `slack create` command with these options.

### Summary

This PR deprecates the the `--branch` and `--template` flags of the `slack samples` command.

### Preview

```
$ slack samples -t slack-samples/bolt-js-getting-started-app
Check /Users/eden.zimbelman/.slack/logs/slack-debug-20250813.log for error logs

⚠️  DEPRECATED: The '--branch' and '--template' flags are deprecated for the 'samples' command; use the 'create' command instead

📂 Created a new Slack project
   Cloning template slack-samples/bolt-js-getting-started-app
   To path ~/programming/tools/slack-cli/optimistic-orca-30

...
```

### Reviewers

Confirm the `samples` command continues to work as expected:

```sh
$ slack samples --list --language node
$ slack samples my-app
```

### Requirements


cmd/project/samples.go
```diff
@@ -66,6 +71,12 @@ func NewSamplesCommand(clients *shared.ClientFactory) *cobra.Command {
 // runSamplesCommand prompts for a sample then clones with the create command
 func runSamplesCommand(clients *shared.ClientFactory, cmd *cobra.Command, args []string) error {
 	ctx := cmd.Context()
+
+	// DEPRECATED(semver:major): Prefer the create command when repository details are known
+	if cmd.Flag("branch").Changed || cmd.Flag("template").Changed {
+		clients.IO.PrintWarning(ctx, "DEPRECATED: The '--branch' and '--template' flags are deprecated for the 'samples' command; use the 'create' command instead")
```
mwbrooks: suggestion: Based on https://github.com/slackapi/slack-cli/pull/194, do you think we should start to use backticks instead of single quotes for the flags and commands?
```suggestion
		clients.IO.PrintWarning(ctx, "DEPRECATED: The `--branch` and `--template` flags are deprecated for the `samples` command; use the `create` command instead")
```

cmd/project/samples.go
```diff
@@ -55,8 +55,13 @@ func NewSamplesCommand(clients *shared.ClientFactory) *cobra.Command {
 		},
 	}
 
-	cmd.Flags().StringVarP(&samplesTemplateURLFlag, "template", "t", "", "template URL for your app")
+	// DEPRECATED(semver:major): Prefer the create command when repository details are known
 	cmd.Flags().StringVarP(&samplesGitBranchFlag, "branch", "b", "", "name of git branch to checkout")
+	cmd.Flag("branch").Hidden = true
```
mwbrooks, line 61: thought: FWIW I think `--branch` has a legit use-case with the `slack samples` command. However, I get the idea that `slack samples` is meant to discover the available templates and isn't meant to target advanced use-cases. The `slack create` command is designed to handle the general-purpose and advanced ones.

mwbrooks (approved):
✅ Looks good to me and thanks for adding the `(semver:major)` comments and deprecation warning!

📝 I left a small suggestion and question. FWIW I'm in favour of deprecated both of these flags to simplify the code and command usage. 👌🏻

cmd/project/samples.go
zimeg: @mwbrooks Thanks for sharing this - I was thinking the same to recommend the `create` command when creating a specific branch.

I'm wondering if we can revisit this in following change to show samples that have an existing branch when this flag is provided to avoid errors during a `git` clone?

Will be thinking about this and related docs! 📚

mwbrooks: Sounds good! All in all, I'm in favour of a simpler `samples` command and leaving the advanced usage to `create`.

PR #194, opened 2025-08-14
fix: generate documentation without template literals and with standard headers

zimeg:
### Summary

This PR removes template literals and changes headers of generated documentation to match previous build scripts used for upstream pages 📚 ✨ 

### Reviewers

With these changes checked out, confirm expected updates to reference are found:

```sh
$ make build && PATH="./bin" slack docgen docs/reference
$ git diff
```

### Requirements


cmd/docgen/docgen.go
```diff
@@ -124,3 +128,153 @@ func runDocGenCommandFunc(clients *shared.ClientFactory, cmd *cobra.Command, arg
...
+	if cmd.Runnable() {
+		fmt.Fprintf(buf, "```\n%s\n```\n\n", cmd.UseLine())
+	}
+	if len(cmd.Example) > 0 {
+		fmt.Fprintf(buf, "## Examples\n\n")
+		fmt.Fprintf(buf, "```\n%s\n```\n\n", cmd.Example)
+	}
+	if err := genMarkdownCommandFlags(buf, cmd); err != nil {
+		return err
+	}
```
zimeg, line 186: 👁️‍🗨️ note: This ordering might be updated in a follow up PR to match `--help` outputs.

internal/slackerror/errors.go
```diff
@@ -374,7 +375,7 @@ Otherwise start your app for local development with: %s`,
 	ErrAppFlagRequired: {
 		Code:        ErrAppFlagRequired,
 		Message:     "The --app flag must be provided",
-		Remediation: fmt.Sprintf("Choose a specific app with %s", style.Highlight("--app <app_id>")),
+		Remediation: "Choose a specific app with `--app <app_id>`",
```
zimeg, line 378: ⚠️ note: Backticks are preferred to `Highlight` since styles are not found in these error messages - the error map is compiled without styles toggled "on".

🐣 note: This matches the fallback behavior of `Commandf` and ought not cause problems in output. An example command might use "`--app ...`" as an option to show this.

zimeg (commented):
💌 Notes on updates follow for the kind reviewers but also please find in an expected `diff`:

- "Synopsis" is changed to "Description"
- "Options" is changed to "Flags"
- "Options inherited from parent commands" is changed to "Global flags"
- "SEE ALSO" is changed to "See also"

cmd/docgen/docgen.go
zimeg: 📣 note: This change is made in https://github.com/slackapi/slack-cli/pull/195!

internal/shared/clients.go
```diff
@@ -66,14 +64,6 @@ type ClientFactory struct {
...
 	// CleanupWaitGroup is a group of wait groups shared by all packages and allow functions to cleanup before the process terminates
 	CleanupWaitGroup sync.WaitGroup
-
-	// Cobra are a group of Cobra functions shared by all packages and enables tests & mocking
-	Cobra struct {
-		// GenMarkdownTree defaults to `doc.GenMarkdownTree(...)` and can be mocked to test specific use-cases
-		// TODO - This can be moved to cmd/docs/docs.go when `NewCommand` returns an instance of that can store `GenMarkdownTree` as
-		//        a private member. The current thinking is that `NewCommand` would return a `SlackCommand` instead of `CobraCommand`
-		GenMarkdownTree func(cmd *cobra.Command, dir string) error
-	}
```
mwbrooks, line 76: 🪓 🎉

cmd/docgen/docgen.go
```diff
@@ -124,3 +128,153 @@ func runDocGenCommandFunc(clients *shared.ClientFactory, cmd *cobra.Command, arg
...
+	if hasSeeAlso(cmd) {
+		fmt.Fprintf(buf, "## See also\n\n")
```
mwbrooks, line 188: ignore-me: While I prefer "Title Case", this looks correct because our docs use "Sentence case". I think it's best for the CLI to stay aligned with our docs. So please ignore me! 🧠

lukegalbraithrussell (approved):
i'm hyped!

PR #195, opened 2025-08-14
docs: list flags before examples in generated reference

zimeg:
### Summary

This PR lists flags before examples in generated reference to match outputs of the `--help` commands 🤖

### Preview

The following page and command might demonstrate the purpose of this change:

- Docs: https://docs.slack.dev/tools/slack-cli/reference/commands/slack_app_link
- CLI: `slack app link --help`

### Requirements

* [x] I've read and understood the [Contributing Guidelines](/blob/main/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).

PR #197, opened 2025-08-15
docs(ci): adds generated reference workflow

lukegalbraithrussell:
### Summary

This PR adds a manual dispatch workflow to be run after release to update the CLI reference docs

This PR also manually updates the CLI reference docs using the recently updated version of `docgen` now in `main`.

I have a PR preview of this but idk if i should be dropping the link in a public PR hmmm

### Requirements


.github/workflows/slack-cli-generate-docs.yaml
```diff
@@ -0,0 +1,43 @@
+name: Generate Slack CLI Docs
+
+on:
+  workflow_dispatch:
```
zimeg, line 4: 👁️‍🗨️ thought: Adding on "version" tags might be interesting! Otherwise we will need to add this to our release runbooks.

📣 ramble: So open to saving this for follow ups of course.

.github/workflows/slack-cli-generate-docs.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+    steps:
+
+    - name: Install Slack CLI
+      run: |
+        curl -fsSL https://downloads.slack-edge.com/slack-cli/install.sh | bash
+
+    - name: Generate a GitHub token
+      id: ghtoken
+      uses: actions/create-github-app-token@v2
```
zimeg, line 18: 💔 todo: We should use pinned versions here though I don't think the healthscore has released such a check yet...

📫 note: I can soon add these versions if needed!

.github/workflows/slack-cli-generate-docs.yaml
```diff
@@ -0,0 +1,43 @@
+name: Generate Slack CLI Docs
+
+on:
+  workflow_dispatch:
+
+jobs:
+  update-docs:
```
zimeg, line 7: 🔐 todo: This'll also need some set of permissions. Do let me know if I can help with this!

zimeg (changes requested):
📚 LGTM! Thanks for following up on #194 and #195!

I left a few comments around the workflow that IMHO should be addressed but otherwise docs changes are looking great. We could perhaps make a separate PR for the generated reference workflow too?

.github/workflows/slack-cli-generate-docs.yaml
zimeg: 🗣️ note: Ugh that was more clear in mind... I meant to suggest "on new tagged releases the workflow should run" but I forget the syntax!

zimeg:
🗣️ Removed this from the next milestone for now. I think we want to refine our release process to autogenerate docs during the **tag**? IIRC suggestion has been floated for:

```
$ make release v3.4.5
```

And perhaps this can be included in workflows somehow in the future?

zimeg:
@lukegalbraithrussell Thanks for starting conversation on how we keep documentation current with a changing release process!

A change to use the following command landed in #228 for improving how we update installation links and #229 was opened to include similar `docgen` updates:

```
$ make tag RELEASE_VERSION=3.1.4
```

I'm so curious if we can follow up on these to use PRs as part of our release process automations - let's continue these explorations? 👾

zimeg:
🗣️ Wanting to keep track of our progress in automating these updates! The above PRs were a success in the most recent release [v3.8.0](https://github.com/slackapi/slack-cli/commit/7ec14709e8c229344d19c852f64e05604b4fa99e)!

Some follow up seems needed still though:

- Pushing to `main` skips perhaps useful checks that CI reviews give. We can perhaps have a bot open these changes as PRs? I'm unsure if tagging can still be automated though...
- Running `docgen` at the time of release seems late IMHO. It'd be most excellent if these changes were included with the related PR since we can now **pin** documentation to a tag!

I'm open to keeping this PR around for more discussion on this topic!

zimeg:
@lukegalbraithrussell Kind thanks for keeping us honest with these reference let's close this with #553 patterns in releases 🔮 ✨

PR #210, opened 2025-09-02
feat: add search app feature to the app manifest

zimeg:
### Changelog

> We now parse the app manifest fields of `features.search` when gathering an app manifest.

### Summary

This PR adds the `search` attribute to the app manifest 👾 ✨ 

### Preview

The following can be added to a manifest as a proof of concept:

```json
{
  "features": {
    "search": {
      "search_function_callback_id": "movie_search",
      "search_filters_function_callback_id": "movie_filters",
      "enable_ai_answers": true
    }
  }
}
```

### Reviewers

Confirm this value is detected when parsing the manifest:

```sh
$ slack create asdf              # Create a Bolt JS application
$ cd asdf
$ vim manifest.json              # Update the app manifest
$ slack manifest --source local  # Find expected values appear!
```

### Requirements

### DIFF
--- a/internal/shared/types/app_manifest.go
+++ b/internal/shared/types/app_manifest.go
@@ -75,6 +75,7 @@ type AppFeatures struct {
 	UnfurlDomains              []string                    `json:"unfurl_domains,omitempty" yaml:"unfurl_domains,flow,omitempty"`
 	ManifestShortcutsItems     []ManifestShortcutsItem     `json:"shortcuts,omitempty" yaml:"shortcuts,flow,omitempty"`
 	ManifestSlashCommandsItems []ManifestSlashCommandsItem `json:"slash_commands,omitempty" yaml:"slash_commands,flow,omitempty"`
+	Search                     *Search                     `json:"search,omitempty" yaml:"search,flow,omitempty"`
 }
 
 type AssistantView struct {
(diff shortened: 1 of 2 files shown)


internal/shared/types/app_manifest.go
```diff
@@ -252,6 +253,13 @@ type ManifestSlashCommandsItem struct {
 	UsageHint    string `json:"usage_hint,omitempty" yaml:"usage_hint,omitempty"`
 }
 
+// Search contains the search function of an app.
+type Search struct {
+	SearchFunctionCallbackID        string `json:"search_function_callback_id,omitempty" yaml:"search_function_callback_id,omitempty"`
+	SearchFiltersFunctionCallbackID string `json:"search_filters_function_callback_id,omitempty" yaml:"search_filters_function_callback_id,omitempty"`
+	EnableAIAnswers                 *bool  `json:"enable_ai_answers,omitempty" yaml:"enable_ai_answers,omitempty"`
```
mwbrooks, line 260: thought: Don't love the `*bool` pointer but we have used this in other manifest properties, so I think it's acceptable. Is there a default value that the Slack API uses for this field? If not, totally understand that the bool pointer is the right choice. It's just a little dangerous to access because it can lead to runtime segfaults.

mwbrooks (approved):
✅ Looks good to me!

🧪 Manual testing works well. The manifest properties are returned and validate.

📝 Left a minor comment about the bool pointer, but I think we can merge as-is.

cchensh (approved):
Thank you very much for the quick update @zimeg :hero

internal/shared/types/app_manifest.go
zimeg: @mwbrooks I felt a similar hesitation but found a `bool` value would default to `false` and caused the included JSON `omitempty` attribute to not take effect.

Super agree that this can cause runtime issues later. I wonder if a linter can catch accesses to `nil` values? 👾

zimeg (commented):
@mwbrooks @cchensh Thank y'all both for the quick and kind reviews! 🎁 

For now this will let us continue to use the CLI for app updates involving this `search` parameter. I didn't include entries in the CLI changelog for this since I don't believe it to be a user-facing change, but please let me know if we should revisit that 📚 ✨

PR #247, opened 2025-10-21
feat: update the manifest model to include the setting for work objects

vegeris:
### Changelog

The app manifest values for `entity_types` are now gathered when reading an app manifest.

### Summary

This PR updates the app manifest model to include work object settings.

#### Test

- Before: an app created via the CLI does not include work object settings (review the manifest in App Settings)
- After: the created app does include work object settings (review the manifest in App Settings)

```
slack create first-bolt-app --template slack-samples/bolt-js-getting-started-app
cd first-bolt-app
# Add the following to the manifest under `features`:
#  "rich_previews": {
#         "entity_types": [
#               "slack#/entities/file"
#         ]
#   }

slack run
```

### Requirements

### DIFF
--- a/internal/shared/types/app_manifest.go
+++ b/internal/shared/types/app_manifest.go
@@ -81,6 +81,7 @@ type AppFeatures struct {
 
 type RichPreviews struct {
 	EntityTypes []string `json:"entity_types,omitempty" yaml:"entity_types,flow,omitempty"`
+	IsActive    bool     `json:"is_active,omitempty" yaml:"is_active,flow,omitempty"`
 }
 
 type AssistantView struct {


internal/shared/types/app_manifest.go
```diff
@@ -76,6 +76,11 @@ type AppFeatures struct {
 	ManifestShortcutsItems     []ManifestShortcutsItem     `json:"shortcuts,omitempty" yaml:"shortcuts,flow,omitempty"`
 	ManifestSlashCommandsItems []ManifestSlashCommandsItem `json:"slash_commands,omitempty" yaml:"slash_commands,flow,omitempty"`
 	Search                     *Search                     `json:"search,omitempty" yaml:"search,flow,omitempty"`
+	RichPreviews               *RichPreviews               `json:"rich_previews,omitempty" yaml:"rich_previews,flow,omitempty"`
+}
+
+type RichPreviews struct {
+	EntityTypes []string `json:"entity_types,omitempty" yaml:"entity_types,flow,omitempty"`
```
zimeg, line 83: 📝 though: The `omitempty` attribute worries me with slices since it's not obvious how to keep an empty list of entity types, but I notice the API errors in that case so no worries!

internal/shared/types/app_manifest_test.go
```diff
@@ -179,6 +179,17 @@ func Test_AppManifest_AppFeatures(t *testing.T) {
...
 			want: `{"app_home":{},"bot_user":{"display_name":"kubrick"},"search":{"search_function_callback_id":"movie_search","search_filters_function_callback_id":"movie_filters","enable_ai_answers":true}}`,
 		},
+		"includes Work Objects settings when provided": {
+			features: AppFeatures{
+				BotUser: BotUser{
+					DisplayName: "business_bot",
+				},
+				RichPreviews: &RichPreviews{
+					EntityTypes: []string{"slack#/entities/file"},
+				},
```
zimeg, line 189: 📠 praise: Thanks for including test, as always!

internal/shared/types/app_manifest.go
```diff
@@ -76,6 +76,12 @@ type AppFeatures struct {
...
+	IsActive    bool     `json:"is_active,omitempty" yaml:"is_active,flow,omitempty"`
```
vegeris, line 84: I was on the fence about including this property but after some discussion, we might want to support the dev setting this rather than needing to remove the entire rich_previews setting to disable (so I added it back)

internal/shared/types/app_manifest.go
mwbrooks: Yea, this is the common struggle of matching the API defaults with Golang language rules. Seems like we'll want to keep the `omitempty` to align with the API.

PR #172, opened 2025-07-30
test: run the windows and unix test suites on release tags

zimeg:
### Summary

This PR runs both the Windows and Unix E2E tests after a release tag is made in hopes of reducing test steps with the release process.

### Notes

Samples continue to run tests with changes on the sample `main` branch unless configured elsewhere.

### Requirements

* [x] I've read and understood the [Contributing Guidelines](/blob/main/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).

PR #255, opened 2025-10-31
test: run the windows tests scheduled and on main branch instead of tags

zimeg:
### Summary

This PR attempts to run the E2E tests on a Windows machine at a cadenced schedule and after merges to `main` for an improved release process. Follows: #172.

### Notes

We remove attempts of testing after a **tag** for two reasons:

- That test workflow ran adjacent to release publishing and also attempted to update the GitHub release which could overwrite artifacts.
- We avoid running these tests during the publishing steps to reduce flakes that can cause a failed build process!

### Requirements

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -554,6 +554,15 @@ workflows:
           context: slack-cli-release
           release_ref: dev-build
       - e2e-test:
+          name: e2e-test-unix
+          manual_trigger_windows: false
+          requires:
+            - create-github-release-and-artifacts
+          context: slack-cli-e2e
+          release_ref: dev-build
+      - e2e-test:
+          name: e2e-test-windows
+          manual_trigger_windows: true
           requires:
             - create-github-release-and-artifacts
           context: slack-cli-e2e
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -695,34 +713,3 @@ workflows:
           requires:
             - create-github-release-and-artifacts
           context: slack-cli-release
-
-  # production tests are run separate from deployment to avoid errors with tags.
-  prod-build-test-e2e-test:
```
mwbrooks, line 700: question: Is this removed because it's unused?

mwbrooks (approved):
✅ Yay, this is exciting! I like the approach of running Unix/Windows tests on merges to main and on a regular schedule because it aligns with how we do other testing. 👌🏻 Looking forward to seeing this in action!

.circleci/config.yml
zimeg: @mwbrooks Thanks for asking! We might've found a race condition between the release uploads of this workflow and our actual publishing one...

Since we're testing things earlier now I thought it'd be faster to remove this instead of debug more 🐛 ✨

PR #272, opened 2025-11-20
feat: update 'collaborator' command to support read-only permission type

vegeris:
### Summary

A while back we added support for setting some app collaborators as read-only for deployed apps but never released the experiment

### Requirements

### DIFF
--- a/cmd/app/app.go
+++ b/cmd/app/app.go
@@ -36,6 +36,7 @@ func NewCommand(clients *shared.ClientFactory) *cobra.Command {
 			{Command: "app list", Meaning: "List all teams with the app installed"},
 			{Command: "app settings", Meaning: "Open app settings in a web browser"},
 			{Command: "app uninstall", Meaning: "Uninstall an app from a team"},
+			{Command: "app unlink", Meaning: "Remove a linked app from the project"},
 			{Command: "app delete", Meaning: "Delete an app and app info from a team"},
 		}),
 		Args: cobra.NoArgs,
@@ -69,6 +70,7 @@ func NewCommand(clients *shared.ClientFactory) *cobra.Command {
 	cmd.AddCommand(NewListCommand(clients))
 	cmd.AddCommand(NewSettingsCommand(clients))
 	cmd.AddCommand(NewUninstallCommand(clients))
+	cmd.AddCommand(NewUnlinkCommand(clients))
 
 	return cmd
 }
--- a/cmd/app/unlink.go
+++ b/cmd/app/unlink.go
@@ -1,0 +1,137 @@
+// Copyright 2022-2025 Salesforce, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package app
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/slackapi/slack-cli/internal/cmdutil"
+	"github.com/slackapi/slack-cli/internal/prompts"
+	"github.com/slackapi/slack-cli/internal/shared"
+	"github.com/slackapi/slack-cli/internal/shared/types"
+	"github.com/slackapi/slack-cli/internal/slacktrace"
+	"github.com/slackapi/slack-cli/internal/style"
+	"github.com/spf13/cobra"
+)
+
+// Handle to function used for testing
+var unlinkAppSelectPromptFunc = prompts.AppSelectPrompt
+
+// NewUnlinkCommand returns a new Cobra command for unlinking apps
+func NewUnlinkCommand(clients *shared.ClientFactory) *cobra.Command {
+	var unlinkedApp types.App // capture app for PostRunE
+
+	cmd := &cobra.Command{
+		Use:   "unlink",
+		Short: "Remove a linked app from the project",
+		Long: strings.Join([]string{
+			"Unlink removes an existing app from the project.",
+			"",
+			"This command removes a saved app ID from the files of a project without deleting",
+			"the app from Slack.",
+		}, "\n"),
+		Example: style.ExampleCommandsf([]style.ExampleCommand{
+			{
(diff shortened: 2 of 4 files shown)


cmd/collaborators/add.go
```diff
@@ -181,40 +180,7 @@ func promptCollaboratorsAddSlackUserPrompts(
 	return slackUser, nil
 }
 
-// promptCollaboratorsAddPermissionPrompts gathers the collaborator permission
-// from selection if the experiment allows
-func promptCollaboratorsAddPermissionPrompts(
```
vegeris, line 186: Changed the behaviour here: before when adding a new collaborator we required the permission type to be set, but I think we can default to 'owner' unless otherwise specified

cmd/collaborators/update.go
```diff
@@ -92,8 +80,10 @@ func runUpdateCommand(cmd *cobra.Command, clients *shared.ClientFactory, args []
 			return err
 		}
 	} else {
-		cmd.Println(fmt.Sprintf("\n%s Specify a permission type for your collaborator with the %s flag\n", style.Emoji("warning"), style.Highlight("--permission-type")))
-		return nil
+		slackUser.PermissionType, err = promptCollaboratorPermissionSelection(ctx, clients)
+		if err != nil {
+			return err
+		}
```
vegeris: When updating an existing collaborator however, maybe it would make sense to prompt for the permission type rather than error out

cmd/collaborators/add.go
```diff
@@ -62,8 +61,7 @@ func NewAddCommand(clients *shared.ClientFactory) *cobra.Command {
 			return runAddCommandFunc(ctx, clients, cmd, args)
 		},
 	}
-	cmd.Flags().StringVarP(&addFlags.permissionType, "permission-type", "P", "", "collaborator permission type: reader, owner")
-	cmd.Flag("permission-type").Hidden = true
+	cmd.Flags().StringVarP(&addFlags.permissionType, "permission-type", "P", "", "collaborator permission type: [reader|owner]")
```
mwbrooks: suggestion: Updating the formatting to follow the other commands (e.g. `manifest --source` or `activity --level`) and use the type values:

Image

```suggestion
	cmd.Flags().StringVarP(&addFlags.permissionType, "permission-type", "P", string(types.OWNER), fmt.Sprintf(
		"collaborator permission type\n(\"%s\" or \"%s\")",
		string(types.OWNER),
		string(types.READER),
	))
```

cmd/collaborators/add.go
mwbrooks: Sounds good!

Since we're now defaulting to the owner, we should update the description of the `--permission-type` to default to owner as well. I'll make a suggestion for that!

cmd/collaborators/update.go
mwbrooks: If we choose to prompt, than we should handle non-TTY (no interactivity) as well. In that case, I think we would error when there is no flag value provided.

suggestion: Personally, I suggest we keep the error when the flag is not provided because it's been tested and works. If we switch to a prompt, I'd rather see it as a separate PR because it changes the functionality rather than simply unhides the experiment (purpose of this PR).

cmd/collaborators/update.go
```diff
@@ -117,3 +107,29 @@ func runUpdateCommand(cmd *cobra.Command, clients *shared.ClientFactory, args []
...
+		"reader",
+	}
+	response, err := clients.IO.SelectPrompt(
+		ctx,
+		"Select a permission type",
+		permissionLabels,
+		iostreams.SelectPromptConfig{
+			Required: true,
+		},
+	)
```
mwbrooks: suggestion: I think this needs to be more robust. The `iostreams.SelectPromptConfig` should be bound to the flag and the flag, so that non-interactive environments can work properly with the prompt.

cmd/collaborators/update.go
```diff
@@ -51,23 +51,11 @@ func NewUpdateCommand(clients *shared.ClientFactory) *cobra.Command {
...
-				}))
-				return nil
-			}
-
 			return runUpdateCommand(cmd, clients, args)
 		},
 	}
 
-	cmd.Flags().StringVarP(&updateFlags.permissionType, "permission-type", "P", "", "collaborator permission type: reader, owner")
+	cmd.Flags().StringVarP(&updateFlags.permissionType, "permission-type", "P", "", "collaborator permission type: [reader|owner]")
```
mwbrooks: suggestion: Updating the format of this to match the `add` command but not setting a default value.

Image

```suggestion
	cmd.Flags().StringVarP(&updateFlags.permissionType, "permission-type", "P", "", fmt.Sprintf(
		"collaborator permission type\n(\"%s\" or \"%s\")",
		string(types.OWNER),
		string(types.READER),
	))```

**Review by mwbrooks — CHANGES_REQUESTED** [review]
🙇🏻 Thanks for unhiding and releasing the read-only-collaborators experiment!

✏️ The majority of the changes look good.

* I've left 2 suggestions that we should merge to display the `--permission-type` description consistent with other commands.
* I've marked this as **Request changes** because I don't think we should introduce a prompt for permission types when updating collaborators. I'd a nice UX improvement, but it has a high risk of introducing bugs for non-interactivity environments. I'd rather see this added as a separate PR that we can review, test, and feel confident in.

File: cmd/collaborators/add.go (hunk 3 above)
Comment (vegeris) [reply cmd/collaborators/add.go]: Sure; I originally used this format to be consistent with this example showing the email vs. user_id option: 
`$ slack collaborator add [email|user_id] [flags]` 😛

File: cmd/collaborators/update.go (hunk 2 above)
Comment (vegeris) [reply cmd/collaborators/update.go]: Sounds good; I'll flip this back to erroring out if the flag isn't provided. I'm fine with keeping that behaviour if the CLI team doesn't see adding the prompt as an enhancement to the user experience

File: cmd/collaborators/update.go (hunk 4 above)
Comment (vegeris) [reply cmd/collaborators/update.go]: Removed the prompt for now but will make note of this if we want to bring it back

File: internal/shared/types/permissions.go [hunk 6]
```diff
@@ -37,7 +37,7 @@ func StringToAppCollaboratorPermission(input string) (AppCollaboratorPermission,
 	case "reader":
 		return READER, nil
 	default:
-		return "", fmt.Errorf("invalid")
```
vegeris, line 40: Updated this error message to be more specific

cmd/collaborators/add.go
vegeris: Thanks!

mwbrooks (approved):
✅ Thanks a bunch @vegeris for releasing this feature out of experimentation! 🚀 

🧪 Manual tests and code review look good to me 👌🏻

PR #310, opened 2026-01-27
feat: restart the app development server on file changes 

zimeg:
### Changelog

> The `slack run` command supports file watching and live reloading Bolt for JavaScript and Bolt for Python projects. When a file is changed, the Slack CLI will automatically restart the app development server. When the `manifest.json` file is changed then the manifest update API is called (`.slack/config.json` must have `manifest.source: "local"`).

### Summary

This PR restarts the app development server on file changes when using the `slack run` command. An update to the hooks file adds options for this:

```json
{
  "config": {
    "watch": {
      "manifest": {
        "paths": ["manifest.json"]
      },
      "app": {
        "paths": ["app.js", "listeners/"],
        "filter-regex": "\\.(ts|js)$"
      }
    }
  }
}
```

### Preview

https://github.com/user-attachments/assets/f122d736-e851-430d-aa77-80f10ba83148

### Reviewers

The changes of this PR can be built with update hook defaults of https://github.com/slackapi/node-slack-sdk/pull/2480 👾 

```sh
$ npm pack                # Package the CLI hooks
$ slack create asdf -t slack-samples/bolt-js-starter-template
$ cd asdf
$ npm install /path/to/cli-hooks/slack-cli-hooks-1.2.1.tgz
$ vim .slack/config.json  # Set manifest source to be "local"
$ slack run
$ vim manifest.json       # Update the manifest to reinstall 
$ vim app.js              # Update the app to restart server
```

### Notes

- This change remains backward compatible with top-level `filter-regex` and `paths` for manifest updates. A follow up PR might add warnings for upcoming deprecation?

### Requirements

### DIFF
--- a/internal/hooks/sdk_config.go
+++ b/internal/hooks/sdk_config.go
@@ -95,6 +95,9 @@ func (w *WatchOpts) IsAvailable() bool {
 
 // GetManifestWatchConfig returns manifest watch config
 func (w *WatchOpts) GetManifestWatchConfig() (paths []string, filterRegex string, enabled bool) {
+	if w == nil {
+		return nil, "", false
+	}
 	if w.Manifest != nil {
 		return w.Manifest.Paths, w.Manifest.FilterRegex, len(w.Manifest.Paths) > 0
 	}
@@ -104,6 +107,9 @@ func (w *WatchOpts) GetManifestWatchConfig() (paths []string, filterRegex string
 
 // GetAppWatchConfig returns app watch config
 func (w *WatchOpts) GetAppWatchConfig() (paths []string, filterRegex string, enabled bool) {
+	if w == nil {
+		return nil, "", false
+	}
 	if w.App != nil {
 		return w.App.Paths, w.App.FilterRegex, len(w.App.Paths) > 0
 	}
--- a/internal/hooks/sdk_config_test.go
+++ b/internal/hooks/sdk_config_test.go
@@ -124,13 +124,19 @@ func Test_WatchOpts_IsAvailable(t *testing.T) {
 
 func Test_WatchOpts_GetManifestWatchConfig(t *testing.T) {
 	tests := map[string]struct {
-		watchOpts       WatchOpts
+		watchOpts       *WatchOpts
 		expectedPaths   []string
 		expectedRegex   string
 		expectedEnabled bool
 	}{
+		"Nil WatchOpts pointer": {
+			watchOpts:       nil,
+			expectedPaths:   nil,
+			expectedRegex:   "",
+			expectedEnabled: false,
+		},
 		"Nested manifest config": {
-			watchOpts: WatchOpts{
+			watchOpts: &WatchOpts{
 				Manifest: &ManifestWatchOpts{
 					Paths:       []string{"manifest.json", "workflows/"},
 					FilterRegex: "\\.json$",
@@ -141,7 +147,7 @@ func Test_WatchOpts_GetManifestWatchConfig(t *testing.T) {
 			expectedEnabled: true,
 		},
 		"Legacy flat config": {
-			watchOpts: WatchOpts{
+			watchOpts: &WatchOpts{
 				Paths:       []string{"manifest.json", "src/"},
 				FilterRegex: "\\.(json|ts)$",
 			},
@@ -150,7 +156,7 @@ func Test_WatchOpts_GetManifestWatchConfig(t *testing.T) {
 			expectedEnabled: true,
 		},
 		"Nested config takes precedence over legacy": {
-			watchOpts: WatchOpts{
+			watchOpts: &WatchOpts{
 				Paths:       []string{"old-path/"},
 				FilterRegex: "old-regex",
 				Manifest: &ManifestWatchOpts{
@@ -163,7 +169,7 @@ func Test_WatchOpts_GetManifestWatchConfig(t *testing.T) {
 			expectedEnabled: true,
 		},
 		"Empty nested manifest config": {
-			watchOpts: WatchOpts{
+			watchOpts: &WatchOpts{
 				Manifest: &ManifestWatchOpts{
 					Paths: []string{},
 				},
@@ -173,7 +179,7 @@ func Test_WatchOpts_GetManifestWatchConfig(t *testing.T) {
 			expectedEnabled: false,
 		},
 		"Empty legacy config": {
-			watchOpts: WatchOpts{
+			watchOpts: &WatchOpts{
 				Paths: []string{},
 			},
 			expectedPaths:   []string{},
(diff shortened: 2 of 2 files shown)


cmd/platform/run.go
```diff
@@ -178,11 +178,14 @@ func newRunLogger(clients *shared.ClientFactory, cmd *cobra.Command) *logger.Log
...
-				path := event.DataToString("cloud_run_watch_file_change")
-				cmd.Println(style.Secondary(fmt.Sprintf("File change detected: %s, reinstalling app...", path)))
-			case "on_cloud_run_watch_file_change_reinstalled":
+			case "on_cloud_run_watch_manifest_change":
+				path := event.DataToString("cloud_run_watch_manifest_change")
+				cmd.Println(style.Secondary(fmt.Sprintf("Manifest change detected: %s, reinstalling app...", path)))
+			case "on_cloud_run_watch_manifest_change_reinstalled":
 				cmd.Println(style.Secondary("App successfully reinstalled"))
+			case "on_cloud_run_watch_app_change":
+				path := event.DataToString("cloud_run_watch_app_change")
```
srtaalej, line 187: praise ⭐

srtaalej (approved):
LGTM! 👏 

[Update](https://github.com/slackapi/slack-cli/pull/310#pullrequestreview-3725119959)
Got it working for manifest.json ⭐ ⭐ ⭐ 
in .slack/config.json change source from remote to local 
```json
{
  "manifest": {
    "source": "local"
  },
  ```
  output:
  ```bash
  Manifest change detected: []/bolt-js-assistant-template/manifest.json, reinstalling app...
Updating local app install for "ale sandbox"
App successfully reinstalled
  ```

File: docs/reference/hooks/index.md [hunk 2]
```diff
@@ -457,7 +467,7 @@ The format for the JSON representing the CLI-SDK interface is as follows:
...
-| config.watch                          | Object with configuration settings for file-watching.                                                                                                                                                                                                                                                                                                                                                           | Optional |
+| config.watch                          | Object with configuration settings for file-watching during `slack run`. Supports updating the `manifest` on change and reloading the `app` server. Read [Watch configurations](#watch-configurations) for details.                                                                                                                                                                                             | Optional |
```
mwbrooks, line 470: thought: Out-of-scope of this PR, but I feel we should start to seriously consider a config.json `version` field. It would allow the CLI to know whether it supports the version of the config file and help developers upgrade when using an older CLI with a newer project.

internal/pkg/platform/localserver.go
```diff
@@ -360,6 +431,94 @@ func (r *LocalServer) Watch(ctx context.Context, auth types.SlackAuth, app types
...
+	// Add provided paths to watcher
+	for _, path := range paths {
+		if err := w.AddRecursive(path); err != nil {
+			r.log.Data["cloud_run_watch_error"] = fmt.Sprintf("app_watcher.paths: %s", err)
+			r.log.Warn("on_cloud_run_watch_error")
+		}
+	}
+
+	// Begin watching for app file changes
+	go func() {
```
mwbrooks, line 483: suggestion: Non-Blocking. There could be issues when multiple files are saved in quick succession. This may cause the Goroutine to restart the app server multiple times in a row.

A solution may be to add a "debouncing delay" where we wait 500ms after the last file change before restarting the server.

I think we should save this for future work and only address it when it's a confirmed issue.

internal/hooks/sdk_config.go
```diff
@@ -76,3 +92,48 @@ type WatchOpts struct {
 func (w *WatchOpts) IsAvailable() bool {
 	return w != nil
 }
+
+// GetManifestWatchConfig returns manifest watch config
+func (w *WatchOpts) GetManifestWatchConfig() (paths []string, filterRegex string, enabled bool) {
```
mwbrooks, line 97: note: Since `GetManifestWatchConfig` and `GetAppWatchConfig` access a `w *WatchOpts` pointer, we may want to check that `w != nil` otherwise the CLI will panic when accessing a nil pointer.

Non-blocker for this PR because we have a lot of unchecked pointers like this.

internal/pkg/platform/localserver.go
```diff
@@ -243,6 +246,39 @@ func (r *LocalServer) Listen(ctx context.Context, errChan chan<- error, done cha
...
+			done := make(chan error)
+			go func() {
+				done <- r.delegateCmd.Wait()
+			}()
+			select {
+			case <-done:
+				r.clients.IO.PrintDebug(ctx, "Previous process stopped successfully")
+			case <-time.After(5 * time.Second):
+				r.clients.IO.PrintDebug(ctx, "Process did not exit in time, force killing")
+				_ = process.Kill()
```
mwbrooks, line 275: note: For completeness, I believe we should wait for the gorountine to complete after the `process.Kill()`. But, I'm happy to keep the code as-is in case we're concerned about indefinite hangs.
```suggestion
				_ = process.Kill()
				<-done  // Wait for goroutine to complete after kill 
```

mwbrooks (approved):
✅ Love Love Love this feature! 💾  ♻️ Great working on implementing this large feature! 👏🏻 

🧪 Manual testing work well on both current projects and projects using the new `@slack/cli-hooks`.

- Server restarts work well even when there is a syntax error that prevents the Bolt server from starting. ❤️  

✏️ Can we tweak the PR Title to be something like "feat: add file watch support to restart Bolt app servers"

❓ Do we have a draft PR for Python?

docs/reference/hooks/index.md
zimeg: @mwbrooks Great callout! I agree so much. We've been careful to make backward compatible changes but adding that field soon would give me more confidence for whatever might happen later.

internal/pkg/platform/localserver.go
zimeg: @mwbrooks Debouncing is such a great idea! For now I agree this might be good to save as follow up to keep this PR implementation more minimal. As feedback arrives let's be sure to take this approach to start 🙏 ✨

internal/hooks/sdk_config.go
zimeg: @mwbrooks What a nice catch! Let's be diligent toward this in changes more 🤖 

I added 5c176af63bbfd0fc9802e3f2f9b5f3ee3d9eff1b to cover this case in hopes that errors are caught in code and expected!

internal/pkg/platform/localserver.go
zimeg: @mwbrooks We might escape the `select` statement after this "kill" instead of blocking. I think exiting after whatever attempt, failed or not might be alright to keep, but if timeouts occur let's explore this more 🔭

PR #312, opened 2026-01-28
chore(claude): add Claude Code support for maintainers

mwbrooks:
### Summary

This pull request adds a basic, shared Claude Code configuration for consistent AI-assisted development across all maintainers. 🤖 

**File structure:**

```
  .claude/                                                                                                                                           
  ├── CLAUDE.md        # Project context, conventions, and coding guidelines
  └── settings.json    # Shared permissions for common commands (go test, make build, etc.)
```

**Extra notes:**

- `.claude/CLAUDE.md` was generated with `/init`
  - I've confirmed (with Claude) that it can live inside of `.claude/` instead of the project root
  - Claude discourages separating the file into `.claude/rules/*.md` because our file is still small and organized
- `.claude/settings.json` auto-approves a few common bash commands
  - I tried to keep it to common, less destructive commands
-  `.gitignore` ignores local Claude Code files

✨ Hoping this is a place that we continue to add more clarity and direction to Claude around our desired coding style and goals!

### Requirements

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -81,7 +81,7 @@ Commands receive the `ClientFactory` and use it to access functionality.
 
 **Tracing**: OpenTracing (Jaeger) is initialized in `main.go` for observability.
 
-**Hook System**: `internal/hooks/` implements a lifecycle hook system that allows SDK projects to inject custom behavior at key points.
+**Hook System**: `internal/hooks/` implements a lifecycle hook system that allows SDK projects to inject custom behavior at key points. The specification exists in `docs/reference/hooks` files.
 
 **Experiment System**: Features can be gated behind experiments defined in `internal/experiment/experiment.go`:
 - Add new experiments as constants


.claude/CLAUDE.md
```diff
@@ -0,0 +1,242 @@
+# CLAUDE.md
```
mwbrooks, line 1: note: Auto-generated with `/init`

.claude/CLAUDE.md
```diff
@@ -0,0 +1,242 @@
...
+### Testing Philosophy
+
+- Unit tests live alongside code with `_test.go` suffix
+- Mock implementations use `_mock.go` suffix
+- Test code uses syntax compatible with the minimum supported Go version (see `go.mod`)
+- The codebase uses `testify` for assertions and `testify/mock` for mocking
+- Mock the `ClientFactory` and its dependencies for testing
+- Always mock file system operations using `afero.Fs` to enable testability
+
+### Table-Driven Test Conventions
```
mwbrooks, line 125: note: I asked Claude to add this note about our Table Test conventions. After doing a little research, it looks like we will prefer **Slice pattern** over **Map pattern**. So, this should encourage Claude to add future tests using the Slice pattern.

The Slice Pattern is used in the Golang documentation, used by Charm _(and we like Charm)_, and guarantees order (which can be a positive or negative, but makes debugging flaky tests easier).

Happy to go with either Pattern. Personally, I like the readability of the Map Pattern, but I'm happy to adopt what is considered best practice.

.claude/settings.json
```diff
@@ -0,0 +1,13 @@
+{
```
mwbrooks, line 1: note: We can always add more here in the future, but feel free to leave a comment if there are any that you strongly feel should be _removed_.

.claude/CLAUDE.md
```diff
@@ -0,0 +1,242 @@
...
+
+### Running Tests in Development
+For targeted test runs during development:
+```bash
+go test -v ./cmd/version -run TestVersionCommand  # Run specific test directly
+go test -ldflags="$(LDFLAGS)" -v ./cmd/auth -race -covermode=atomic -coverprofile=coverage.out
+```
+
+Note: When running tests directly with `go test`, use the LDFLAGS from the Makefile to inject version information.
+
```
zimeg:
```suggestion
```

🪓 suggestion(non-blocking): Might the earlier "testing" section be sufficient? IMO preferring `make` commands might be most simple and something I have to recommend so often for Claude at this time?

.claude/CLAUDE.md
```diff
@@ -0,0 +1,242 @@
...
+- `IO` - I/O streams
+- `Fs` - File system (afero)
+
+Commands receive the `ClientFactory` and use it to access functionality.
+
+**Context-Based State**: The codebase uses `context.Context` extensively to pass runtime state (session IDs, trace IDs, versions) through the call stack via `internal/slackcontext`.
+
+**Tracing**: OpenTracing (Jaeger) is initialized in `main.go` for observability.
+
+**Hook System**: `internal/hooks/` implements a lifecycle hook system that allows SDK projects to inject custom behavior at key points.
```
zimeg:
```suggestion
**Hook System**: `internal/hooks/` implements a lifecycle hook system that allows SDK projects to inject custom behavior at key points. The specification exists in `docs/reference/hooks` files.
```

📚 suggestion: We might point to written docs as well! I found this helpful when making recent changes with documentation-driven development!

.claude/CLAUDE.md
zimeg: @mwbrooks I'm also open to either pattern but think the map pattern has the **advantage** of undetermined orderings at test time. A test that flakes out of order might be worth investigating - IIRC we had some tests that shared authentication states while also making changes within the test case in much earlier commits which caused us to test something unexpected.

.claude/settings.json
zimeg: @mwbrooks I'm not so keen on best practices at the moment but this looks solid!

I might still prefer to encourage `make` testing commands instead of `go` versions but no harm in following up of course!

🔗 https://code.claude.com/docs/en/settings

zimeg (approved):
@mwbrooks LGTM! Such a welcomed change too! I look forward to enhancing available tooling in follow up PRs but this is scoped well to start IMO 👾 ✨ 

I'm finding also that a nested `CLAUDE.md` finds the right context. Clever setups:

```
/context
└ ~/.claude/CLAUDE.md: 101 tokens
└ .claude/CLAUDE.md: 2.3k tokens
```

.claude/CLAUDE.md
mwbrooks: Cool, if you're also preferring the **map pattern** then let's go with it. I agree that I'd rather find flaky tests (or false positives due to variable state carrying across tests) then the opposite.

Commit ea4a490 updates `CLAUDE.md` to prefer map pattern table tests.

.claude/settings.json
mwbrooks: Works for me, I like the idea of having it consistently use our `make` commands as well. We can improve the `CLAUDE.md` over time with these biases. For now, commit a42bca5 removes auto-approving `go test` and `golangci-lint run` commands.

.claude/CLAUDE.md
mwbrooks: Sure, let's axe it. We can always add it back later or change it. All of this was auto-generated to begin with. 🪓

.claude/CLAUDE.md
mwbrooks: Great idea! 💡

PR #317, opened 2026-01-30
fix: cleanup after checks for saved apps when opening "app settings" outside of a project

zimeg:
### Summary

This PR follows #314 to remove the temporary ".slack" directory made when the `app settings` command is run outside of a project directory.

### Reviewers

The example below shows expected behavior after this change:

```
$ cd /tmp
$ slack app settings

🏠 App Settings
   https://api.slack.com/apps

$ ls .slack
".slack": No such file or directory (os error 2)
```

### Requirements

### DIFF
--- a/cmd/app/settings.go
+++ b/cmd/app/settings.go
@@ -101,6 +101,9 @@ func appSettingsCommandRunE(clients *shared.ClientFactory, cmd *cobra.Command, a
 	if err != nil {
 		// If no apps exist, open the list of all apps known to the developer
 		if slackerror.Is(err, slackerror.ErrInstallationRequired) {
+			// Clean up any empty .slack directory and files created during app selection
+			clients.AppClient().CleanUp()
+
 			host := clients.API().Host()
 			parsed, err := url.Parse(host)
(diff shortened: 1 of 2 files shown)


cmd/app/settings.go
```diff
@@ -101,6 +101,9 @@ func appSettingsCommandRunE(clients *shared.ClientFactory, cmd *cobra.Command, a
 	if err != nil {
 		// If no apps exist, open the list of all apps known to the developer
 		if slackerror.Is(err, slackerror.ErrInstallationRequired) {
+			// Clean up any empty .slack directory and files created during app selection
+			clients.AppClient().CleanUp()
```
zimeg, line 105: 📝 note: Another approach might check first if this command was run outside of a project **before** prompting. I'm open to either change, but wanted to be more concise to start.

mwbrooks (approved):
✅ Great find! Eventually, we will want to revisit how we create the `.slack/` directory. A lot has changed in our approach (e.g. `init`) since the first implementation that auto-creates `.slack/`.

PR #316, opened 2026-01-29
refactor(tests): use consistent naming in table tests with the map pattern

mwbrooks:
### Summary

Related to PR #315 

This pull request standardizes the naming convention in table-driven tests that use the **map pattern** across the codebase.

- Standardizes on `name` and `tc` (test case) in all table-driven tests using the slice pattern
- Some common patterns you'll see in this PR:
    - `for name, tt := range` → `for name, tc := range`
    - `for ttName, tt := range` → `for name, tc := range`
    - `for name, test := range` → `for name, tc := range`
    - `for name, s := range` → `for name, tc := range`

### Motivation

This is the second step in refactoring all of our Table Tests to use a consistent style. It creates consistent naming between our Slice and Map Pattern table tests.

The next PR will refactor all simple Slice Pattern tests to Map Pattern.

### Requirements

* [x] I've read and understood the [Contributing Guidelines](/blob/main/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).

PR #320, opened 2026-01-31
refactor(tests): update slice pattern table tests to map pattern (low risk tests)

mwbrooks:
### Summary

Related #315 and #316

This pull request updates our existing Table Tests using the Slice Pattern to the Map Pattern. 

While the Slice Pattern is the chosen format in Golang documentation and many Golang projects, we prefer the Map Pattern for the following reasons:

- Map order is not guaranteed, so the tests are randomized each run. This prevents state from carrying over into the next tests, but can make debugging tests more difficult.
- Test Case names are displayed on a single line improving readability

_There is wrong answer (as far as we know), but we're going with maps for readability and reliability._

### Reviewers

This only updates the simple use-cases in order to keep the PR an easy read. We have 47 remaining Table Tests using the Slice Pattern that will need to be handled later. Over half should be on the low-to-medium risk.

### Requirements

### DIFF
--- a/internal/goutils/slices_test.go
+++ b/internal/goutils/slices_test.go
@@ -57,11 +57,36 @@ func Test_Contains(t *testing.T) {
 		isCaseSensitive bool
 		want            bool
 	}{
-		"not_case_sensitive_success":   {listToCheck: []string{"hi", "hey"}, toFind: "Hey", isCaseSensitive: false, want: true},
-		"case_sensitive_success":       {listToCheck: []string{"hi", "Hey"}, toFind: "Hey", isCaseSensitive: true, want: true},
-		"case_sensitive_fail":          {listToCheck: []string{"hi", "hey", "hello", "apple", "pear"}, toFind: "Hey", isCaseSensitive: true, want: false},
-		"not_case_sensitive_fail":      {listToCheck: []string{"hi", "hey", "hello", "apple", "pear"}, toFind: "Peach", isCaseSensitive: false, want: false},
-		"not_case_sensitive_substring": {listToCheck: []string{"hi", "hey hello"}, toFind: "hey", isCaseSensitive: false, want: false},
+		"not_case_sensitive_success": {
+			listToCheck:     []string{"hi", "hey"},
+			toFind:          "Hey",
+			isCaseSensitive: false,
+			want:            true,
+		},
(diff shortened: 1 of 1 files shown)


internal/goutils/slices_test.go
```diff
@@ -21,55 +21,75 @@ import (
 )
 
 func Test_AppendStringIfNotMember(t *testing.T) {
-	tests := []struct {
-		name          string
+	tests := map[string]struct {
```
zimeg, line 24: ⭐ praise: Another thing I love about structured table tests is the guarantee of unique test names. While rare, it makes a failing test faster to find in past experience!

internal/goutils/slices_test.go
```diff
@@ -21,55 +21,75 @@ import (
...
 		listToCheck     []string
 		toFind          string
 		isCaseSensitive bool
 		want            bool
 	}{
-		{name: "not_case_sensitive_success", listToCheck: []string{"hi", "hey"}, toFind: "Hey", isCaseSensitive: false, want: true},
-		{name: "case_sensitive_success", listToCheck: []string{"hi", "Hey"}, toFind: "Hey", isCaseSensitive: true, want: true},
-		{name: "case_sensitive_fail", listToCheck: []string{"hi", "hey", "hello", "apple", "pear"}, toFind: "Hey", isCaseSensitive: true, want: false},
-		{name: "not_case_sensitive_fail", listToCheck: []string{"hi", "hey", "hello", "apple", "pear"}, toFind: "Peach", isCaseSensitive: false, want: false},
-		{name: "not_case_sensitive_substring", listToCheck: []string{"hi", "hey hello"}, toFind: "hey", isCaseSensitive: false, want: false},
```
zimeg, line 69: ⭐ praise: Thanks for separating fields onto new lines. It's much easier to review these changes!

internal/goutils/slices_test.go
mwbrooks: Agreed and I believe we had a few duplicate names. While rare, it's just another "don't make me think" moment to have the compiler fail early.

internal/goutils/slices_test.go
mwbrooks: No problem and glad you agree on the single line. I wonder if we can configure our linter to enforce this?

PR #326, opened 2026-02-03
feat: add 'create agent' command

srtaalej:
### Changelog
Now when you run the `slack create agent` command a list of Agentic Apps are highlighted for quick selection. 

### Summary
Adds a convenient agent argument to skip category selection and go directly to AI app template selection.

### Usage
```
  # Create an agent app, prompts for language only
  slack create agent

  # Create an agent app with a specific name
  slack create agent my-agent-app

  # Create an app named "agent" with a specific template (not a shortcut)
  slack create agent --template slack-samples/bolt-js-starter-template
```
  Behavior

  | Command                             | Category Prompt | Result                                    |
  |-------------------------------------|-----------------|-------------------------------------------|
  | slack create agent                  | Skipped         | AI app with generated name                |
  | slack create agent my-agent-app     | Skipped         | AI app named "my-agent-app"               |
  | slack create agent --template <url> | Shown           | App named "agent" with specified template |

### Requirements

### DIFF
--- a/cmd/project/create.go
+++ b/cmd/project/create.go
@@ -58,7 +58,7 @@ func NewCreateCommand(clients *shared.ClientFactory) *cobra.Command {
 		Long:       `Create a new Slack project on your local machine from an optional template`,
 		Example: style.ExampleCommandsf([]style.ExampleCommand{
 			{Command: "create my-project", Meaning: "Create a new project from a template"},
-			{Command: "create agent my-agent-app", Meaning: "Create a new AI agent app"},
+			{Command: "create agent my-agent-app", Meaning: "Create a new AI Agent app"},
 			{Command: "create my-project -t slack-samples/deno-hello-world", Meaning: "Start a new project from a specific template"},
 		}),
 		Args: cobra.MaximumNArgs(2),


cmd/project/create.go
```diff
@@ -58,9 +58,10 @@ func NewCreateCommand(clients *shared.ClientFactory) *cobra.Command {
 		Long:       `Create a new Slack project on your local machine from an optional template`,
 		Example: style.ExampleCommandsf([]style.ExampleCommand{
 			{Command: "create my-project", Meaning: "Create a new project from a template"},
+			{Command: "create agent my-agent-app", Meaning: "Create a new AI agent app"},
```
mwbrooks: nit: Perhaps we can capitalize "Agent" as well, since it's becoming the term we use for "AI Agent".
```suggestion
			{Command: "create agent my-agent-app", Meaning: "Create a new AI Agent app"},
```

cmd/project/create.go
mwbrooks, line 56: question: I wonder if we should be updating this usage or perhaps extending the description.

I think the accurate format would be:

```suggestion
		Use:        "create [agent] [name] [flags]",
```

But, I admit that it doesn't look very friendly.

cmd/project/create.go
```diff
@@ -80,14 +81,35 @@ func runCreateCommand(clients *shared.ClientFactory, cmd *cobra.Command, args []
...
+			if !templateFlagProvided {
+				// Shortcut to AI apps category
+				categoryShortcut = "agent"
+				// Check if a second argument was provided as the app name
+				if len(args) > 1 {
+					appNameArg = args[1]
+				}
+			} else {
+				// When --template is provided, "agent" is the app name
+				appNameArg = args[0]
```
mwbrooks, line 104: note: This is a clever way to handle the edge-case. Personally, I start to get confused when `slack create agent` will name my app "agent" vs show me a prompt.

I wonder if a more obvious approach (for the simple minded folks like me) would be anytime `slack create agent` is used, it's focused on the agent prompt.

The drawback to my suggestion is that we'd eventually need to introduce a `--name` flag to allow people to customize their app, such as `slack create --name agent --template <url>`.

I'm okay moving forward with the current approach if it makes sense to @srtaalej and @zimeg. It comes down to personal choice.

cmd/project/create_test.go
```diff
@@ -107,6 +107,94 @@ func TestCreateCommand(t *testing.T) {
...
+			ExpectedAsserts: func(t *testing.T, ctx context.Context, cm *shared.ClientsMock) {
+				template, err := create.ResolveTemplateURL("slack-samples/bolt-js-starter-template")
+				require.NoError(t, err)
+				expected := create.CreateArgs{
+					AppName:  "agent",
+					Template: template,
+				}
+				createClientMock.AssertCalled(t, "Create", mock.Anything, mock.Anything, mock.Anything, expected)
+			},
+		},
```
mwbrooks, line 197: suggestion: Thoughts on adding a test for the use-case `slack create agent some-name --template <url>`. Seems like you've done the hard work so we could bump the test coverage a little more.

cmd/project/create.go
zimeg: @mwbrooks Ahha this inspired similar thought for me! I wonder if we'd be alright to make "agent" a reserved word to prefer:

```
$ slack create [name|agent]
```

And I agree with later comment that a `--name` flag might be optional for this setup 👾

PR #327, opened 2026-02-04
feat(cmd): add --name flag to disambiguate 'agent' shortcut

srtaalej:
### Summary

The `create agent` command is a shortcut to AI Agent templates, but users who want to name their app `agent` had no clear way to do so. Add `--name` flag to explicitly set the app name, bypassing the shortcut behavior.

Follow up to PR https://github.com/slackapi/slack-cli/pull/326

### Behavior

  | Command                 | Result                              |
  |-------------------------|-------------------------------------|
  | create my-app           | App named "my-app"                  |
  | create agent            | AI Agent shortcut                   |
  | create agent my-app     | AI Agent, named "my-app"            |
  | create --name agent     | App named "agent" (NOT an AI Agent) |
  | create agent --name foo | AI Agent, named "foo"               |
### Requirements

### DIFF
--- a/cmd/project/create.go
+++ b/cmd/project/create.go
@@ -1,4 +1,4 @@
-// Copyright 2022-2025 Salesforce, Inc.
+// Copyright 2022-2026 Salesforce, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/cmd/project/create_test.go
+++ b/cmd/project/create_test.go
@@ -1,4 +1,4 @@
-// Copyright 2022-2025 Salesforce, Inc.
+// Copyright 2022-2026 Salesforce, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/.circleci/runner/src/__init__.py
+++ b/.circleci/runner/src/__init__.py
@@ -1,4 +1,4 @@
-# Copyright 2022-2025 Salesforce, Inc.
+# Copyright 2022-2026 Salesforce, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/.circleci/runner/src/circlerunner.py
+++ b/.circleci/runner/src/circlerunner.py
@@ -1,5 +1,5 @@
 #!/usr/bin/python3
-# Copyright 2022-2025 Salesforce, Inc.
+# Copyright 2022-2026 Salesforce, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/.circleci/runner/src/macrunner.py
+++ b/.circleci/runner/src/macrunner.py
@@ -1,5 +1,5 @@
 #!/usr/bin/python3
-# Copyright 2022-2025 Salesforce, Inc.
+# Copyright 2022-2026 Salesforce, Inc.
 #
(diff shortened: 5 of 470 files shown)


cmd/project/create.go
```diff
@@ -53,13 +54,17 @@ const viewMoreSamples = "slack-cli#view-more-samples"
...
-		Long:       `Create a new Slack project on your local machine from an optional template`,
+		Long: `Create a new Slack project on your local machine from an optional template.
+
+The 'agent' argument is a shortcut to create an AI Agent app. If you want to
+name your app 'agent' (not create an AI Agent), use the --name flag instead.`,
 		Example: style.ExampleCommandsf([]style.ExampleCommand{
 			{Command: "create my-project", Meaning: "Create a new project from a template"},
 			{Command: "create agent my-agent-app", Meaning: "Create a new AI Agent app"},
 			{Command: "create my-project -t slack-samples/deno-hello-world", Meaning: "Start a new project from a specific template"},
+			{Command: "create --name agent", Meaning: "Create a project named 'agent' (not an AI Agent)"},
```
mwbrooks: suggestion: I think we can keep this example simpler and avoid the AI Agent topic, since it's addressed in the Long Description. Building on the `my-project` example above it, perhaps we can show how to use `--name` with:
```suggestion
			{Command: "create --name my-project", Meaning: "Create a project named 'my-project'"},
```

cmd/project/create.go
```diff
@@ -71,6 +76,7 @@ func NewCreateCommand(clients *shared.ClientFactory) *cobra.Command {
 	// Add flags
 	cmd.Flags().StringVarP(&createTemplateURLFlag, "template", "t", "", "template URL for your app")
 	cmd.Flags().StringVarP(&createGitBranchFlag, "branch", "b", "", "name of git branch to checkout")
+	cmd.Flags().StringVarP(&createAppNameFlag, "name", "n", "", "name for your app (use this to name your app 'agent')")
```
mwbrooks: suggestion: I think we can remove the agent protip and keep this more general and clarify what happens if you have `create my-app --name my-project`:
```suggestion
	cmd.Flags().StringVarP(&createAppNameFlag, "name", "n", "", "name for your app (overrides the name argument)")
```

cmd/project/create.go
```diff
@@ -108,6 +115,12 @@ func runCreateCommand(clients *shared.ClientFactory, cmd *cobra.Command, args []
 		}
 	}
 
+	// --name flag overrides any positional app name argument
+	// This allows users to name their app "agent" without triggering the AI Agent shortcut
+	if nameFlagProvided {
+		appNameArg = createAppNameFlag
+	}
+
```
mwbrooks, line 123: praise: Love the comment calling out that `--name flag overrides any position app name argument`.

cmd/project/create_test.go
```diff
@@ -195,6 +195,70 @@ func TestCreateCommand(t *testing.T) {
 				createClientMock.AssertCalled(t, "Create", mock.Anything, mock.Anything, mock.Anything, expected)
 			},
 		},
+		"creates an app named agent using name flag without triggering shortcut": {
+			CmdArgs: []string{"--name", "agent"},
```
mwbrooks, line 199: suggestion: Since our intent is for `--name` flag to override the `<name>` argument, perhaps we should have test for the combination to guarantee that future code changes don't create a regression in this behaviour:.

It may also be a good idea to test the agent use-case.

The test cases could be:

1. `slack create my-project --name my-name` → Expects `my-name`
2. `slack create agent my-project --name my-name` → Expects `my-name`

PR #331, opened 2026-02-05
feat(auth): auto-prompt login when no workspaces connected

srtaalej:
### Changelog

> We now prompt a login if no authentications are available when running a command that requires credentials instead of exiting with an error.

### Summary

When running commands like `slack run` without any authenticated workspaces, automatically prompt the user to sign in instead of showing an error.
Screenshot 2026-02-05 at 12 25 47 PM
> Manually test by clearing credentials (`slack logout --all`) and running `slack run`
### Requirements


internal/prompts/app_select.go
```diff
@@ -245,12 +245,21 @@ func getAuths(ctx context.Context, clients *shared.ClientFactory) ([]types.Slack
...
+				WithRemediation("Run %s to sign in to a workspace", style.Commandf("login", false))
+		}
+		clients.IO.PrintInfo(ctx, false, "\n%s", style.Sectionf(style.TextSection{
+			Emoji: "wave",
+			Text:  "No workspaces connected. Sign in to get started.",
+		}))
+		newAuth, _, err := authpkg.LoginWithClients(ctx, clients, "", false)
 		if err != nil {
-			return nil, slackerror.New(slackerror.ErrNotAuthed)
+			return nil, err
```
mwbrooks: suggestion: We should wrap this error rather than returning the raw error, since it may be displayed the user. We wrapped the original error as well, so it's something we'd like to keep doing.
```suggestion
			return nil, slackerror.New(slackerror.ErrNotAuthed).WithCause(err)
```

mwbrooks (approved):
✅ Works beautifully @srtaalej and very elegant implementation! 💅🏻 

🧪 Testing locally works well on both `slack run` and `slack manifest validate`. Also confirmed commands like `slack create` and `slack auth list` do not prompt to login.

✏️ I've left 1 suggestion but that would be nice to add, but non-blocking.

PR #333, opened 2026-02-06
feat(cmd): add 'list' flag to create command

srtaalej:
### Changelog

The templates available to the `create` command can now be shown without requiring an immediate selection with the `--list` flag.

### Summary

This PR adds `--list` flag to `slack create` to print available app templates non-interactively and to `slack create agent` to print only agent templates.

Screenshot 2026-02-06 at 5 01 33 PM

### Requirements


cmd/project/create_template.go
```diff
@@ -240,6 +241,41 @@ func confirmExternalTemplateSelection(cmd *cobra.Command, clients *shared.Client
...
+	for _, category := range categories {
+		templates := getSelectionOptions(clients, category.id)
+		secondary := make([]string, len(templates))
+		for i, tmpl := range templates {
+			secondary[i] = tmpl.Repository
+		}
+		clients.IO.PrintInfo(ctx, false, style.Sectionf(style.TextSection{
+			Text:      style.Bold(category.name),
+			Secondary: secondary,
+		}))
```
zimeg, line 274:
```suggestion
		clients.IO.PrintInfo(ctx, false, style.Sectionf(style.TextSection{
			Emoji:     "house_buildings",
			Text:      style.Bold(category.name),
			Secondary: secondary,
		}))
```

🏘️ suggestion(non-blocking): IIRC the emoji identation and bold can be included with a text section, but I'm unsure if these groupings are needed...

cmd/project/create_template.go
```diff
@@ -240,6 +241,41 @@ func confirmExternalTemplateSelection(cmd *cobra.Command, clients *shared.Client
...
+	} else {
+		categories = []categoryInfo{
+			{id: "slack-cli#getting-started", name: "Getting started"},
+			{id: "slack-cli#ai-apps", name: "AI Agent apps"},
+			{id: "slack-cli#automation-apps", name: "Automation apps"},
+		}
+	}
+
+	for _, category := range categories {
+		templates := getSelectionOptions(clients, category.id)
```
zimeg, line 265: ⭐ praise: This is a nice filter setup!

cmd/project/create_template.go
mwbrooks: suggestion: I prefer no emojis for the `--list` so that it's easier for systems to parse and users to copy & paste. :two-cents:

cmd/project/create_template.go
```diff
@@ -240,6 +241,41 @@ func confirmExternalTemplateSelection(cmd *cobra.Command, clients *shared.Client
...
+	if categoryShortcut == "agent" {
+		categories = []categoryInfo{
+			{id: "slack-cli#ai-apps", name: "AI Agent apps"},
+		}
+	} else {
+		categories = []categoryInfo{
+			{id: "slack-cli#getting-started", name: "Getting started"},
+			{id: "slack-cli#ai-apps", name: "AI Agent apps"},
+			{id: "slack-cli#automation-apps", name: "Automation apps"},
+		}
```
mwbrooks, line 261: suggestion(non-blocking): Not a deal breaker, but I don't like that we're duplicating the categories in 2 areas. In `getSelectionOptionsForCategory(...)` we already list the category titles and now we're duplicating the categories here. We're also slightly changing the titles ("AI Agent app" vs "AI Agent apps" / "Starter app" vs "Getting started"). This will become more error-prone if we add or change categories later on.

We don't need to refactor this if it's complicated, but we should at least add a comment in both sections noting that changes must occur in the other section. And we should try to keep the naming consistent.

PR #340, opened 2026-02-12
feat(create): replace the generated default app name with an optional input prompt

srtaalej:
### Changelog

The `slack create` command now prompts for a custom app name if the randomly generated default isn't quite right for whatever reason. We prompt this before scaffolding the new project from a template!

When making new project directories with a custom app name, we now replace spaces with a dash (instead of removing spaces altogether) to make navigation a bit better across systems. 🦋 

#### Fixes

- **fix: replace spaces with hyphens in app directory names** — `slack create` with an app name containing spaces (e.g. `"my cool app"`) now produces `my-cool-app` instead of `mycoolapp`. Leading and trailing whitespace is also trimmed.
- **fix: generate default app name in non-TTY mode** — Running `slack create -t <template> | cat` (or in scripts) no longer errors with "app name is required". A random name is generated automatically when no name is provided and the terminal is non-interactive.

### Summary

This PR enables users to input an app name if one wasn't provided on `slack create`. Generated name can still be used by pressing `Enter`.

Screenshot 2026-02-12 at 12 49 48 PM

  ## Test plan
  - [x] `make test testdir=cmd/project testname=TestCreateCommand` passes
  - [x] `make test testdir=internal/pkg/create testname=TestGetProjectDirectoryName` passes
  - [x] Manual: `./bin/slack create` shows generated name in gray hint, input field is empty
  - [x] Manual: `./bin/slack create -t slack-samples/bolt-js-starter-template | cat` generates a default name instead of erroring

### Requirements


internal/pkg/create/create.go
```diff
@@ -150,8 +150,8 @@ func Create(ctx context.Context, clients *shared.ClientFactory, log *logger.Logg
 	return appDirPath, nil
 }
 
-// generateRandomAppName will create a random app name based on two words and a number
-func generateRandomAppName() string {
+// GenerateRandomAppName will create a random app name based on two words and a number
+func GenerateRandomAppName() string {
```
zimeg: 🪬 question: Can we move this logic into `cmd/project/create.go` instead of exporting from `pkg`? I'm hoping we can move all of these "inputs" to the command level as we move away from `internal/pkg` long-term!

zimeg: ⭐ ramble: We might prefer erroring below - L165 - instead of attempting to generate this too!

zimeg: 🤖 ramble: Also, so unrelated, but we might want to replace " " spaces with "-" in that `getAppDirName` function... While still removing surrounding spaces of course!

It's strange to combine words otherwise IMHO!

internal/iostreams/survey.go
```diff
@@ -175,7 +175,8 @@ var InputQuestionTemplate = fmt.Sprintf(`
 
 // InputPromptConfig holds additional config for an Input prompt
 type InputPromptConfig struct {
-	Required bool // Whether the input must be non-empty
+	Required bool   // Whether the input must be non-empty
+	Default  string // Default value pre-filled in the prompt
```
zimeg: 👁️‍🗨️ question: Is this used in the defaulting prompt logic above?

No blocker but I'm hoping we can move toward [`huh`](https://github.com/charmbracelet/huh) instead of updating these packages overall!

cmd/project/create.go
```diff
@@ -127,6 +128,21 @@ func runCreateCommand(clients *shared.ClientFactory, cmd *cobra.Command, args []
 		return err
 	}
 
+	// Prompt for app name if not provided via flag or argument
+	if appNameArg == "" {
```
zimeg, line 141:
```suggestion
	if appNameArg == "" && clients.IO.IsTTY() {
```

🤖 issue: We'll want to prefer the generated default if the CLI is run in a "headless" mode and inputs aren't accepted. This "IsTTY" function lets us check if prompting is available, but other changes following might be needed too-

internal/iostreams/survey.go
srtaalej: good catch - default is not used! missed that before committing 😬

cmd/project/create_test.go
```diff
@@ -46,6 +46,7 @@ func TestCreateCommand(t *testing.T) {
 	testutil.TableTestCommand(t, testutil.CommandTests{
 		"creates a bolt application from prompts": {
 			Setup: func(t *testing.T, ctx context.Context, cm *shared.ClientsMock, cf *shared.ClientFactory) {
+				cm.IO.On("IsTTY").Return(true)
```
zimeg, line 49: 🧪 suggestion: We might want to include a test case alongside https://github.com/slackapi/slack-cli/pull/340/commits/af97eca70d91427199c072ffb9e1416a36608db5 changes - do we have a unit test that's checking for `--template` arguments while `IsTTY` is also set to false?

internal/pkg/create/create.go
```diff
@@ -150,23 +148,15 @@ func Create(ctx context.Context, clients *shared.ClientFactory, log *logger.Logg
...
 func getAppDirName(appName string) (string, error) {
 	if len(appName) <= 0 {
-		return generateRandomAppName(), nil
+		return "", fmt.Errorf("app name is required")
 	}
 
 	// trim whitespace
-	appName = strings.ReplaceAll(appName, " ", "")
+	appName = strings.TrimSpace(appName)
+	appName = strings.ReplaceAll(appName, " ", "-")
```
zimeg, line 159: 📣 note: Let's call this out in the changelog as a **fix** as well! I'll make a few changes to this since we'll also want to keep it focused on just developer-facing features-

cmd/project/create.go
```diff
@@ -138,17 +138,21 @@ func runCreateCommand(clients *shared.ClientFactory, cmd *cobra.Command, args []
...
-		if err != nil {
-			return err
-		}
-		if name != "" {
-			appNameArg = name
+	if appNameArg == "" {
+		if clients.IO.IsTTY() {
+			defaultName := generateRandomAppName()
+			cmd.Print(style.Secondary(fmt.Sprintf("  Press Enter to use the generated name: %s", defaultName)), "\n")
+			name, err := clients.IO.InputPrompt(ctx, "Name your app:", iostreams.InputPromptConfig{})
```
zimeg, line 145: 😸 praise: This is a fun prompt!

internal/pkg/create/constants.go
zimeg, line 1: 📠 nit(non-blocking): If these aren't used elsewhere, can we move these constants to `cmd` as well? And perhaps make the slices private again?

PR #344, opened 2026-02-17
feat(create): update app name in 'pyproject.toml' and 'package.json'

srtaalej:
### Changelog

The `create` command now updates the app name in `package.json` and `pyproject.toml` project files.

### Summary
This PR extends UpdateDefaultProjectFiles to also replace the app name in package.json and pyproject.toml when creating a new project

> Before
Screenshot 2026-02-17 at 3 33 52 PM

> After
Screenshot 2026-02-17 at 3 33 32 PM

### Requirements

### DIFF
--- a/internal/app/app.go
+++ b/internal/app/app.go
@@ -162,7 +162,7 @@ func regexReplaceAppNameInManifest(src []byte, appName string) []byte {
 
 // regexReplaceAppNameInPackageJSON replaces the top-level "name" field in a package.json file
 func regexReplaceAppNameInPackageJSON(src []byte, appName string) []byte {
-	re := regexp.MustCompile(`(?m)^(\s*"name"\s*:\s*")([^"]*)(")`)
+	re := regexp.MustCompile(`(?m)^(\s{2}"name"\s*:\s*")([^"]*)(")`)
 	loc := re.FindSubmatchIndex(src)
 	if loc == nil {
 		return src


test/testdata/pyproject-app-name.toml
```diff
@@ -0,0 +1,16 @@
...
+name = "vibrant-butterfly-1234"
+version = "0.1.0"
+requires-python = ">=3.9"
+dependencies = [
+    "slack-sdk==3.40.0",
+    "slack-bolt==1.27.0",
+    "slack-cli-hooks<1.0.0",
+]
+
+[tool.ruff]
```
zimeg, line 12:
```suggestion
# This comment appears before more details
[tool.ruff]
```

👾 suggestion(non-blocking): I understand the advantage of using regex to be comments and indentation kept during replacements. Using a package like [`go-toml`](https://github.com/pelletier/go-toml) would make me more confident in updating structured file, but IIRC templates might include comments that we want to avoid removing on accident!

test/testdata/pyproject.toml
```diff
@@ -0,0 +1,16 @@
...
+name = "bolt-python-ai-agent-template"
+version = "0.1.0"
+requires-python = ">=3.9"
+dependencies = [
+    "slack-sdk==3.40.0",
+    "slack-bolt==1.27.0",
+    "slack-cli-hooks<1.0.0",
+]
+
+[tool.ruff]
```
zimeg, line 12:
```suggestion
# This comment appears before more details
[tool.ruff]
```

🔭 suggestion(non-blocking): This comment follows the one before!

internal/app/app.go
```diff
@@ -149,3 +159,17 @@ func regexReplaceAppNameInManifest(src []byte, appName string) []byte {
 
 	return srcUpdated
 }
+
+// regexReplaceAppNameInPackageJSON replaces the top-level "name" field in a package.json file
+func regexReplaceAppNameInPackageJSON(src []byte, appName string) []byte {
+	re := regexp.MustCompile(`(?m)^(\s*"name"\s*:\s*")([^"]*)(")`)
+	repl := fmt.Sprintf("${1}%s${3}", appName)
+	return re.ReplaceAll(src, []byte(repl))
+}
```
mwbrooks, line 176: suggestion(blocking): I think this regex will match all `"name"` keys in `package.json` instead of just the top-level "name" field. While other `"name"` keys are rare, I think that's a little too risky.

A few ideas:
1. Update the regex to only match the top-level key
    - Not trivial in regex
    - This may work where `$2` is the name field:
    - [`(.*\{[^{]*"name"\s*:\s*")([^"]*)(".*)`](https://rubular.com/r/DaaZvWfEVKvXUe)
2. Just match the name key with 2 spaces
    - The standard indentation used by `npm init`
    - This may work: [`(?m)^(\s{2}"name"\s*:\s*")([^"]*)(")`](https://rubular.com/r/V5bBUrx2wNFJ1c)
    - Personally, this is my preference since it feels "safer"
3. Do nothing when we more than one `"name"` key exists (better to error on doing nothing than corrupting the `package.json` file)

Here is an example of multiple matches and why we'd need a fix:

Image

internal/app/app.go
```diff
@@ -149,3 +159,17 @@ func regexReplaceAppNameInManifest(src []byte, appName string) []byte {
...
+
+// regexReplaceAppNameInPyprojectToml replaces the "name" field under the [project] section in a pyproject.toml file
+func regexReplaceAppNameInPyprojectToml(src []byte, appName string) []byte {
+	re := regexp.MustCompile(`(\[project\][^\[]*?name\s*=\s*")([^"]*)(")`)
+	repl := fmt.Sprintf("${1}%s${3}", appName)
+	return re.ReplaceAll(src, []byte(repl))
+}
```
mwbrooks, line 183: note: I've confirmed that this will only match `name=` under the `[project]` section. Great work @srtaalej!

While we're using teh `re.ReplaceAll` it should only match "all" names under the project and seems okay.

internal/app/app_test.go
```diff
@@ -161,3 +195,43 @@ func Test_RegexReplaceAppNameInManifest(t *testing.T) {
 		})
 	}
 }
+
+func Test_RegexReplaceAppNameInPackageJSON(t *testing.T) {
```
mwbrooks, line 199: suggestion: Can we add a test where `package.json` contains 2 name keys. I think this would build confidence in our regex and help avoid regretful regressions later in life by developers unfamiliar with the regex.

We could put the `"name"` in the config section:

```json
{
  "name": "vibrant-butterfly-1234",
  "version": "1.0.0",
  "description": "A Slack app built with Bolt",
  "main": "app.js",
  "scripts": {
    "start": "node app.js"
  },
  "dependencies": {
    "@slack/bolt": "^4.0.0"
  },
  "config": {
    "name": "local-server-name",
    "host": "localhost",
    "port": "8080"
  }
}
```

internal/app/app_test.go
```diff
@@ -161,3 +195,43 @@ func Test_RegexReplaceAppNameInManifest(t *testing.T) {
...
+	}
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			actualSrc := regexReplaceAppNameInPackageJSON(tc.src, tc.appName)
+			require.Equal(t, tc.expectedSrc, actualSrc)
+		})
+	}
+}
+
+func Test_RegexReplaceAppNameInPyprojectToml(t *testing.T) {
```
mwbrooks, line 274: suggestion: Likewise, I think we should add an extra `name =` under another section to make sure we never replace other name keys.

An example could be:
```
[project]
name = "vibrant-butterfly-1234"
version = "0.1.0"
requires-python = ">=3.9"
dependencies = [
    "slack-sdk==3.40.0",
    "slack-bolt==1.27.0",
    "slack-cli-hooks<1.0.0",
]

[tool.ruff]
[tool.ruff.lint]
[tool.ruff.format]

[tool.pytest.ini_options]
testpaths = ["tests"]

[project.scripts]
name = "my_package.name:main_function"
```

test/testdata/pyproject-app-name.toml
mwbrooks: Unfortunately, I don't think we should do that. I've experimented with `go-toml` for our `pyproject.toml` work. When you read-in and write-out, it will re-order the sections and keys based on a random order of the Golang map. So, I've taken a regex approach with the venv as well.

test/testdata/testdata.go
```diff
@@ -27,3 +27,15 @@ var ManifestSDKTS []byte
...
+var PackageJSON []byte
+
+//go:embed package-app-name.json
+var PackageJSONAppName []byte
+
+//go:embed pyproject.toml
+var PyprojectTOML []byte
+
+//go:embed pyproject-app-name.toml
+var PyprojectTOMLAppName []byte
```
mwbrooks, line 41: praise: Great use of the embeds to keep test data clean and easy to maintain!

internal/app/app.go
srtaalej: wow thank you for the detailed explanantion! this is super helpful ❤️ looking into this ASAP 🚀

PR #345, opened 2026-02-19
feat(create): add '--subdir <path>' flag to 'create' command

srtaalej:
### Changelog

Added `--subdir` flag to slack create for extracting a subdirectory from a template repository as the project root. This supports monorepo-style templates where multiple apps live in subdirectories (e.g. `slack create my-app -t org/monorepo --subdir apps/my-app`). 🦋 

### Summary

This PR adds a `--subdir <path>` flag to slack create that works alongside `--template` and `--branch`. When provided, the full template is cloned into a temporary directory, then only the specified subdirectory is copied to the final project path.

  Behavior:
  - `--subdir pydantic-ai/` → extracts the pydantic-ai directory from the template
  - `--subdir . or --subdir /` → equivalent to omitting the flag (full repo)
  - `--subdir ../escape` → rejected with an error (path traversal)
  - `--subdir nonexistent` → error with remediation hint pointing to the template
  - `--list --subdir foo` → --list returns early, --subdir is ignored

### Requirements

### DIFF
--- a/internal/shared/clients.go
+++ b/internal/shared/clients.go
@@ -222,7 +222,12 @@ func (c *ClientFactory) InitSDKConfig(ctx context.Context, dirPath string) error
 			return slackerror.New(slackerror.ErrHooksJSONLocation)
 		}
 		// Move upward one directory level
-		dirPath = filepath.Dir(dirPath)
+		parentDir := filepath.Dir(dirPath)
+		if parentDir == dirPath {
+			// Reached a filesystem root not covered above (e.g. D:\ when SYSTEMROOT is on C:\)
+			return slackerror.New(slackerror.ErrHooksJSONLocation)
+		}
+		dirPath = parentDir
 	}
 	configFileBytes, err := afero.ReadFile(c.Fs, hooksJSONFilePath)
 	if err != nil {
--- a/internal/shared/clients_test.go
+++ b/internal/shared/clients_test.go
@@ -110,6 +110,12 @@ func Test_ClientFactory_InitSDKConfig(t *testing.T) {
 			mockWorkingDirectory:  filepath.Join("path", "outside", "home", "to", "project"),
 			expectedError:         slackerror.New(slackerror.ErrHooksJSONLocation),
 		},
+		"errors if traversal reaches a filesystem root": {
+			mockHooksJSONContent:  "{}",
+			mockHooksJSONFilePath: filepath.Join(string(filepath.Separator), "other", "volume", "project", "package.json"),
+			mockWorkingDirectory:  filepath.Join(string(filepath.Separator), "other", "volume", "project"),
+			expectedError:         slackerror.New(slackerror.ErrHooksJSONLocation),
+		},
 	}
 	for name, tc := range tests {
 		t.Run(name, func(t *testing.T) {


cmd/project/create.go
```diff
@@ -79,6 +81,7 @@ name your app 'agent' (not create an AI Agent), use the --name flag instead.`,
 	cmd.Flags().StringVarP(&createGitBranchFlag, "branch", "b", "", "name of git branch to checkout")
 	cmd.Flags().StringVarP(&createAppNameFlag, "name", "n", "", "name for your app (overrides the name argument)")
 	cmd.Flags().BoolVar(&createListFlag, "list", false, "list available app templates")
+	cmd.Flags().StringVar(&createSubdirFlag, "subdir", "", "subdirectory within the template to use as project root")
```
mwbrooks: suggestion: An attempt to shorten the description so that it doesn't wrap on 80 character terminals

```suggestion
	cmd.Flags().StringVar(&createSubdirFlag, "subdir", "", "subdirectory in the template to use as project")
```

**Before:**
Image

**After:**
Image

internal/pkg/create/create.go
```diff
@@ -343,6 +355,60 @@ func createApp(ctx context.Context, dirPath string, template Template, gitBranch
 	return nil
 }
 
+// normalizeSubdir cleans the subdir path and returns "" if it resolves to root.
+func normalizeSubdir(subdir string) (string, error) {
```
mwbrooks, line 355: praise: very nice! 👌🏻

internal/pkg/create/create.go
```diff
@@ -343,6 +355,60 @@ func createApp(ctx context.Context, dirPath string, template Template, gitBranch
...
+	if subdir == "" {
+		return "", nil
+	}
+	cleaned := filepath.Clean(subdir)
+	if cleaned == "." || cleaned == "/" {
+		return "", nil
+	}
```
mwbrooks, line 362: suggestion: After `filepath.Clean()` maybe we should use [`filepath.isLocal(cleaned)`](https://pkg.go.dev/path/filepath#IsLocal). 

This function appears to check if the file path is within the subtree and prevent traversal attacks where the path tries to access files outside of the root directory (template directory).

I haven't used it personally, but it seems like a good security measure.

Image

internal/pkg/create/create.go
```diff
@@ -121,8 +122,19 @@ func Create(ctx context.Context, clients *shared.ClientFactory, log *logger.Logg
...
+
+	if subdir != "" {
+		if err := createAppFromSubdir(ctx, projectDirPath, createArgs.Template, createArgs.GitBranch, subdir, log, clients.Fs); err != nil {
+			return "", slackerror.Wrap(err, slackerror.ErrAppCreate)
+		}
+	} else {
+		if err := createApp(ctx, projectDirPath, createArgs.Template, createArgs.GitBranch, log, clients.Fs); err != nil {
+			return "", slackerror.Wrap(err, slackerror.ErrAppCreate)
+		}
```
mwbrooks, line 141: praise: clean implementation choice!

internal/pkg/create/create.go
```diff
@@ -343,6 +355,60 @@ func createApp(ctx context.Context, dirPath string, template Template, gitBranch
...
+		return "", slackerror.New(slackerror.ErrSubdirNotFound).
+			WithMessage("subdirectory path %q must be relative and within the template", subdir)
+	}
+	return cleaned, nil
+}
+
+// createAppFromSubdir clones the full template into a temp directory, then copies
+// only the specified subdirectory to the final project path.
+func createAppFromSubdir(ctx context.Context, dirPath string, template Template, gitBranch string, subdir string, log *logger.Logger, fs afero.Fs) error {
+	tmpDir, err := os.MkdirTemp("", "slack-create-*")
```
mwbrooks: suggestion: We should not use `os.` calls directly whenever possible. These calls are very hard to test because we can't mock it - it will create a temp directory during the unit test.

Instead, we should use our `fs afero.Fs` filesystem library. This uses `os` in production and a memory-based file system in tests.

I imagine we can use [`afero.GetTempDir()`](https://pkg.go.dev/github.com/spf13/afero#GetTempDir) and [`afero.TempDir()`](https://pkg.go.dev/github.com/spf13/afero#TempDir) to look something like:

```golang
tempDirRoot := afero.GetTempDir(fs, "")
tempDirPath, err := afero.TempDir(fs, tmpDirRoot, "slack-create-")
```

internal/pkg/create/create.go
```diff
@@ -343,6 +355,60 @@ func createApp(ctx context.Context, dirPath string, template Template, gitBranch
...
+	if err != nil {
+		return slackerror.Wrap(err, "failed to create temporary directory")
+	}
+	// Remove so createApp can create it fresh (go-git requires non-existent target)
+	os.Remove(tmpDir)
+	defer os.RemoveAll(tmpDir)
```
mwbrooks: suggestion: Likewise, here I think we should use `fs` so that we can write unit tests that do not modify the actual file system.

* [`fs.Remove(tmpDir)`](https://pkg.go.dev/github.com/spf13/afero#BasePathFs.Remove)
* [`defer fs.RemoveAll(tmpDir)`](https://pkg.go.dev/github.com/spf13/afero#BasePathFs.RemoveAll)

internal/pkg/create/create.go
```diff
@@ -343,6 +355,60 @@ func createApp(ctx context.Context, dirPath string, template Template, gitBranch
...
+
+	if err := createApp(ctx, tmpDir, template, gitBranch, log, fs); err != nil {
+		return err
+	}
+
+	subdirPath := filepath.Join(tmpDir, subdir)
+	info, err := os.Stat(subdirPath)
```
mwbrooks: suggestion: If we're using the `fs` then it's important that we'll need to use [`fs.Stat(subdirPath)`](https://pkg.go.dev/github.com/spf13/afero#BasePathFs.Stat) so that it reads the real file system in production and memory-based file system during tests.

internal/pkg/create/create.go
```diff
@@ -343,6 +355,60 @@ func createApp(ctx context.Context, dirPath string, template Template, gitBranch
...
+	if err != nil {
+		if os.IsNotExist(err) {
```
mwbrooks, line 388: suggestion: I think this is okay, but if you want to mock the result during tests (to test both exist and not exist stats) then you can use the `clients.Os.IsNotExist(err)` method. Usually we do this by passing `clients.Os` in function similar to `fs` (which is `clients.Fs`).

internal/pkg/create/create.go
```diff
@@ -343,6 +355,60 @@ func createApp(ctx context.Context, dirPath string, template Template, gitBranch
...
+	if err != nil {
+		return slackerror.Wrap(err, "failed to create temporary directory")
+	}
+	// Remove so createApp can create it fresh (go-git requires non-existent target)
+	os.Remove(tmpDir)
+	defer os.RemoveAll(tmpDir)
+
+	if err := createApp(ctx, tmpDir, template, gitBranch, log, fs); err != nil {
+		return err
+	}
```
mwbrooks: praise: Clever and clean approach!

internal/pkg/create/create.go
```diff
@@ -343,6 +355,60 @@ func createApp(ctx context.Context, dirPath string, template Template, gitBranch
...
+
+	subdirPath := filepath.Join(tmpDir, subdir)
+	info, err := os.Stat(subdirPath)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return slackerror.New(slackerror.ErrSubdirNotFound).
+				WithMessage("subdirectory %q was not found in the template", subdir).
+				WithRemediation("Check that the path exists in the template at %q", template.GetTemplatePath())
```
mwbrooks, line 391: TIL: `%q` vs `%s`

internal/pkg/create/create.go
```diff
@@ -343,6 +355,60 @@ func createApp(ctx context.Context, dirPath string, template Template, gitBranch
...
+		return slackerror.New(slackerror.ErrSubdirNotFound).
+			WithMessage("path %q in the template is not a directory", subdir)
+	}
+
+	return goutils.CopyDirectory(goutils.CopyDirectoryOpts{
+		Src:               subdirPath,
+		Dst:               dirPath,
+		IgnoreDirectories: []string{".git", ".venv", "node_modules"},
+		IgnoreFiles:       []string{".DS_Store"},
+	})
```
mwbrooks, line 405: suggestion: I noticed that we have the same `IgnoreDirectories` and `IgnoreFiles` on L337. Thoughts on pulling those two slices out to use in both places (less error prone if we need to ignore more files/directories). Alternatively, this could become a function used in both places.

cmd/project/create.go
```diff
@@ -141,6 +144,7 @@ func runCreateCommand(clients *shared.ClientFactory, cmd *cobra.Command, args []
 		AppName:   appNameArg,
 		Template:  template,
 		GitBranch: createGitBranchFlag,
+		Subdir:    createSubdirFlag,
```
mwbrooks, line 176: question: Right now `--subdir` without `--template` is silently accepted. For example, `slack create --subdir src` will use the `src/` of whatever template is selected in the prompts.

Personally, I'd learn toward erroring if `--subdir` is used without `--template` because it avoids unexpected behaviour or exploits and we can "loosen" that contraint later if we decided we want that experience.

internal/pkg/create/create.go
```diff
@@ -343,6 +355,60 @@ func createApp(ctx context.Context, dirPath string, template Template, gitBranch
...
+
+// createAppFromSubdir clones the full template into a temp directory, then copies
+// only the specified subdirectory to the final project path.
+func createAppFromSubdir(ctx context.Context, dirPath string, template Template, gitBranch string, subdir string, log *logger.Logger, fs afero.Fs) error {
+	tmpDir, err := os.MkdirTemp("", "slack-create-*")
+	if err != nil {
+		return slackerror.Wrap(err, "failed to create temporary directory")
+	}
+	// Remove so createApp can create it fresh (go-git requires non-existent target)
+	os.Remove(tmpDir)
```
mwbrooks: question: Why do we make the temp directory on L377 and then immediately delete it?

cmd/project/create.go
srtaalej: ty for the example 🙌 🫶

internal/pkg/create/create.go
srtaalej: ty for sharing - i didnt know we were using a package for temp directories 👍 this is a game changer 💯

internal/pkg/create/create.go
srtaalej: `MkdirTemp` generates unique path names which we want for `createApp`. But we don't want the directory itself which is why its deleted right away. Its a workaround to extract the unique name creation part of the `MkdirTemp` function so we can use it in `createApp`.

cmd/project/create.go
zimeg: 👁️‍🗨️ Agreed! This might've been added with https://github.com/slackapi/slack-cli/pull/345/commits/b1312a7a0252f7f8d35b1f486f839fe4aa7f71fd - I'll mark this as resolved

cmd/project/create.go
```diff
@@ -128,6 +132,12 @@ func runCreateCommand(clients *shared.ClientFactory, cmd *cobra.Command, args []
 		return listTemplates(ctx, clients, categoryShortcut)
 	}
 
+	// --subdir requires --template
+	if cmd.Flags().Changed("subdir") && !templateFlagProvided {
+		return slackerror.New(slackerror.ErrMismatchedFlags).
+			WithMessage("the --subdir flag requires the --template flag")
```
zimeg:
```suggestion
			WithMessage("The --subdir flag requires the --template flag")
```

🏁 polish: To match surrounding error messages. Though I'm not confident in how consistent we are with this... Let's take note to include this in a **Style Guide** soon!

internal/slackerror/errors.go
```diff
@@ -225,6 +225,7 @@ const (
 	ErrSocketConnection                              = "socket_connection_error"
 	ErrScopesExceedAppConfig                         = "scopes_exceed_app_config"
 	ErrStreamingActivityLogs                         = "streaming_activity_logs_error"
+	ErrSubdirNotFound                                = "subdir_not_found"
```
zimeg, line 228: 💡 thought(non-blocking): Would it make sense to group this alongside other template errors? To me the referenced `subdir` wouldn't be clear at a glance. I'd suggest:

```
template_subdir_not_found
```

Which is so close to `template_path_not_found`!

internal/pkg/create/create.go
zimeg: 👁️‍🗨️ Agreed! Marking as resolved from an earlier commit.

internal/pkg/create/create.go
zimeg: 🪬 question(non-blocking): Would passing `subdir` to `createApp` in all instances - either a default "." or the normalized path - reduce this implementation more? I'm concerned that a change to the temporary directory logic might require repeated change if we keep separate setups for this.

internal/pkg/create/create.go
```diff
@@ -343,6 +361,60 @@ func createApp(ctx context.Context, dirPath string, template Template, gitBranch
...
+	if subdir == "" {
+		return "", nil
+	}
+	cleaned := filepath.Clean(subdir)
+	if cleaned == "." || cleaned == "/" {
+		return "", nil
+	}
+	if !filepath.IsLocal(cleaned) {
+		return "", slackerror.New(slackerror.ErrSubdirNotFound).
+			WithMessage("subdirectory path %q must be relative and within the template", subdir)
```
zimeg:
```suggestion
			WithMessage("Subdirectory path %q must be relative and within the template", subdir)
```

🧮 quibble: Casing to errors as a complete sentence!

internal/pkg/create/create.go
```diff
@@ -343,6 +361,60 @@ func createApp(ctx context.Context, dirPath string, template Template, gitBranch
...
+	if err := createApp(ctx, cloneDir, template, gitBranch, log, fs); err != nil {
+		return err
+	}
+
+	subdirPath := filepath.Join(cloneDir, subdir)
+	info, err := fs.Stat(subdirPath)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return slackerror.New(slackerror.ErrSubdirNotFound).
+				WithMessage("subdirectory %q was not found in the template", subdir).
```
zimeg:
```suggestion
				WithMessage("Subdirectory %q was not found in the template", subdir).
```

🚢 quibble: Casings to the sentence structure!

internal/pkg/create/create.go
```diff
@@ -343,6 +361,60 @@ func createApp(ctx context.Context, dirPath string, template Template, gitBranch
...
+				WithRemediation("Check that the path exists in the template at %q", template.GetTemplatePath())
+		}
+		return slackerror.Wrap(err, "failed to access subdirectory")
+	}
+	if !info.IsDir() {
+		return slackerror.New(slackerror.ErrSubdirNotFound).
+			WithMessage("path %q in the template is not a directory", subdir)
```
zimeg:
```suggestion
			WithMessage("Path %q in the template is not a directory", subdir)
```

📠 quibble: Forbid this error be reached, but let's start with a capital path here.

PR #346, opened 2026-02-20
feat(python): add support for creating python .venv and installing dependencies

mwbrooks:
### Changelog

> Updated the `slack create` and `slack init` commands to support creating a Python virtual environment (`.venv`) when it doesn't exist and installing the project's dependencies from `requirements.txt` and `pyproject.toml`.

### Summary

This pull request adds support for creating a Python virtual environment when it doesn't already exist and installs the dependencies defined in `requirements.txt` and `pyproject.toml`.

Out-of-scope:
- Activating the virtual environment on each Slack CLI commands. We can follow-up in another PR with that feature.

Known issues:
- When a `pyproject.toml` doesn't include a `[project]` section or `dependencies` array, the Slack CLI will display an error. It gracefully fails by continuing with the installation process for requirements.txt. However, I think we have an opportunity to update the Slack CLI to display a "Skipped updating pyproject.toml: no dependencies array" message instead of an error. That can be handled in a follow-up PR.

Related PRs:
- https://github.com/slack-samples/bolt-python-starter-template/pull/71
- https://github.com/slack-samples/bolt-python-custom-step-template/pull/69

### Preview

Example of an app with a valid **`pyproject.toml`:**

image

**Example of an app with an incomplete `pyproject.toml`:**

image

### Reviewers

Example of an app with a valid `pyproject.toml`:

```bash
$ lack create my-test-app/
# → Select "AI Agent App"
# → Select "Bolt for Python"

# Validate output:
# → Confirm: "Created virtual environment at .venv"
# → Confirm: "Installed dependencies from pyproject.toml"
# → Confirm: "Installed dependencies from requirements.txt"
# → Confirm: No errors

# Validate .env
$ ls -al my-test-app/.venv
# → Confirm: Exists

# Test app dependencies installed
$ cd my-test-app/
$ source .venv/bin/activate
$ lack run
# → Confirm: App starts

# Clean up
$ cd ../
$ rm -rf my-test-app/
```

Example of an app with an incomplete `pyproject.toml`:

```bash
$ lack create my-test-app/
# → Select "Starter App"
# → Select "Bolt for Python"

# Validate output:
# → Confirm: "Created virtual environment at .venv"
# → Confirm: "Error updating pyproject.toml: pyproject.toml missing project section"
# → Confirm: "Installed dependencies from pyproject.toml"
# → Confirm: "Installed dependencies from requirements.txt"

# Validate .env
$ ls -al my-test-app/.venv
# → Confirm: Exists

# Test app dependencies installed
$ cd my-test-app/
$ source .venv/bin/activate
$ lack run
# → Confirm: App starts

# Clean up
$ cd ../
$ rm -rf my-test-app/
```

### Requirements


internal/runtime/python/python.go
```diff
@@ -128,18 +190,18 @@ func installPyProjectToml(fs afero.Fs, projectDirPath string) (output string, er
 	projectSection, exists := config["project"]
 	if !exists {
 		err := fmt.Errorf("pyproject.toml missing project section")
-		return fmt.Sprintf("Error: %s", err), err
+		return fmt.Sprintf("Error updating pyproject.toml: %s", err), err
```
mwbrooks, line 206: note: Improving error messages when there is an error adding `slack-cli-hooks` to `pyproject.toml`.

mwbrooks (commented):
Comments for the kind folks 👓

internal/runtime/python/python.go
```diff
@@ -62,6 +63,80 @@ func (p *Python) IgnoreDirectories() []string {
...
+		Name:    "CreateVirtualEnvironment",
+		Command: fmt.Sprintf("%s -m venv .venv", getPythonExecutable()),
+	}
+	stdout := bytes.Buffer{}
+	hookExecOpts := hooks.HookExecOpts{
+		Hook:      hookScript,
+		Stdout:    &stdout,
+		Directory: projectDirPath,
+	}
+	_, err := hookExecutor.Execute(ctx, hookExecOpts)
```
mwbrooks, line 108: note: While it's verbose, we use the `hookExecutor` so that we can mock the exec calls during tests. This is how we also handle `npm install`.

cmd/project/init.go
```diff
@@ -52,7 +52,7 @@ func NewInitCommand(clients *shared.ClientFactory) *cobra.Command {
 			"Installs your project dependencies when supported:",
 			"- Deno:    Supported",
 			"- Node.js: Supported",
-			"- Python:  Unsupported",
+			"- Python:  Supported",
```
zimeg, line 55: 🐍 praise: Such a nice `diff` to find!

internal/runtime/python/python.go
```diff
@@ -210,44 +284,26 @@ func (p *Python) InstallProjectDependencies(ctx context.Context, projectDirPath
...
-		if hasPyProjectToml {
-			outputs = append(outputs, fmt.Sprintf("  Install project dependencies: %s", style.CommandText("pip install -e .")))
-		}
+	// Ensure at least one dependency file exists
+	if !hasRequirementsTxt && !hasPyProjectToml {
+		err := fmt.Errorf("no Python dependency file found (requirements.txt or pyproject.toml)")
+		return fmt.Sprintf("Error: %s", err), err
+	}
 
-		outputs = append(outputs, fmt.Sprintf("  Learn more: %s", style.Underline("https://docs.python.org/3/tutorial/venv.html")))
```
zimeg, line 239: 🪓 praise: Farewell setup steps! I notice #347 brings changes to make the activation step automatic too! Really neat!

internal/runtime/python/python.go
zimeg: @mwbrooks This is a solid approach! IIRC the `default` hook execution protocol is used here since we don't have hooks loaded at this point, and the testing approach is great!

I'm hesitant on the `CreateVirtualEnvironment` hook name right now - it's almost an extension of a shared `InstallDependencies` to me. Rambles but this isn't exposed in configurations so for iteration ongoing 📚

internal/runtime/python/python.go
```diff
@@ -210,44 +284,26 @@ func (p *Python) InstallProjectDependencies(ctx context.Context, projectDirPath
...
+	// Get virtual environment path
+	venvPath := getVenvPath(projectDirPath)
 
-		// Get first error or nil
-		var firstErr error
-		if len(errs) > 0 {
-			firstErr = errs[0]
+	// Create virtual environment if it doesn't exist
+	if !venvExists(fs, venvPath) {
+		ios.PrintDebug(ctx, "Creating Python virtual environment")
```
zimeg, line 298: 🌟 praise: Debugging the **start** of something and printing the **end** is a super fascinating pattern!

internal/runtime/python/python_test.go
```diff
@@ -63,6 +64,238 @@ func Test_Python_IgnoreDirectories(t *testing.T) {
...
+name = "my-app"
+dependencies = ["slack-bolt>=1.0.0", "pytest==8.3.2"]`,
+			shouldContain:  []string{"slack-bolt", "slack-cli-hooks", "pytest"},
+			expectedOutput: "Updated pyproject.toml with",
+			expectedError:  false,
+		},
+		"Add at end when slack-bolt doesn't exist": {
+			existingContent: `[project]
+name = "my-app"
+dependencies = ["pytest==8.3.2"]`,
```
zimeg, line 224: 🧪 suggestion: We might want to follow up to support something as following, but I also understand that the initial update is most meaningful since this file can be modified to preference afterward:

```toml
[project.optional-dependencies]
cli = [
  "slack-cli-hooks<1.0.0",
]
```

🔗 https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#dependencies-and-requirements

internal/runtime/python/python.go
```diff
@@ -62,6 +63,80 @@ func (p *Python) IgnoreDirectories() []string {
...
+	}
+	_, err := hookExecutor.Execute(ctx, hookExecOpts)
+	output := stdout.String()
+	if err != nil {
+		return output, fmt.Errorf("pip install failed: %w", err)
+	}
+	return output, nil
+}
+
 // installRequirementsTxt handles adding slack-cli-hooks to requirements.txt
```
zimeg, line 140: 🐣 note: I found the test cases confusing for these "install" functions that are also updating the requirements. Not a blocker for this PR, but we might change this to be something as:

```
initializeAndInstallRequirementsTxt
```

If changes are ongoing? Apologies for calling this out in an unrelated PR too!

internal/runtime/python/python_test.go
zimeg: 👁️‍🗨️ note: I do fear that'd complicate our current `InstallProjectDependencies` hook to require for certain cases:

```
$ pip install -e ".[cli]"
```

zimeg (approved):
> When a `pyproject.toml` doesn't include a `[project]` section or dependencies array, the Slack CLI will display an error. It gracefully fails by continuing with the installation process for `requirements.txt`.

@mwbrooks Thanks for calling this out! I think the current approach is solid with your updates to sample apps that add "pyproject.toml" files - this seems sometimes recommended alongside "requirements.txt" for pinned dependencies but I'm not sure how frequent that is? I suggest we remove "requirements.txt" but we should decide later!

I'm approving now after testing well on the `create` experience and leaving a few notes! FWIW this adds multiple seconds to this step - 5 seconds on a solid work machine and 55 seconds on the Windows laptop I keep. The `.venv` path seems standard of "uv" and "pip" but that time might feel extra if "uv" [benchmarks](https://github.com/astral-sh/uv/blob/main/BENCHMARKS.md) are true! 🚢 💨 

This does address the extra step I often forget when switching between `javascsript` and `python` so please know these notes are rambles. Let's get this merged for next release 🎁 🐍 ✨

internal/runtime/python/python.go
mwbrooks: @zimeg Your hesitation is understandable. We already use this identical approach in other areas of the code. I think one small improvement that we could make is to rename the `Name:` to be something standard like `Name: "CustomExec"` or something.

PR #347, opened 2026-02-21
feat(python): auto-activate venv before hook execution

mwbrooks:
### Changelog

> When running Slack CLI commands in a Bolt Python project, the CLI will attempt to automatically activate the venv. This will allow commands that use the `slack-cli-hooks` Python package to safely and successfully run even when the terminal system doesn't have the virtual environment activated.

### Summary

This pull request gives Python the **"create and run"** experience that Node and Deno enjoy. 

It enables auto-activation of venv, when it exists, so that commands can run safely and successfully even when the terminal session doesn't have a virtual environment activated. When the `.venv` is missing, the CLI will continue like normal.

Test coverage:
- I noticed that the test coverage is reported as being low
- Checking locally, it should be higher than reported
- I'm hopeful that once `mwbrooks-python-venv` is merged, this branch will report higher coverage

Merge order:
- This PR depends on the branch `mwbrooks-python-venv`
- First merge `mwbrooks-python-venv` then merge this PR into `main`

### Preview

https://github.com/user-attachments/assets/c8841592-5299-464b-95b7-fb2bc3999f24

### Reviewers

**Bolt for Python:**
```
$ lack create my-test-app/
# → Select "AI Agent App"
# → Select "Bolt for Python"

# Validate output:
# → Confirm: "Created virtual environment at .venv"
# → Confirm: "Installed dependencies from pyproject.toml"
# → Confirm: "Installed dependencies from requirements.txt"
# → Confirm: No errors

# Run the app
$ cd my-test-app/
$ lack run
# → Confirm: App starts

# Clean up
$ cd ../
$ rm -rf my-test-app/
```

**Bolt for JavaScript:**
```
$ lack create my-test-app/
# → Select "AI Agent App"
# → Select "Bolt for JavaScript" <---- Not Python

# Validate output:
# → Confirm: No errors

# Run the app
$ cd my-test-app/
$ lack run
# → Confirm: App starts

# Clean up
$ cd ../
$ rm -rf my-test-app/
```

### Requirements


internal/runtime/python/python.go
```diff
@@ -76,6 +77,43 @@ func getPythonExecutable() string {
...
+
+	binDir := getVenvBinDir(venvPath)
+
+	if err := os.Setenv("VIRTUAL_ENV", venvPath); err != nil {
+		return false, err
+	}
+	if err := os.Setenv("PATH", binDir+string(filepath.ListSeparator)+os.Getenv("PATH")); err != nil {
+		return false, err
+	}
+	os.Unsetenv("PYTHONHOME")
```
mwbrooks: note: I imagine we should create wrappers for these calls in our `slackdeps.Os` 🤔

internal/runtime/runtime.go
```diff
@@ -66,6 +66,14 @@ func New(runtimeName string) (Runtime, error) {
...
 }
 
+// ActivatePythonVenvIfPresent activates a Python virtual environment if one
+// exists in the given project directory. This sets VIRTUAL_ENV, prepends the
+// venv bin directory to PATH, and unsets PYTHONHOME on the current process so
+// that child processes (hook scripts) inherit the activated venv.
+func ActivatePythonVenvIfPresent(fs afero.Fs, projectDir string) (bool, error) {
+	return python.ActivateVenvIfPresent(fs, projectDir)
+}
+
```
mwbrooks, line 76: note: While it feels wrong to have Python in the generic `runtime` this allows us to activate it for any project. We could move this into the Python runtime and think of a more general function that could be used across all runtimes.

Upside of this approach is that JS projects that include a `.venv` (perhaps for scripts) can be activated automatically.

internal/shared/clients.go
```diff
@@ -224,6 +224,14 @@ func (c *ClientFactory) InitSDKConfig(ctx context.Context, dirPath string) error
 		// Move upward one directory level
 		dirPath = filepath.Dir(dirPath)
 	}
+	// Activate Python virtual environment if present, so hook scripts
+	// can resolve the venv's Python and installed packages.
+	if activated, err := runtime.ActivatePythonVenvIfPresent(c.Fs, dirPath); err != nil {
+		c.IO.PrintDebug(ctx, "failed to activate Python virtual environment: %s", err)
+	} else if activated {
+		c.IO.PrintDebug(ctx, "Activated Python virtual environment .venv")
```
mwbrooks, line 237: note: I'd love to display this in the regular output, but it's tough to get a nice format because it can be displayed in many different situations.

mwbrooks (commented):
Comments for those kind reviewers! 🙇🏻

internal/runtime/python/python.go
mwbrooks: note: I updated this PR to use `slackdeps.Os` and added `slackdeps.Os.lookupEnv`.

zimeg: @mwbrooks 🌲 So good! I think we'll want to reuse this in other places soon?

internal/runtime/python/python.go
```diff
@@ -76,6 +76,45 @@ func getPythonExecutable() string {
 	return "python3"
 }
 
+// getVenvBinDir returns the platform-specific bin directory inside a virtual environment
+func getVenvBinDir(venvPath string) string {
+	if runtime.GOOS == "windows" {
+		return filepath.Join(venvPath, "Scripts")
+	}
+	return filepath.Join(venvPath, "bin")
+}
```
zimeg, line 85: 👁️‍🗨️ issue(non-blocking): I'm finding the `windows` option is erroring with perhaps unrelated hook errors in execution explored in https://github.com/slackapi/python-slack-hooks/pull/96

internal/runtime/python/python.go
```diff
@@ -76,6 +76,45 @@ func getPythonExecutable() string {
...
+// The three env vars (VIRTUAL_ENV, PATH, PYTHONHOME) are the stable contract
+// defined by PEP 405 (Python 3.3, 2012). Other tools (Poetry, tox, pipx) use
+// the same approach. Sourcing the shell-specific activate script (activate,
+// activate.fish, Activate.ps1) would be higher maintenance because it varies
+// by shell.
+func ActivateVenvIfPresent(fs afero.Fs, osClient types.Os, projectDir string) (bool, error) {
+	venvPath := getVenvPath(projectDir)
+	if !venvExists(fs, venvPath) {
+		return false, nil
+	}
```
zimeg, line 101: 🐍 question: Do we want to avoid reactivating a virtual environment if one is present? This might be an issue if different environments are activated for different reasons, but I'm unsure if that's a common practice.

I'm alright merging this without a clear answer to extend these approaches based on feedback! The happy path works so well and `.venv` appears to be [convention](https://docs.python.org/3/library/venv.html) without suggesting `.venv.testing` or similar.

internal/shared/clients.go
zimeg: ⚠️ suggestion(non-blocking): Your approach is solid, but we might consider outputting the errors as a warning while keeping expected activations to the debug logs?

Would love to hear your thoughts on this! IIRC logic similar to what's here exists for the `get-hooks` activation, but we often miss problems that should be at least reported. Even if the command continues I think!

mwbrooks: That's a very good idea! Especially if we're already using a similar approach for the `get-hooks` activation.

internal/runtime/python/python.go
mwbrooks: I thought about this as well. I think we want to re-activate the environment every time. While we can detect that a virtual environment is already active, we don't know if it's the virtual environment for this project or a different project. The more reliable approach is to activate it each time. Just my two-cents.

internal/shared/clients.go
mwbrooks: PR: https://github.com/slackapi/slack-cli/pull/368

PR #352, opened 2026-02-24
feat: adds `docs` command with optional search flag

lukegalbraithrussell:
### Changelog

The `slack docs` command opens the slack docs site; its `--search` flag opens the slack docs search page with the provided search query.

### Summary

This PR adds a `docs` command.

| Command | Result |
|---|---|
| `slack docs`  | opens `docs.slack.dev` 
| `slack docs --search` | opens `docs.slack.dev/search` |
| `slack docs --search something cool` | `docs.slack.dev/search/?q=something+cool` | 
| `slack docs something` | oopsie you did mess up there try again maybe with the `--search` flag?|

This doesn't take advantage of the charm experiment. Next steps though!!!
### Requirements

- [X] I've read and understood the [Contributing Guidelines](https://github.com/slackapi/slack-cli/blob/main/.github/CONTRIBUTING.md) and have done my best effort to follow them.
- [X] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).

### DIFF
--- a/internal/iostreams/charm.go
+++ b/internal/iostreams/charm.go
@@ -22,40 +22,50 @@ import (
 	"slices"
 
 	"github.com/charmbracelet/huh"
+	"github.com/slackapi/slack-cli/internal/style"
 )
+
+// buildInputForm constructs a huh form for text input prompts.
+func buildInputForm(message string, cfg InputPromptConfig, input *string) *huh.Form {
+	field := huh.NewInput().
+		Title(message).
+		Value(input)
+	if cfg.Required {
+		field.Validate(huh.ValidateMinLength(1))
+	}
+	return huh.NewForm(huh.NewGroup(field)).WithTheme(style.ThemeSlack())
+}
 
 // charmInputPrompt prompts for text input using a charm huh form
 func charmInputPrompt(_ *IOStreams, _ context.Context, message string, cfg InputPromptConfig) (string, error) {
 	var input string
-	field := huh.NewInput().
-		Title(message).
-		Value(&input)
-	if cfg.Required {
-		field.Validate(huh.ValidateMinLength(1))
-	}
-	err := huh.NewForm(huh.NewGroup(field)).WithTheme(ThemeSlack()).Run()
+	err := buildInputForm(message, cfg, &input).Run()
 	if err != nil {
 		return "", err
 	}
 	return input, nil
 }
 
+// buildConfirmForm constructs a huh form for yes/no confirmation prompts.
+func buildConfirmForm(message string, choice *bool) *huh.Form {
+	field := huh.NewConfirm().
+		Title(message).
+		Value(choice)
+	return huh.NewForm(huh.NewGroup(field)).WithTheme(style.ThemeSlack())
+}
+
 // charmConfirmPrompt prompts for a yes/no confirmation using a charm huh form
 func charmConfirmPrompt(_ *IOStreams, _ context.Context, message string, defaultValue bool) (bool, error) {
 	var choice = defaultValue
-	field := huh.NewConfirm().
-		Title(message).
-		Value(&choice)
-	err := huh.NewForm(huh.NewGroup(field)).WithTheme(ThemeSlack()).Run()
+	err := buildConfirmForm(message, &choice).Run()
 	if err != nil {
 		return false, err
 	}
 	return choice, nil
 }
 
-// charmSelectPrompt prompts the user to select one option using a charm huh form
-func charmSelectPrompt(_ *IOStreams, _ context.Context, msg string, options []string, cfg SelectPromptConfig) (SelectPromptResponse, error) {
-	var selected string
+// buildSelectForm constructs a huh form for single-selection prompts.
+func buildSelectForm(msg string, options []string, cfg SelectPromptConfig, selected *string) *huh.Form {
 	var opts []huh.Option[string]
 	for _, opt := range options {
 		key := opt
@@ -70,13 +80,19 @@ func charmSelectPrompt(_ *IOStreams, _ context.Context, msg string, options []st
 	field := huh.NewSelect[string]().
 		Title(msg).
 		Options(opts...).
-		Value(&selected)
+		Value(selected)
 
 	if cfg.PageSize > 0 {
 		field.Height(cfg.PageSize + 2)
 	}
 
-	err := huh.NewForm(huh.NewGroup(field)).WithTheme(ThemeSlack()).Run()
+	return huh.NewForm(huh.NewGroup(field)).WithTheme(style.ThemeSlack())
+}
+
+// charmSelectPrompt prompts the user to select one option using a charm huh form
+func charmSelectPrompt(_ *IOStreams, _ context.Context, msg string, options []string, cfg SelectPromptConfig) (SelectPromptResponse, error) {
+	var selected string
+	err := buildSelectForm(msg, options, cfg, &selected).Run()
 	if err != nil {
 		return SelectPromptResponse{}, err
 	}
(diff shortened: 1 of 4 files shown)


cmd/docs/docs_test.go
```diff
@@ -0,0 +1,106 @@
...
+	"testing"
+
+	"github.com/slackapi/slack-cli/internal/shared"
+	"github.com/slackapi/slack-cli/internal/slacktrace"
+	"github.com/slackapi/slack-cli/test/testutil"
+	"github.com/spf13/cobra"
+	"github.com/stretchr/testify/mock"
+)
+
+func Test_Docs_DocsCommand(t *testing.T) {
```
lukegalbraithrussell, line 28: valid if these are too many tests lol

cmd/docs/docs.go
```diff
@@ -0,0 +1,91 @@
...
+		Emoji: "books",
+		Text:  sectionText,
+		Secondary: []string{
+			docsURL,
+		},
+	}))
+
+	clients.Browser().OpenURL(docsURL)
+
+	// Add trace for analytics
```
zimeg: 🧪 note: We use these traces for testing certain paths are reached in our E2E tests:

```
$ SLACK_TEST_TRACE=1 slack docs
```

It might be nice to update or remove this comment to avoid ongoing confusion toward this!

cmd/docs/docs.go
```diff
@@ -0,0 +1,91 @@
...
+		sectionText = "Slack developer docs"
+	}
+
+	clients.IO.PrintInfo(ctx, false, "\n%s", style.Sectionf(style.TextSection{
+		Emoji: "books",
+		Text:  sectionText,
+		Secondary: []string{
+			docsURL,
+		},
+	}))
```
zimeg, line 100: 📠 suggestion: Our sytles are not so consistent but we might want to use the command name for "text" with more details following?

```
📚 Docs Open
   https://docs.slack.dev/
```

Or

```
📚 Docs Search
   https://docs.slack.dev/search?q=chatbot
```

cmd/docs/docs_test.go
zimeg: @lukegalbraithrussell These are solid cases to cover! It makes me confident that we're covering edge cases from the start 🤓

cmd/docs/docs.go
lukegalbraithrussell: super good point! changed it

cmd/docs/docs_test.go
```diff
@@ -0,0 +1,106 @@
...
+			ExpectedAsserts: func(t *testing.T, ctx context.Context, cm *shared.ClientsMock) {
+				expectedURL := "https://docs.slack.dev"
+				cm.Browser.AssertCalled(t, "OpenURL", expectedURL)
+				cm.IO.AssertCalled(t, "PrintTrace", mock.Anything, slacktrace.DocsSuccess, mock.Anything)
+			},
+			ExpectedOutputs: []string{
+				"Docs Open",
+				"https://docs.slack.dev",
+			},
+		},
```
zimeg: 🧪 issue: This case is a bit misleading to me! I tried using the search flag without an argument to find this error:

```
$ slack docs --search
Check /Users/eden.zimbelman/.slack/logs/slack-debug-20260226.log for error logs
flag needs an argument: --search
```

> should this instead open `docs.slack.dev/search`?

I do think this would be ideal - perhaps "https://docs.slack.dev/search/" complete - if possible, without an argument?

cmd/root.go
```diff
@@ -95,6 +96,7 @@ func NewRootCommand(clients *shared.ClientFactory, updateNotification *update.Up
 			{Command: "init", Meaning: "Initialize an existing Slack app"},
 			{Command: "run", Meaning: "Start a local development server"},
 			{Command: "deploy", Meaning: "Deploy to the Slack Platform"},
+			{Command: "docs", Meaning: "Open Slack developer docs"},
```
zimeg, line 99: 📚 praise: This is nice to find toward the bottom of `help` alongside other calls to action!

cmd/docs/docs.go
```diff
@@ -0,0 +1,90 @@
...
+				Command: "docs --search 'Block Kit'",
+			},
+		}),
+		Args: cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runDocsCommand(clients, cmd, args)
+		},
+	}
+
+	cmd.Flags().StringVar(&searchFlag, "search", "", "search query for Slack docs")
```
zimeg: 🪬 suggestion: Related to the "required" search value if the "--search" flag is provided - I wonder if we can change this to be a "boolean" value?

This might allow this command to accept 1 argument instead of 0 for sake of being the search query? We might then search if that flag is provided or if an argument is used:

```
$ slack docs --search
$ slack docs "Block Kit"
$ slack docs --search "Block Kit"
```

We ought not need to document the second case IMHO since it's identical to the third, and I don't think such approach limits future changes. Not a blocker here but the examples might need updating before merge!

cmd/docs/docs.go
```diff
@@ -0,0 +1,112 @@
...
+
+// runDocsCommand opens Slack developer docs in the browser
+func runDocsCommand(clients *shared.ClientFactory, cmd *cobra.Command, args []string) error {
+	ctx := cmd.Context()
+
+	var docsURL string
+	var sectionText string
+
+	// Validate: if there are arguments, --search flag must be used
+	if len(args) > 0 && !cmd.Flags().Changed("search") {
```
zimeg, line 68: 🪬 suggestion: We might still want to keep a max arguments setting for this command, but set to `1`. It can be confusing otherwise to find:

```
(.venv) $ lack docs --search one two

📚 Docs Search
   https://docs.slack.dev/search/?q=one

```

Although we can also update this section with "variadic" arguments to join the arguments together - IMHO this might be ideal if not so complicated 👻

zimeg (commented):
💡 @lukegalbraithrussell This is getting so good! I'm leaving one more comment of edge cases found, but I'm so optimistic!

cmd/docs/docs.go
lukegalbraithrussell: it now joins the arguments together!

`slack docs --search one two three` now returns the search query of "one two three"

cmd/docs/docs.go
```diff
@@ -0,0 +1,115 @@
...
+				Meaning: "Open Slack docs search page",
+				Command: "docs --search",
+			},
+		}),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runDocsCommand(clients, cmd, args)
+		},
+	}
+
+	cmd.Flags().BoolVar(&searchMode, "search", false, "open Slack docs search page or search with query")
```
zimeg, line 55:
```suggestion
	cmd.Flags().Bool("search", false, "open Slack docs search page or search with query")
```

🪓 suggestion(non-blocking): We can perhaps remove `searchMode` as a variable with the lookup patterns that follow. IMHO it's a nice change that we could bring to other commands...

cmd/docs/docs.go
```diff
@@ -0,0 +1,115 @@
...
+
+	"github.com/slackapi/slack-cli/internal/shared"
+	"github.com/slackapi/slack-cli/internal/slackerror"
+	"github.com/slackapi/slack-cli/internal/slacktrace"
+	"github.com/slackapi/slack-cli/internal/style"
+	"github.com/spf13/cobra"
+)
+
+var searchMode bool
+
```
zimeg, line 30:
```suggestion
```

🪓 suggesetion(non-blocking): Related to the earlier comment!

zimeg (approved):
@lukegalbraithrussell This is so awesome alongside the fast load times 🤩 ✨ 

Thanks so much for all of the iteration toward this. It's in a solid place to ship I'm finding - LGTM 🚢 💨

PR #366, opened 2026-03-06
refactor: remove logger from create command

mwbrooks:
### Changelog

- N/A

### Summary

This pull request is part 2 of removing internal/logger package and is focused on the the `create` command.

- The `appCreateSpinner` is removed because it was never used
  - The production code successfully stopped it, but it never started it. Stops are ignored when the spinner has not started.
  - The installing dependencies spinner continues to display as it does in production
  - I looked into re-adding it, but the spinner but it no longer looks good with our new create UX

### Test Steps

```bash
$ lack create my-app
# → Select "Starter app"
# → Select "Bolt for Python"
# → Confirm spinner displayed when installing dependencies
# → Confirm all output looks as expected

# Cleanup
$ rm -rf my-app/
```

### Requirements

### DIFF
--- a/cmd/project/create_test.go
+++ b/cmd/project/create_test.go
@@ -548,7 +548,7 @@ func TestCreateCommand(t *testing.T) {
 					return nil
 				}
 				createClientMock = new(CreateClientMock)
-				createClientMock.On("Create", mock.Anything, mock.Anything, mock.Anything, mock.Anything).Return("", nil)
+				createClientMock.On("Create", mock.Anything, mock.Anything, mock.Anything).Return("", nil)
 				CreateFunc = createClientMock.Create
 			},
 			Teardown: func() {
@@ -561,7 +561,7 @@ func TestCreateCommand(t *testing.T) {
 					AppName:  "my-charm-app",
 					Template: template,
 				}
-				createClientMock.AssertCalled(t, "Create", mock.Anything, mock.Anything, mock.Anything, expected)
+				createClientMock.AssertCalled(t, "Create", mock.Anything, mock.Anything, expected)
 				// Verify that the survey-based SelectPrompt for category was NOT called
 				cm.IO.AssertNotCalled(t, "SelectPrompt", mock.Anything, "Select an app:", mock.Anything, mock.Anything)
 			},
@@ -586,7 +586,7 @@ func TestCreateCommand(t *testing.T) {
 			},
 			ExpectedErrorStrings: []string{"user cancelled"},
 			ExpectedAsserts: func(t *testing.T, ctx context.Context, cm *shared.ClientsMock) {
-				createClientMock.AssertNotCalled(t, "Create", mock.Anything, mock.Anything, mock.Anything, mock.Anything)
+				createClientMock.AssertNotCalled(t, "Create", mock.Anything, mock.Anything, mock.Anything)
 			},
 		},
 		"lists agent templates with agent --list flag": {


.claude/CLAUDE.md
```diff
@@ -31,6 +31,10 @@ make lint               # Run golangci-lint
 golangci-lint --version # Verify linter version
 ```
 
+### Formatting
+
+Always run `gofmt -w` on changed Go files after making edits to ensure proper formatting.
+
```
mwbrooks, line 37: note: Just adding this to scratch an itch where Claude fails to format files correctly.

cmd/project/create.go
```diff
@@ -43,11 +42,6 @@ var createSubdirFlag string
 // TODO - Find best practice, such as using an Interface and Struct to create a client
 var CreateFunc = create.Create
 
-var appCreateSpinner *style.Spinner
-
-const copyTemplate = "Copying"
-const cloneTemplate = "Cloning"
-
```
mwbrooks, line 50: note: Removing because the spinner was never started in the current code.

cmd/project/create.go
```diff
@@ -177,88 +165,15 @@ func runCreateCommand(clients *shared.ClientFactory, cmd *cobra.Command, args []
...
-// newCreateLogger creates a logger instance to receive event notifications
-func newCreateLogger(clients *shared.ClientFactory, cmd *cobra.Command) *logger.Logger {
-	return logger.New(
-		// OnEvent
-		func(event *logger.LogEvent) {
-			switch event.Name {
-			case "on_app_create_template_default":
-				printAppCreateDefaultemplate(cmd, event)
-			case "on_app_create_template_custom":
-				printAppCreateCustomTemplate(cmd, event)
```
mwbrooks, line 203: note: Neither of these events were ever called. This is why the spinner was never started.

cmd/project/create.go
```diff
@@ -177,88 +165,15 @@ func runCreateCommand(clients *shared.ClientFactory, cmd *cobra.Command, args []
...
-			case "on_app_create_completion":
-				printProjectCreateCompletion(clients, cmd, event)
```
mwbrooks, line 205: note: This event was correctly called, but the spinner was never started so any calls to stop it are ignored.

internal/pkg/create/create.go
```diff
@@ -222,15 +218,7 @@ func parentDirExists(dirPath string) (bool, error) {
...
 
 // createApp will create the app directory using the default app template or a specified template URL.
-func createApp(ctx context.Context, dirPath string, template Template, gitBranch string, log *logger.Logger, fs afero.Fs) error {
-	log.Data["templatePath"] = template.path
-	log.Data["isGit"] = template.isGit
-	log.Data["gitBranch"] = gitBranch
-	log.Data["isSample"] = template.IsSample()
-
-	// Notify listeners
-	log.Log("info", "on_app_create_template")
```
mwbrooks, line 246: note: This event name doesn't exist 🤦🏻 It should have been `on_app_create_template_[default|custom]`.

mwbrooks (commented):
Comments for the patient reviewers 🙇🏻

cmd/project/create.go
zimeg: ⭐ praise: @mwbrooks This is upsetting to find! I'm sure these codeblocks have sent others down a rabbit hole before... Let's keep things simple! Thanks for these changes.

mwbrooks: @zimeg Totally agree, I imagine others have followed this logic and weaved their changes into it. It would have been very difficult to debug only to realize that the code is never actually executed. 😢

mwbrooks:
Thanks again for the review @zimeg! 🎉 

It looks like the create command is ready to be merged free from the `package logger`! 🚀

PR #379, opened 2026-03-10
feat: list your developer sandboxes

vegeris:
### Changelog

> Experimental Feature: --experiment sandboxes
> 
> We're adding support for managing Slack developer sandboxes from within the CLI.
> 
> slack sandbox list --experiment  sandboxes

### Summary

This PR adds a `sandbox list` command, which will list the user's developer sandboxes. More specifically, it lists all active or archived sandboxes owned by the developer account with the same email address as the auth being used in the request.

#### Testing

Try out the command:
```
% hermes sandbox list --experiment=sandboxes 
```

### Requirements


cmd/sandbox/sandbox.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+func requireSandboxExperiment(clients *shared.ClientFactory) error {
+	if !clients.Config.WithExperimentOn(experiment.Sandboxes) {
+		return slackerror.New(slackerror.ErrMissingExperiment).
+			WithMessage("%sThe sandbox management commands are under construction", style.Emoji("construction")).
+			WithRemediation("To try them out, just add the --experiment=sandboxes flag to your command!")
+	}
+	return nil
+}
```
vegeris, line 63: Preview:
```
%  hermes sandbox
Check /Users/evegeris/.slack/logs/slack-debug-20260310.log for error logs

🚫 🚧 The sandbox management commands are under construction (missing_experiment)

💡 Suggestion
   To try them out, just add the --experiment=sandboxes flag to your command!
```

cmd/sandbox/list.go
```diff
@@ -0,0 +1,143 @@
...
+			cmd.Printf("    %s\n", style.Secondary(fmt.Sprintf("%s %s", label, archivedTime.Format(timeFormat))))
+		}
+
+		cmd.Println()
+	}
+
+	clients.IO.PrintInfo(ctx, false, "Learn more at %s", style.Secondary("https://docs.slack.dev/tools/developer-sandboxes"))
+
+	return nil
+}
```
vegeris, line 150: Preview:
```
% hermes sandbox list  --experiment=sandboxes 
? Select a team for authentication happy-clam E0ACE2P3GDB

🏖️  Developer Sandboxes
   Owned by Slack developer account elainevegeris@gmail.com

  Happy Clam (E0ACE2P3GDB)
    URL: https://happy-clam.slack.com
    Created: 2026-02-02
    🟢 Status: ACTIVE
    Active until: 2027-02-02

Learn more at https://docs.slack.dev/tools/developer-sandboxes
```

cmd/sandbox/list.go
```diff
@@ -0,0 +1,143 @@
...
+			email = userInfo.Profile.Email
+		}
+	}
+
+	section := style.TextSection{
+		Emoji: "beach_with_umbrella",
+		Text:  " Developer Sandboxes",
+	}
+
+	if email != "" {
```
vegeris, line 100: Email wouldn't be included in the users.info response for existing auths as we just today added the `users:read.email` scope which allows newly authed tokens to access the email address

tldr; users will need to logout / log back in to see this printed

internal/api/sandbox.go
```diff
@@ -0,0 +1,87 @@
...
+		for _, v := range listSandboxFilterEnum {
+			if status == v {
+				valid = true
+				break
+			}
+		}
+		if !valid {
+			return nil, errInvalidArguments.WithRootCause(fmt.Errorf("allowed values for sandbox status filter: %v", listSandboxFilterEnum))
+		}
+	}
```
zimeg: 👁️‍🗨️ question: Is this validation we can leave to the API? I'm not so familiar with the `developer.sandbox.list` method at this time, but this might require a separate patch release if changed-

👾 ramble: Moving this toward the `cmd` paths might also be nice! I'm perhaps hoping most to keep the logic of `api` to minimum?

cmd/sandbox/auth_test.go
zimeg, line 1: 🧪 quibble: Can we use the `testutil.TableTestCommand` methods for these tests? IIRC this is meant most for the `cmd` callsites:

https://github.com/slackapi/slack-cli/blob/2ae86a3c553348a88b4f6fd7b4afde340cff3965/cmd/auth/login_test.go#L44

cmd/sandbox/sandbox.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/slackapi/slack-cli/internal/shared"
+	"github.com/slackapi/slack-cli/internal/slackerror"
+	"github.com/slackapi/slack-cli/internal/style"
+	"github.com/spf13/cobra"
+)
+
+func NewCommand(clients *shared.ClientFactory) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "sandbox <subcommand> [flags] --experiment=sandboxes",
+		Short: "Manage your sandboxes",
```
zimeg: 📚 suggestion: Can we include an alias for `sandboxes`? I tried this to list those known but found error with:

```
$ slack sandboxes list --experiment sandboxes
```

cmd/sandbox/list.go
```diff
@@ -0,0 +1,143 @@
...
+	clients.IO.PrintInfo(ctx, false, "%s", style.Sectionf(section))
+
+	if len(sandboxes) == 0 {
+		clients.IO.PrintInfo(ctx, false, "%s\n", style.Secondary("No sandboxes found. Create one with `slack sandbox create --name <name>`"))
+		return nil
+	}
+
+	timeFormat := "2006-01-02" // We only support the granularity of the day for now, rather than a more precise datetime
+	for _, s := range sandboxes {
+		cmd.Printf("  %s (%s)\n", style.Bold(s.SandboxName), s.SandboxTeamID)
```
zimeg: 🔍 quibble(non-blocking): Unsure if this makes later changes more clear, but can we append these lines to a `[]string` for outputs toward the end of the loop?

cmd/sandbox/list.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	if email != "" {
+		section.Secondary = []string{fmt.Sprintf("Owned by Slack developer account %s", email)}
+	}
+
+	clients.IO.PrintInfo(ctx, false, "%s", style.Sectionf(section))
+
+	if len(sandboxes) == 0 {
+		clients.IO.PrintInfo(ctx, false, "%s\n", style.Secondary("No sandboxes found. Create one with `slack sandbox create --name <name>`"))
```
zimeg:
```suggestion
		clients.IO.PrintInfo(ctx, false, "No sandboxes found. Create one with %s", style.Commandf("sandbox create --name <name>"))
```

🌚 suggestion: We can avoid aliased binaries with this!

cmd/sandbox/auth.go
```diff
@@ -0,0 +1,129 @@
...
+	// Check for the global --team flag
+	if clients.Config.TeamFlag != "" {
+		for _, auth := range auths {
+			if auth.TeamID == clients.Config.TeamFlag || auth.TeamDomain == clients.Config.TeamFlag {
+				return &auth, nil
+			}
+		}
+		return nil, slackerror.New(slackerror.ErrTeamNotFound).
+			WithMessage("No auth found for team: %s", clients.Config.TeamFlag).
+			WithRemediation("Run 'slack auth list' to see your authorized workspaces")
```
zimeg:
```suggestion
			WithRemediation("Run %s to see your authorized workspaces", style.Commandf("auth list", false))
```

🎨 suggestion: For the unusual aliases-

cmd/sandbox/auth.go
```diff
@@ -0,0 +1,129 @@
...
+
+	auths, err := clients.Auth().Auths(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(auths) == 0 {
+		return nil, slackerror.New(slackerror.ErrCredentialsNotFound).
+			WithMessage("You must be logged in to manage sandboxes").
+			WithRemediation("Run 'slack login' to authenticate, or use --token for CI/CD")
```
zimeg:
```suggestion
			WithRemediation("Run %s to authenticate, or use --token for CI/CD", style.Commandf("login", false))
```

📝 suggestion: I had the wrong command output at first!

cmd/sandbox/auth.go
```diff
@@ -0,0 +1,129 @@
...
+
+	return auth.Token, auth, nil
+}
+
+// resolveAuthForSandbox determines what auth to use for sandbox operations.
+// If the global --token flag is set, we will use that.
+// Else if the global --team flag is set, we will use the associated token for that team.
+// Else if the user is only logged in to one team, we will default to that team's auth.
+// Else we will prompt the user to select a team to use for authentication.
+func resolveAuthForSandbox(ctx context.Context, clients *shared.ClientFactory) (*types.SlackAuth, error) {
```
zimeg: 👾 note: We might want to make a standard "team" selection for authentications, separate from saved apps. IIRC the `app link` command has a similar selection that uses similar logic. Not a blocker for merging this, but I'm curious if we could reuse these selections or if `sandbox` authentications have edge cases?

cmd/sandbox/auth.go
```diff
@@ -0,0 +1,129 @@
...
+	switch {
+	case selection.Flag:
+		for _, opt := range options {
+			if opt.auth.TeamID == selection.Option || opt.auth.TeamDomain == selection.Option {
+				return &opt.auth, nil
+			}
+		}
+		return nil, slackerror.New(slackerror.ErrTeamNotFound).
+			WithMessage("No auth found for team: %s", selection.Option).
+			WithRemediation("Run 'slack auth list' to see your authorized workspaces")
```
zimeg:
```suggestion
			WithRemediation("Run %s to see your authorized workspaces", style.Commandf("auth list", false))
```

🎯 suggestion: IIRC this is an alright pattern if `style` is imported?

cmd/sandbox/list.go
```diff
@@ -0,0 +1,143 @@
...
+
+func runListCommand(cmd *cobra.Command, clients *shared.ClientFactory) error {
+	ctx := cmd.Context()
+
+	token, auth, err := getSandboxAuth(ctx, clients)
+	if err != nil {
+		return err
+	}
+
+	fmt.Println()
```
zimeg:
```suggestion
```

🪓 todo: Hoping we can remove outputs outside of the `clients.IO` packages. Ongoing efforts to standardize format might appreciate this, but also I fear `fmt` causes outputs in tests which are confusing to debug.

zimeg (approved):
@vegeris Super cool to see sandboxes brought to the CLI! Thanks for these changes 🤩 ✨ 

I'm approving with test commands working well, but am leaving a few notes on output and testing conventions that might be nice to make before merging. Also I leave a question to the `sandboxes` alias but error outputs have such helpful remediation 🚢 💨 

📚 Also I added the `semver:patch` with an `experiment` label since this is alright to release without calling attention in a "minor" release, though I hope we have enhancements adjacent!

cmd/sandbox/auth_test.go
vegeris: 💯

cmd/sandbox/auth.go
vegeris: Oh yes--that can be reused! I'll move the function in `app link` into the prompts package and it can be imported into both

cmd/sandbox/list.go
vegeris: How about `clients.IO.PrintInfo(ctx, false, "")` instead? 👀

cmd/sandbox/list.go
```diff
@@ -0,0 +1,143 @@
...
+	filter string
+}
+
+var listCmdFlags listFlags
+
+func NewListCommand(clients *shared.ClientFactory) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "list [flags]",
+		Short: "List your sandboxes",
+		Long:  `List details of your sandboxes`,
```
mwbrooks: suggestion: A small suggestion to add a little more detail to these descriptions. Often we forget to go back and improve these later, so I thought I'd suggest it while it's top-of-mind.

```suggestion
		Short: "List developer sandboxes",
		Long: strings.Join([]string{
			"List details about your developer sandboxes.",
			"",
			"The listed developer sandboxes belong to a developer program account",
			"that matches the email address of the authenticated user.",
		}, "\n"),
```

Image

cmd/sandbox/list.go
```diff
@@ -0,0 +1,143 @@
...
+		Example: style.ExampleCommandsf([]style.ExampleCommand{
+			{Command: "sandbox list", Meaning: "List your sandboxes"},
```
mwbrooks: suggestion: Removing the word "your"
```suggestion
			{Command: "sandbox list", Meaning: "List developer sandboxes"},
```

cmd/sandbox/list.go
```diff
@@ -0,0 +1,143 @@
...
+		}),
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			return requireSandboxExperiment(clients)
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runListCommand(cmd, clients)
+		},
+	}
+
+	cmd.Flags().StringVar(&listCmdFlags.filter, "filter", "", "Filter by status: active, archived")
```
mwbrooks: suggestion(non-blocking): Are there more than 2 statuses?

Instead of `--filter`, I'd suggest `--status=active` or `--status=archived`. I believe we can also accept a list, so `--status=active,archived` would work as well (if the API supports multiple choices).

I have 2 reasons to not use the term filter:
1. Whenever possible, it's best to match the API arg name (`status`). This builds on developer existing knowledge and helps to avoid fragmenting our documentation with a bunch of words that mean the same thing.
2. Filter implies some advanced filtering where this is a simple show/hide.

cmd/sandbox/list.go
```diff
@@ -0,0 +1,143 @@
...
+	email := ""
+	if auth != nil && auth.UserID != "" {
+		if userInfo, err := clients.API().UsersInfo(ctx, token, auth.UserID); err == nil && userInfo.Profile.Email != "" {
+			email = userInfo.Profile.Email
+		}
+	}
+
+	section := style.TextSection{
+		Emoji: "beach_with_umbrella",
+		Text:  " Developer Sandboxes",
```
mwbrooks, line 95: suggestion: I noticed a double-space on my terminal. Some emojis add 2 spaces, while others add a single. I think we can remove this extra space (unless you experienced something different on your terminal)
```suggestion
		Text:  "Developer Sandboxes",
```

Image

cmd/sandbox/list.go
mwbrooks: suggestion: Perhaps you could add a comment above this conditional to explain why we're checking for an unset email address - that it's for legacy login reasons.

cmd/sandbox/list.go
mwbrooks: heh, I was going to suggest the same thing! 🧠

cmd/sandbox/list.go
```diff
@@ -0,0 +1,143 @@
...
+		if s.DateCreated > 0 {
+			cmd.Printf("    %s\n", style.Secondary(fmt.Sprintf("Created: %s", time.Unix(s.DateCreated, 0).Format(timeFormat))))
+		}
+
+		if s.Status != "" {
+			status := style.Secondary(fmt.Sprintf("Status: %s", strings.ToTitle(s.Status)))
+			if strings.EqualFold(s.Status, "archived") {
+				cmd.Printf("    %s %s\n", style.Emoji("warning"), status)
+			} else {
+				cmd.Printf("    %s%s\n", style.Emoji("green_circle"), status)
```
mwbrooks: suggestion: I'd suggest removing this emoji. We try to not over-do the emojis. You could use the green text colour `style.Styler().Green(status),` instead.

If emojis are an absolute must, it should be at the end of the list instead of the beginning.

cmd/sandbox/list.go
```diff
@@ -0,0 +1,143 @@
...
+			}
+		}
+
+		if s.DateArchived > 0 {
+			archivedTime := time.Unix(s.DateArchived, 0).In(time.Local)
+			now := time.Now()
+			archivedDate := time.Date(archivedTime.Year(), archivedTime.Month(), archivedTime.Day(), 0, 0, 0, 0, time.Local)
+			todayDate := time.Date(now.Year(), now.Month(), now.Day(), 0, 0, 0, 0, time.Local)
+			label := "Active until:"
```
mwbrooks: suggestion: "Active until" reads kinda odd. Would "Expires:" be accurate?

cmd/sandbox/list.go
```diff
@@ -0,0 +1,143 @@
...
+	for _, s := range sandboxes {
+		cmd.Printf("  %s (%s)\n", style.Bold(s.SandboxName), s.SandboxTeamID)
+
+		if s.SandboxDomain != "" {
+			cmd.Printf("    %s\n", style.Secondary(fmt.Sprintf("URL: https://%s.slack.com", s.SandboxDomain)))
+		}
+
+		if s.DateCreated > 0 {
+			cmd.Printf("    %s\n", style.Secondary(fmt.Sprintf("Created: %s", time.Unix(s.DateCreated, 0).Format(timeFormat))))
+		}
```
mwbrooks: suggestion: I'd suggest grouping the dates together: Created and Expires (Active until) since they are related information.

cmd/sandbox/sandbox.go
mwbrooks: agreed: We often have aliases for the plural or singular form. So, it makes sense to do that here too.

cmd/sandbox/sandbox.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/slackapi/slack-cli/internal/shared/types"
+	"github.com/slackapi/slack-cli/internal/slackerror"
+	"github.com/slackapi/slack-cli/internal/style"
+	"github.com/spf13/cobra"
+)
+
+func NewCommand(clients *shared.ClientFactory) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "sandbox <subcommand> [flags] --experiment=sandboxes",
+		Short: "Manage your sandboxes",
```
mwbrooks: suggestion: Swapping the word "your" for "developer".
```suggestion
		Short: "Manage developer sandboxes",
```

cmd/sandbox/sandbox.go
```diff
@@ -0,0 +1,82 @@
...
+		Long: `Manage your Slack developer sandboxes without leaving your terminal.
```
mwbrooks: praise: Thanks so much for adding a nice long description!

cmd/sandbox/sandbox.go
```diff
@@ -0,0 +1,82 @@
...
+Use the --team flag to select the authentication to use for these commands.
+
+Prefer a UI? Head over to {{LinkText "https://api.slack.com/developer-program/sandboxes"}}
+
+New to the Developer Program? Sign up at {{LinkText "https://api.slack.com/developer-program/join"}}`,
```
mwbrooks: suggestion: Can we display the URLs on a separate newline, please? URLs tend to exceed the 80 character length of default terminal windows.

cmd/sandbox/sandbox.go
```diff
@@ -0,0 +1,82 @@
...
+			WithMessage("%sThe sandbox management commands are under construction", style.Emoji("construction")).
+			WithRemediation("To try them out, just add the --experiment=sandboxes flag to your command!")
+	}
+	return nil
+}
+
+// getSandboxAuth returns the auth to be used for sandbox management.
+// Uses the global --token or --team flag if present, otherwise prompts the user to select a team.
+func getSandboxAuth(ctx context.Context, clients *shared.ClientFactory) (*types.SlackAuth, error) {
+
```
mwbrooks: nit: No empty newlines at the start of a function
```suggestion
```

cmd/root.go
```diff
@@ -175,6 +176,7 @@ func Init(ctx context.Context) (*cobra.Command, *shared.ClientFactory) {
 		openformresponse.NewCommand(clients),
 		platform.NewCommand(clients),
 		project.NewCommand(clients),
+		sandbox.NewCommand(clients),
```
mwbrooks, line 179: praise: Alphabetical! 🔡

internal/api/sandbox.go
mwbrooks: Agreed, if we can leave this validation to the API then the code will be more future proof. Since it's a tool installed locally on people's machines, local validation logic can sometimes block API improvements.

internal/experiment/experiment.go
```diff
@@ -41,6 +41,9 @@ const (
 	// Charm experiment enables beautiful prompts.
 	Charm Experiment = "charm"
 
+	// Sandboxes experiment lets users who have joined the Slack Developer Program use the CLI to manage their sandboxes.
+	Sandboxes Experiment = "sandboxes"
+
```
mwbrooks, line 46: suggestion: Would you mind documenting this experiment here as well:

https://github.com/slackapi/slack-cli/blob/main/docs/reference/experiments.md#available-experiments

internal/shared/types/sandbox.go
```diff
@@ -0,0 +1,25 @@
...
+package types
+
+// Sandbox represents a Slack Developer Sandbox from the developer.sandbox.list API.
+type Sandbox struct {
+	SandboxTeamID string `json:"sandbox_team_id"` // Encoded team ID of the developer sandbox
+	SandboxName   string `json:"sandbox_name"`    // Name of the developer sandbox
+	SandboxDomain string `json:"sandbox_domain"`  // Domain of the developer sandbox
+	DateCreated   int64  `json:"date_created"`    // When the developer sandbox was created, as epoch seconds
+	DateArchived  int64  `json:"date_archived"`   // When the developer sandbox is or will be archived, as epoch seconds
+	Status        string `json:"status"`          // Status of the developer sandbox: Active or Archived
```
mwbrooks: nit(non-blocking): Appreciate if it these are alphabetized to make it easier to maintain in the future.

cmd/sandbox/list.go
vegeris: Good point! I'll add that comment

internal/api/sandbox.go
vegeris: That's fine by me; the only caveat is that the API will only return a more generic error, `🚫 Slack API request parameters are invalid (invalid_arguments)`. The enum value for this input param is defined in the API contract rather than checked in the handler; in the handler we would have have the option of returning an extended error with more details

Just double-checked the full API response; it does point to the status param but doesn't list valid enum values:
```
"response_metadata": {
        "messages": [
            "[ERROR] must be a valid enum value [json-pointer:/status]"
        ]
    }
    ```

File: cmd/sandbox/sandbox.go (hunk 5 above)
Comment (vegeris) [reply cmd/sandbox/sandbox.go]: 💯

File: cmd/sandbox/list.go (hunk 6 above)
Comment (vegeris) [reply cmd/sandbox/list.go]: Oh I just realised I'm using `cmd.Printf()` a lot; I'm guessing we want that switched over to `clients.IO.PrintInfo()`

Edit: updated that

File: cmd/sandbox/list.go (hunk 7 above)
Comment (vegeris) [reply cmd/sandbox/list.go]: Sounds good!

Comment (vegeris) [reply cmd/sandbox/list.go]: Oh, actually I removed that line for now because the `create` command isn't being added in this PR 🤡

File: cmd/sandbox/auth.go (hunk 8 above)
Comment (vegeris) [reply cmd/sandbox/auth.go]: (This bit was removed as it was redundant)

File: cmd/sandbox/auth.go (hunk 11 above)
Comment (vegeris) [reply cmd/sandbox/auth.go]: (print statement removed due to being redundant)

File: cmd/sandbox/list.go (hunk 15 above)
Comment (vegeris) [reply cmd/sandbox/list.go]: Works for me (I believe it's just active or archived; sandboxes can also be 'deleted' but we're not returning those)

File: internal/experiment/experiment.go (hunk 25 above)
Comment (vegeris) [reply internal/experiment/experiment.go:46]: Will do! Thanks for calling that out

Comment (vegeris) [reply internal/experiment/experiment.go:46]: By the way, is it preferred to include a reference to experimental features in the Changelog, or to leave them out? (I had assumed we left them out but would be fine with documenting it with the expectation that things could break)

File: cmd/sandbox/list.go (hunk 17 above)
Comment (vegeris) [reply cmd/sandbox/list.go]: Sure; I'll do the green text colour 🙂

Comment (vegeris) [reply cmd/sandbox/list.go]: Using `clients.IO.PrintInfo(ctx, false, "    %s", style.Styler().Green(status))` I don't actually see the colour applied 👀 Not sure if that's specific to my terminal. It's fine either way though

File: cmd/sandbox/list.go (hunk 18 above)
Comment (vegeris) [reply cmd/sandbox/list.go]: I was debating between `Will be archived` or `Active until` because `Archived` sounds like the sandbox has already been archived. `Expires` sounds good 🙂 

(After the archive date, the sandbox is scheduled for deletion, which is currently 90 days after it's archived IIRC. It can be unarchived at any time before it's deleted and used normally, though the user would need to reactivate their users etc.)

File: cmd/sandbox/sandbox.go (hunk 21 above)
Comment (vegeris) [reply cmd/sandbox/sandbox.go]: I take it we would like to remove 'your' here as well?

File: cmd/sandbox/list.go (hunk 16 above)
Comment (vegeris) [reply cmd/sandbox/list.go:95]: Interesting; in my terminal, I see no space without it added to the text there 
 
Screenshot 2026-03-11 at 11 47 07 AM

File: internal/api/sandbox.go (hunk 4 above)
Comment (vegeris) [reply internal/api/sandbox.go]: Would we be okay with keeping the enum in the help page? (This way users can see currently accepted values but they wouldn't actually be blocked if accepted values change on the backend)

```
FLAGS
  -h, --help            help for list
      --status string   Filter by status: active, archived
```

mwbrooks: 💯 Please always list the enum values in the help page under the flags - just like the above example 🙇🏻 

Re: Slack API `invalid_arguments` - displaying a Slack API error and resolution to a Slack CLI user isn't very helpful anyways. The user would want to know the `--status <active | archived>` flag. So, i think the current approach is looking good!

cmd/sandbox/list.go
mwbrooks: note: Under the hood, `cmd.Printf()` should be using the `clients.IO.Writer` but we generally try to keep things consistent by using the `clients.IO...` for printing. Thanks for updating it!

cmd/sandbox/list.go
mwbrooks: Cool, I think the `--status <active | archived>` design will allow us to also show deleted, if there's ever a decision to return those.

cmd/sandbox/list.go
mwbrooks: Fun times - I imagine it's based on our font choices. Don't sweat it. Looking _way_ ahead, we're going to remove showing emojis (except at the end of lines) to workaround this issue.

cmd/sandbox/sandbox.go
mwbrooks: Sure, that would be great! 🙇🏻

internal/experiment/experiment.go
mwbrooks: We haven't really decided on a standard approach. Personally, I like keeping them as part of the changelog (just prefixed as an experiment) so that the public can try it out. But, for features that are only available internally, it really doesn't matter.

PR #388, opened 2026-03-11
test: add test coverage for untested files

mwbrooks:
### Changelog

- N/A

### Summary

This pull request add unit test coverage for 10 previously untested files across internal packages:

- `internal/api`
- `internal/archiveutil`
- `internal/cmdutil`
- `internal/config`
- `internal/goutils`
- `internal/image`
- `internal/iostreams`
- `internal/ioutils`
- `internal/prompts`

_Only test files were changed in this pull request._

### Requirements

### DIFF
--- a/cmd/app/add.go
+++ b/cmd/app/add.go
@@ -20,7 +20,6 @@ import (
 	"github.com/slackapi/slack-cli/internal/app"
 	"github.com/slackapi/slack-cli/internal/cmdutil"
 	"github.com/slackapi/slack-cli/internal/config"
-	"github.com/slackapi/slack-cli/internal/experiment"
 	"github.com/slackapi/slack-cli/internal/pkg/apps"
 	"github.com/slackapi/slack-cli/internal/prompts"
 	"github.com/slackapi/slack-cli/internal/shared"
@@ -82,9 +81,6 @@ func preRunAddCommand(ctx context.Context, clients *shared.ClientFactory, cmd *c
 	err := cmdutil.IsValidProjectDirectory(clients)
 	if err != nil {
 		return err
-	}
-	if !clients.Config.WithExperimentOn(experiment.BoltFrameworks) {
-		return nil
 	}
 	clients.Config.SetFlags(cmd)
 	return nil
--- a/cmd/app/add_test.go
+++ b/cmd/app/add_test.go
@@ -23,7 +23,6 @@ import (
 	"github.com/slackapi/slack-cli/internal/cache"
 	"github.com/slackapi/slack-cli/internal/cmdutil"
 	"github.com/slackapi/slack-cli/internal/config"
-	"github.com/slackapi/slack-cli/internal/experiment"
 	"github.com/slackapi/slack-cli/internal/iostreams"
 	"github.com/slackapi/slack-cli/internal/prompts"
 	"github.com/slackapi/slack-cli/internal/shared"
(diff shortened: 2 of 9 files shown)


internal/archiveutil/archiveutil_test.go
```diff
@@ -0,0 +1,159 @@
...
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TODO: Refactor to use afero.Fs once ExtractAndWriteFile accepts it. Currently uses t.TempDir() which is safe.
+func Test_ExtractAndWriteFile(t *testing.T) {
```
mwbrooks, line 31: note: I've decided to implement these tests now using the `t.TempDir()` approach, but I've added a comment to refactor the code being tested to inject `afero.Fs` so that we can change the tests to be our memory-base FS.

internal/goutils/recursivecopy_test.go
```diff
@@ -0,0 +1,218 @@
...
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			result := ShouldIgnore(tc.val, tc.list, tc.ignoreFunc)
+			assert.Equal(t, tc.expected, result)
+		})
+	}
+}
+
+// TODO: Refactor to use afero.Fs once Copy accepts it. Currently uses t.TempDir() which is safe.
+func Test_Copy(t *testing.T) {
```
mwbrooks, line 87: note: More reminders to refactor this, but these tests are safe to run since they are in the temp directory (just slower).

internal/goutils/recursivecopy_test.go
```diff
@@ -0,0 +1,218 @@
...
+		// Verify the symlink target is the same
+		linkTarget, err := os.Readlink(dstLink)
+		require.NoError(t, err)
+		assert.Equal(t, targetFile, linkTarget)
+	})
+}
+
+// TODO: Refactor to use afero.Fs once CopyDirectory accepts it. Currently uses t.TempDir() which is safe.
+func Test_CopyDirectory(t *testing.T) {
+	t.Run("copies directory structure", func(t *testing.T) {
```
mwbrooks, line 141: note: Not using Table Tests here because each test case needs unique setup. To find a balance, I'm using the `t.Run("...")` to still group the test cases together in the same spirit as table tests.

internal/archiveutil/archiveutil_test.go
zimeg: @mwbrooks Amazing callout! Having these tests in place will make that change much more confident I think! 🧪 ✨

internal/api/collaborators_test.go
zimeg, line 1: ⭐ praise: These tests aren't so complicated but that's amazing I think! It gives me confidence that existing API methods are gathering expected responses as we expect!

internal/image/image_test.go
zimeg, line 1: 📸 note: TIL about this package!

internal/prompts/permissions_test.go
zimeg, line 1: 🧪 thought: Increasing coverage is appreciated so much but I'm worried that it builds confidence in patterns we're wanting to move from. I don't think command specific prompts ought be separate from commands but let's not address that now!

internal/ioutils/host_test.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_GetHostname(t *testing.T) {
+	t.Run("returns a non-empty hashed hostname", func(t *testing.T) {
+		hostname := GetHostname()
+		assert.NotEmpty(t, hostname)
+		// The hostname should be hashed, not the raw hostname
+		// It should not be "unknown" on a normal system
+		assert.NotEqual(t, "", hostname)
```
zimeg: 🗣️ question(non-blocking): Is not empty the same as not equal to empty string?

internal/archiveutil/archiveutil_test.go
mwbrooks: Happy to hear that you agree. I was on the fence, since we are trying to avoid I/O during testing. But I'll be sure to follow-up and remove these in an upcoming PR.

internal/prompts/permissions_test.go
mwbrooks: thought: I think it's entirely okay to delete old, unrelated tests. So I hope it doesn't discourage future refactors.

internal/ioutils/host_test.go
mwbrooks: @zimeg Good eye! This should have tested for `assert.NotEqual(t, "unknown", hostname)` 🙇🏻 Fixed in commit 758ce76b60b23ec29fff0bebdc5a2f9f2ca99672

PR #389, opened 2026-03-11
feat(experiment): create and delete developer sandboxes

vegeris:
### Changelog

> Experimental Feature: --experiment sandboxes
> 
> We're adding support for managing Slack developer sandboxes from within the CLI. We now allow one to create or delete a sandbox from the CLI.
> - `sandbox create` allows you to create a new developer sandbox.
> - `sandbox delete` allows you to delete an existing developer sandbox.

### Summary

This PR adds `sandbox create` and `sandbox delete` commands, which allow one to create or delete a sandbox from the CLI. 

Follow up to https://github.com/slackapi/slack-cli/pull/379

#### Testing

Try out the commands:
```
% hermes sandbox create --experiment=sandboxes --name "my-test-box" --password "secretPassword"

% hermes sandbox delete --experiment=sandboxes --sandbox-id E012345
```

### Requirements

### DIFF
--- a/cmd/sandbox/create_test.go
+++ b/cmd/sandbox/create_test.go
@@ -25,7 +25,6 @@ import (
 	"github.com/slackapi/slack-cli/internal/shared/types"
 	"github.com/slackapi/slack-cli/test/testutil"
 	"github.com/spf13/cobra"
-	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
 )
 
@@ -555,27 +554,39 @@ func Test_getTemplateID(t *testing.T) {
 }
 
 func Test_domainFromName(t *testing.T) {
-	tests := map[string]struct {
-		in      string
-		want    string
-		wantErr bool
-	}{
-		"simple":           {"test-box", "test-box", false},
-		"spaces":           {"My Test Box", "my-test-box", false},
-		"uppercase":        {"MyBox", "mybox", false},
-		"mixed":            {"Hello_World 123", "hello-world-123", false},
-		"leading trailing": {"-test-", "test", false},
-		"empty":            {"", "", true},
-	}
-	for name, tt := range tests {
-		t.Run(name, func(t *testing.T) {
-			got, err := domainFromName(tt.in)
-			if tt.wantErr {
-				assert.Error(t, err)
-				return
-			}
-			assert.NoError(t, err)
-			assert.Equal(t, tt.want, got)
-		})
-	}
+	testutil.TableTestCommand(t, testutil.CommandTests{
+		"handles invalid URL characters": {
+			CmdArgs: []string{
+				"--experiment=sandboxes",
+				"--token", "xoxb-test-token",
+				"--name", "-Hello_World 123-",
+				"--password", "pass",
+			},
+			Setup: func(t *testing.T, ctx context.Context, cm *shared.ClientsMock, cf *shared.ClientFactory) {
+				setupCreateMocks(t, ctx, cm, "-Hello_World 123-", "hello-world-123", "pass", int64(0), false)
+			},
+			ExpectedStdoutOutputs: []string{"Sandbox Created"},
+			ExpectedAsserts: func(t *testing.T, ctx context.Context, cm *shared.ClientsMock) {
+				cm.API.AssertCalled(t, "CreateSandbox", mock.Anything, "xoxb-test-token", "-Hello_World 123-", "hello-world-123", "pass", "", "", 0, "", int64(0), false)
+			},
+		},
+		"empty": {
+			CmdArgs: []string{
+				"--experiment=sandboxes",
+				"--token", "xoxb-test-token",
+				"--name", "",
+				"--password", "pass",
+			},
+			Setup: func(t *testing.T, ctx context.Context, cm *shared.ClientsMock, cf *shared.ClientFactory) {
+				setupCreateAuthOnly(t, ctx, cm)
+				cm.IO.On("InputPrompt", mock.Anything, "Enter a name for the sandbox", mock.Anything).Return("", nil)
+			},
+			ExpectedErrorStrings: []string{"Provide a valid domain name"},
+			ExpectedAsserts: func(t *testing.T, ctx context.Context, cm *shared.ClientsMock) {
+				cm.API.AssertNotCalled(t, "CreateSandbox", mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything)
+			},
+		},
+	}, func(cf *shared.ClientFactory) *cobra.Command {
+		return NewCreateCommand(cf)
+	})
 }


cmd/sandbox/create.go
```diff
@@ -0,0 +1,211 @@
...
+	clients.IO.PrintInfo(ctx, false, "\n%s", style.Sectionf(style.TextSection{
+		Emoji: "beach_with_umbrella",
+		Text:  " Sandbox Created",
+		Secondary: []string{
+			fmt.Sprintf("Team ID: %s", teamID),
+			fmt.Sprintf("URL: %s", url),
+		},
+	}))
+	clients.IO.PrintInfo(ctx, false, "Manage this sandbox from the CLI or visit\n%s", style.Secondary("https://api.slack.com/developer-program/sandboxes"))
+}
```
vegeris, line 278: Preview:
```
% hermes sandbox create --experiment=sandboxes --name "EV test box 7245"  --password "jhsdjdkhfkdfgkgd"

🏖️  Sandbox Created
   Team ID: E0197Q6CTV1
   URL: https://ev-test-box-7245.enterprise.dev.slack.com/

Manage this sandbox from the CLI or visit
https://api.slack.com/developer-program/sandboxes
```

cmd/sandbox/delete.go
```diff
@@ -0,0 +1,115 @@
...
+		},
+	}))
+
+	err = printSandboxes(cmd, clients, auth.Token, auth)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
```
vegeris, line 124: Preview:
```
% hermes sandbox delete --experiment=sandboxes --sandbox-id E0196ATAVGT
Choose a Slack team where your email address matches your Slack developer account
? Select a team for authentication slackforceorg E014LMDF01H

⚠️  Danger zone
   Sandbox (E0196ATAVGT) and all of its data will be permanently deleted
   This cannot be undone

? Are you sure you want to delete the sandbox? Yes

✅ Sandbox deleted
   Sandbox E0196ATAVGT has been permanently deleted

🏖️  Developer Sandboxes
   Owned by Slack developer account ev@mail.com

  EV test box 8274368 (E0123456)
    URL: https://ev-test-box-29378.slack.com
    Status: ACTIVE
    Created: 2026-03-12
    Expires: 2026-09-12

Learn more at https://docs.slack.dev/tools/developer-sandboxes

```

cmd/sandbox/create.go
```diff
@@ -0,0 +1,211 @@
...
+
+func NewCreateCommand(clients *shared.ClientFactory) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "create [flags]",
+		Short: "Create a developer sandbox",
+		Long:  `Create a new Slack developer sandbox`,
+		Example: style.ExampleCommandsf([]style.ExampleCommand{
+			{Command: "sandbox create --name test-box --password mypass", Meaning: "Create a sandbox named test-box"},
+			{Command: "sandbox create --name test-box --password mypass --domain test-box --archive-ttl 1d", Meaning: "Create a temporary sandbox that will be archived in 1 day"},
+			{Command: "sandbox create --name test-box --password mypass --domain test-box --archive-date 2025-12-31", Meaning: "Create a sandbox that will be archived on a specific date"},
```
vegeris, line 60: I'm wondering if it makes sense to support both of these ways of setting archive date; alternatively we could just accept the date. We currently archive sandboxes via a nightly cron job so even if you create a sandbox with a time to live of a few hours, it won't get archived until that job runs

internal/api/sandbox.go
zimeg, line 1: 🧪 question: Do we have adjacent unit test or two for these API methods?

cmd/sandbox/sandbox.go
```diff
@@ -75,6 +77,7 @@ func getSandboxAuth(ctx context.Context, clients *shared.ClientFactory) (*types.
 	}
 
 	// Prompt the user to select a team to use for authentication
+	clients.IO.PrintInfo(ctx, false, "%s", style.Secondary("Choose a Slack team where your email address matches your Slack developer account"))
```
zimeg, line 81: 💌 praise: Super helpful comment for the unfamiliar explorers I think! Thanks for adding this and team prompts also!

🔭 thought: We might have adjacent experimental changes that we can use to inline this as "help" text within the prompt - https://github.com/slackapi/slack-cli/pull/400. It might be nice to revisit this output once that becomes stable?


```suggestion
	// TODO(experiment:charm): Change this to prompt "help" message once charm is stable
	clients.IO.PrintInfo(ctx, false, "%s", style.Secondary("Choose a Slack team where your email address matches your Slack developer account"))
```

internal/shared/types/sandbox.go
```diff
@@ -14,12 +14,12 @@
 
 package types
 
-// Sandbox represents a Slack Developer Sandbox from the developer.sandbox.list API.
+// Sandbox represents a Slack Developer Sandbox
 type Sandbox struct {
-	DateArchived  int64  `json:"date_archived"`   // When the developer sandbox is or will be archived, as epoch seconds
-	DateCreated   int64  `json:"date_created"`    // When the developer sandbox was created, as epoch seconds
```
zimeg, line 20: 🌟 praise: To kind epoch!

cmd/sandbox/create.go
```diff
@@ -0,0 +1,211 @@
...
+			}
+		}
+	}
+	// Trim leading/trailing hyphens
+	for len(b) > 0 && b[0] == '-' {
+		b = b[1:]
+	}
+	for len(b) > 0 && b[len(b)-1] == '-' {
+		b = b[:len(b)-1]
+	}
```
zimeg: 🐭 suggestion: We can perhaps use [`strings.Trim`](https://pkg.go.dev/strings#Trim) here instead?

cmd/sandbox/create.go
```diff
@@ -0,0 +1,211 @@
...
+}
+
+// domainFromName derives domain-safe text from the name of the sandbox (lowercase, alphanumeric + hyphens).
+func domainFromName(name string) string {
+	var b []byte
+	for _, r := range name {
+		if (r >= 'a' && r <= 'z') || (r >= '0' && r <= '9') {
+			b = append(b, byte(r))
+		} else if r >= 'A' && r <= 'Z' {
+			b = append(b, byte(r+32))
```
zimeg: 🔍 suggestion: I forget what "string" characters are available in a loop, but would the [`strings.ToLower`](https://pkg.go.dev/strings#ToLower) or the [`unicode.ToLower`](https://pkg.go.dev/unicode#ToLower) function be more clear here?

cmd/sandbox/create.go
```diff
@@ -0,0 +1,211 @@
...
+		} else if r == ' ' || r == '-' || r == '_' {
+			if len(b) > 0 && b[len(b)-1] != '-' {
+				b = append(b, '-')
+			}
+		}
```
zimeg: 🧵 suggestion: The [`strings.ReplaceAll`](https://pkg.go.dev/strings#ReplaceAll) method might be similar before iterations here?

cmd/sandbox/create.go
```diff
@@ -0,0 +1,211 @@
...
+		return "sandbox"
+	}
+	return string(b)
+}
+
+func printCreateSuccess(cmd *cobra.Command, clients *shared.ClientFactory, teamID, url string) {
+	ctx := cmd.Context()
+	clients.IO.PrintInfo(ctx, false, "\n%s", style.Sectionf(style.TextSection{
+		Emoji: "beach_with_umbrella",
+		Text:  " Sandbox Created",
```
zimeg:
```suggestion
		Text:  "Sandbox Created",
```

🔍 question: Would this be alright removed to keep alignment in outputs?

cmd/sandbox/delete.go
```diff
@@ -0,0 +1,114 @@
...
+		}
+	}
+
+	if err := clients.API().DeleteSandbox(ctx, auth.Token, deleteCmdFlags.sandboxID); err != nil {
+		return err
+	}
+
+	clients.IO.PrintInfo(ctx, false, "\n%s", style.Sectionf(style.TextSection{
+		Emoji: "white_check_mark",
+		Text:  "Sandbox deleted",
```
zimeg:
```suggestion
		Text:  "Sandbox Deleted",
```

🧮 suggestion: Title casings for section headers!

cmd/sandbox/delete.go
zimeg: 🫧 thought: This is solid! I'm optimistic that a conclusion of the `charm` experiment might let us hide selections and warnings earlier for outputs that are focused on the result, but nothing to think about in this PR!

cmd/sandbox/create.go
```diff
@@ -0,0 +1,211 @@
...
+
+	// If one's developer account is managed by multiple Production Slack teams, one of those team IDs must be provided in the command
+	cmd.Flags().StringVar(&createCmdFlags.owningOrgID, "owning-org-id", "", "Enterprise team ID that manages your developer account, if applicable")
+
+	if err := cmd.MarkFlagRequired("name"); err != nil {
+		panic(err)
+	}
+	if err := cmd.MarkFlagRequired("password"); err != nil {
+		panic(err)
+	}
```
zimeg: 🏁 issue(non-blocking): We might want prompt alternatives for required flags but no blocker for this PR!

cmd/sandbox/create.go
```diff
@@ -0,0 +1,211 @@
...
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			return requireSandboxExperiment(clients)
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runCreateCommand(cmd, clients)
+		},
+	}
+
+	cmd.Flags().StringVar(&createCmdFlags.name, "name", "", "Organization name for the new sandbox")
+	cmd.Flags().StringVar(&createCmdFlags.domain, "domain", "", "Team domain (e.g., pizzaknifefight). If not provided, derived from org name")
```
zimeg: 🍕 praise: Toward common workplace shenanigan IIRC!

cmd/sandbox/create_test.go
```diff
@@ -0,0 +1,317 @@
...
+			},
+		},
+	}, func(cf *shared.ClientFactory) *cobra.Command {
+		return NewCreateCommand(cf)
+	})
+}
+
+func Test_getEpochFromTTL(t *testing.T) {
+	tests := []struct {
+		name    string
```
zimeg:
```suggestion
	tests := map[string]struct {
```

🧪 quibble: Let's use table tests here! I was confused in thinking "24h" was both the expected case and actual result at a glance...

zimeg: 👾 ramble: And perhaps a bounds of expected archive date might be most confident in testing? Like, some "24h" is greater than now but less than "48h" from now?

cmd/sandbox/create.go
mwbrooks: two-cents: I actually like both flags `--archive-date <YYYY-MM-DD>` and `--archive-ttl <str>`. In CI/CD environments, I think the `--archive-ttl` will be easier to use while `--archive-date` is more preferred for a human.

If the Slack API supports both, I think we should support both. We'll just need to make sure that developers can't provide both together and document the various expressions supported by TTL.

cmd/sandbox/create.go
```diff
@@ -0,0 +1,211 @@
...
+	cmd.Flags().StringVar(&createCmdFlags.password, "password", "", "Password used to log into the sandbox")
+	cmd.Flags().StringVar(&createCmdFlags.locale, "locale", "", "Locale (eg. en-us, languageCode-countryCode)")
+	cmd.Flags().StringVar(&createCmdFlags.template, "template", "", "Template ID for pre-defined data to preload")
+	cmd.Flags().StringVar(&createCmdFlags.eventCode, "event-code", "", "Event code for the sandbox")
+	cmd.Flags().StringVar(&createCmdFlags.archiveTTL, "archive-ttl", "", "Time-to-live duration; sandbox will be archived at end of day after this period (e.g., 2h, 1d, 7d)")
```
mwbrooks: suggestion: In the Long Description can be list all of the available formats for TTL? `1h, 1d, 1w?, 1m? 1y?, etc`

cmd/sandbox/create.go
```diff
@@ -0,0 +1,211 @@
...
+	cmd.Flags().StringVar(&createCmdFlags.archiveDate, "archive-date", "", "Explicit archive date in yyyy-mm-dd format. Cannot be used with --archive")
```
mwbrooks: question: Should this be `Cannot be used with --archive-ttl`?
```suggestion
	cmd.Flags().StringVar(&createCmdFlags.archiveDate, "archive-date", "", "Explicit archive date in yyyy-mm-dd format. Cannot be used with --archive-ttl")
```

cmd/sandbox/create.go
```diff
@@ -0,0 +1,211 @@
...
+
+	// If one's developer account is managed by multiple Production Slack teams, one of those team IDs must be provided in the command
+	cmd.Flags().StringVar(&createCmdFlags.owningOrgID, "owning-org-id", "", "Enterprise team ID that manages your developer account, if applicable")
+
+	if err := cmd.MarkFlagRequired("name"); err != nil {
+		panic(err)
```
mwbrooks: question: Have we checked what happens when we throw a `panic(err)` here? I don't want developers to see some Golang stack trace. 

I don't think we've used the `.MarkFlagRequired` anywhere before.

cmd/sandbox/create.go
mwbrooks: praise: Excellent formatting choice!

cmd/sandbox/create_test.go
mwbrooks: Agreed, we have a standard table test format now and all of our other tests have been updated to use it. Let's use it here too.

cmd/sandbox/delete.go
```diff
@@ -0,0 +1,114 @@
...
+	force     bool
+}
+
+var deleteCmdFlags deleteFlags
+
+func NewDeleteCommand(clients *shared.ClientFactory) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "delete [flags]",
+		Short: "Delete a developer sandbox",
+		Long:  `Permanently delete a sandbox and all of its data`,
```
mwbrooks, line 38: question: For my own reference, delete is different from archive? Or, is delete archiving it?

cmd/sandbox/delete.go
```diff
@@ -0,0 +1,114 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runDeleteCommand(cmd, clients)
+		},
+	}
+
+	cmd.Flags().StringVar(&deleteCmdFlags.sandboxID, "sandbox-id", "", "Sandbox team ID to delete")
+	cmd.Flags().BoolVar(&deleteCmdFlags.force, "force", false, "Skip confirmation prompt")
+
+	if err := cmd.MarkFlagRequired("sandbox-id"); err != nil {
+		panic(err)
```
mwbrooks: suggestion: Again, let's just check that `panic(err)` is handled gracefully before we ship this.

cmd/sandbox/delete.go
```diff
@@ -0,0 +1,114 @@
...
+func runDeleteCommand(cmd *cobra.Command, clients *shared.ClientFactory) error {
+	ctx := cmd.Context()
+
+	auth, err := getSandboxAuth(ctx, clients)
+	if err != nil {
+		return err
+	}
+
+	skipConfirm := deleteCmdFlags.force
+	if !skipConfirm {
```
mwbrooks, line 80: praise: Thanks for supporting `--force`!

cmd/sandbox/delete.go
mwbrooks: praise: Excellent formatting choice!

cmd/sandbox/sandbox.go
```diff
@@ -49,6 +49,8 @@ New to the Developer Program? Sign up at
 	}
 
 	cmd.AddCommand(NewListCommand(clients))
+	cmd.AddCommand(NewCreateCommand(clients))
+	cmd.AddCommand(NewDeleteCommand(clients))
```
mwbrooks: suggestion: Please alphabetize these commands
```suggestion
	cmd.AddCommand(NewCreateCommand(clients))
	cmd.AddCommand(NewDeleteCommand(clients))
	cmd.AddCommand(NewListCommand(clients))
```

internal/api/api_mock.go
```diff
@@ -220,6 +220,16 @@ func (m *APIMock) ListSandboxes(ctx context.Context, token string, filter string
...
+func (m *APIMock) CreateSandbox(ctx context.Context, token, name, domain, password, locale, owningOrgID, template, eventCode string, archiveDate int64) (string, string, error) {
+	args := m.Called(ctx, token, name, domain, password, locale, owningOrgID, template, eventCode, archiveDate)
+	return args.String(0), args.String(1), args.Error(2)
+}
+
+func (m *APIMock) DeleteSandbox(ctx context.Context, token, sandboxID string) error {
+	args := m.Called(ctx, token, sandboxID)
+	return args.Error(0)
+}
+
```
mwbrooks: suggestion: We'd really appreciate it if you can alphabetize these to be declared above the `ListSandboxes` function. 🙏🏻

internal/api/sandbox.go
```diff
@@ -17,18 +17,25 @@ package api
...
 	"github.com/slackapi/slack-cli/internal/goutils"
 	"github.com/slackapi/slack-cli/internal/shared/types"
 	"github.com/slackapi/slack-cli/internal/slackerror"
 )
 
-const sandboxListMethod = "developer.sandbox.list"
+const (
+	sandboxListMethod   = "developer.sandbox.list"
+	sandboxCreateMethod = "enterprise.signup.createDevOrg"
+	sandboxDeleteMethod = "developer.sandbox.delete"
```
mwbrooks: nit: Alphabetical please
```suggestion
	sandboxCreateMethod = "enterprise.signup.createDevOrg"
	sandboxDeleteMethod = "developer.sandbox.delete"
	sandboxListMethod   = "developer.sandbox.list"
```

internal/api/sandbox.go
```diff
@@ -17,18 +17,25 @@ package api
...
+)
 
 // SandboxClient is the interface for sandbox-related API calls
 type SandboxClient interface {
 	ListSandboxes(ctx context.Context, token string, filter string) ([]types.Sandbox, error)
+	CreateSandbox(ctx context.Context, token, name, domain, password, locale, owningOrgID, template, eventCode string, archiveDate int64) (teamID, sandboxURL string, err error)
+	DeleteSandbox(ctx context.Context, token, sandboxID string) error
```
mwbrooks: nit: Alphabetical please
```suggestion
	CreateSandbox(ctx context.Context, token, name, domain, password, locale, owningOrgID, template, eventCode string, archiveDate int64) (teamID, sandboxURL string, err error)
	DeleteSandbox(ctx context.Context, token, sandboxID string) error
	ListSandboxes(ctx context.Context, token string, filter string) ([]types.Sandbox, error)
```

internal/api/sandbox.go
```diff
@@ -69,3 +76,83 @@ func (c *Client) ListSandboxes(ctx context.Context, token string, status string)
 
 	return resp.Sandboxes, nil
 }
+
+type createSandboxResponse struct {
+	extendedBaseResponse
+	TeamID string `json:"team_id"`
+	UserID string `json:"user_id"`
+	URL    string `json:"url"`
+}
```
mwbrooks, line 46: nit: Please declare this at the top of the file.

internal/api/sandbox.go
```diff
@@ -69,3 +76,83 @@ func (c *Client) ListSandboxes(ctx context.Context, token string, status string)
...
+
+// CreateSandbox creates a new developer sandbox
+func (c *Client) CreateSandbox(ctx context.Context, token, name, domain, password, locale, owningOrgID, template, eventCode string, archiveDate int64) (teamID, sandboxURL string, err error) {
```
mwbrooks: nit: Please put the `CreateSandbox` and `DeleteSandbox` above the `ListSandbox`.

cmd/sandbox/create.go
vegeris: Sure, we can keep both for now 🙂 One clarification: the Slack API accepts one archive param which is a datetime in epoch seconds, so when either of these flags is used, the given value is being converted into an epoch timestamp for that value

cmd/sandbox/sandbox.go
vegeris: Nice; I was trying to find a 'subtitle' option for displaying that text but it wasn't currently supported; I'll add that reminder comment :)

cmd/sandbox/create.go
vegeris: Whoops; that makes more sense!

cmd/sandbox/create.go
vegeris: 💯

cmd/sandbox/create.go
mwbrooks: Sounds good, I think both offer a lot of value. After reading over the PR, I also noticed that the API only accepted the one and it's a CLI syntactic-sugar 🍯 that adds these flags. I think this is a great example of where the CLI can add additional value.

internal/api/sandbox.go
vegeris: Added!

cmd/sandbox/create.go
vegeris: In my terminal, I see no spacing between the emoji and the text if I don't add it to this string, although if I copy paste that text somewhere, there actually is one :notsureif:

```
🏖️ Sandbox Created
```
Screenshot 2026-03-19 at 10 44 51 AM

cmd/sandbox/create.go
```diff
@@ -0,0 +1,240 @@
...
+	name        string
+	domain      string
+	password    string
+	locale      string
+	owningOrgID string
+	template    string
+	eventCode   string
+	archiveTTL  string // TTL duration, e.g. 1d, 2h
+	archiveDate string // explicit date yyyy-mm-dd
+	partner     bool
```
vegeris, line 41: Decided to include support for the `partner` flag, since we have validation on the backend to make sure only partner developers can create partner sandboxes

cmd/sandbox/create.go
```diff
@@ -0,0 +1,240 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runCreateCommand(cmd, clients)
+		},
+	}
+
+	cmd.Flags().StringVar(&createCmdFlags.name, "name", "", "Organization name for the new sandbox")
+	cmd.Flags().StringVar(&createCmdFlags.domain, "domain", "", "Team domain. If not provided, will be derived from org name")
+	cmd.Flags().StringVar(&createCmdFlags.password, "password", "", "Password used to log into the sandbox")
+	cmd.Flags().StringVar(&createCmdFlags.locale, "locale", "", "Locale (eg. en-us, languageCode-countryCode)")
+	cmd.Flags().StringVar(&createCmdFlags.template, "template", "", "Template for pre-defined data to preload (default, empty)")
```
vegeris: I updated the handling of the `template` flag; the idea is that users will provide one of the hardcoded template names such as 'default' or 'empty' (the only templates currently supported), but we also quietly accept an integer value. The API accepts an integer value, so we need to map the user-friendly template name to the integer ID.

cmd/sandbox/create.go
zimeg: @vegeris Ahha I wasn't sure if that was a typo! Let's add this case to our known spacings I think?

https://github.com/slackapi/slack-cli/blob/7e4a60f6242c829601e0f76677e228d3fdcb16fe/internal/style/style.go#L87-L122

cmd/sandbox/create.go
vegeris: Good point; I also updated the parsing of archiveTTL to only recognize day/week/month; no 'hour' option (though we'll likely look at adding support for these shorter archive times in the future)

cmd/sandbox/create.go
vegeris: Yeah, it'll output:
```
% hermes sandboxes create  --experiment=sandboxes
Check /Users/evegeris/.slack/logs/slack-debug-20260319.log for error logs

🚫 required flag(s) "name", "password" not set
```

Screenshot 2026-03-19 at 1 28 45 PM

cmd/sandbox/delete.go
vegeris: Yeah, different states; when a sandbox is created, by default it has a future archive date set 6 months in the future. The user can keep extending the archive date, but if it isn't extended, it will be archived on that date. When archived, users are deactivated and apps uninstalled, I think. After a grace period of 90 days during which the developer can unarchive it to continue using it, it will then be deleted. It cannot be recovered after deletion

vegeris: We may want to add CLI support for extending the archive date in the future

cmd/sandbox/delete.go
vegeris: I would consider it pretty graceful: https://github.com/slackapi/slack-cli/pull/389#discussion_r2961602711

internal/api/api_mock.go
vegeris: Will do!

cmd/sandbox/create.go
vegeris: Added!

cmd/sandbox/create.go
vegeris: I changed this line to just `Team domain. If not provided, will be derived from org name`, thinking we don't need to give an example of a domain. If it makes the messaging more fun / Slack-y I can bring it back!

cmd/sandbox/create.go
zimeg: @vegeris Do we have documentation on these mappings to include in comment? I'd be quick to think these are CLI specific mappings I fear 👻 

Edit: Perhaps "reference" or a link otherwise more than "documentation" forgive me!

vegeris: These would indeed be mappings specific to the CLI; it would be listed from the `--help` flag and when publicly released (from behind the experiment flag) it could be noted in the docs, but the backend only recognizes the integer template IDs:
```
--template string        Template for pre-defined data to preload (default, empty)
```

If this doesn't seem user-friendly, we could just accept the ints; I defer to you guys on the user experience 😛

zimeg: @vegeris Ah I perhaps meant documentation more for us maintainers! I don't know what template ID `3` might be... 🔭

cmd/sandbox/create.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+
+	maxAllowed := now.AddDate(0, 6, 0)
+	minAllowed := now.AddDate(0, 0, 1)
+	if target.Before(minAllowed) {
+		return 0, slackerror.New(slackerror.ErrInvalidArchiveTTL)
+	}
+	if target.After(maxAllowed) {
+		return 0, slackerror.New(slackerror.ErrInvalidArchiveTTL)
+	}
```
vegeris, line 207: 🏁 I added some basic validation around acceptable archive dates but I'm not sure if we want to keep this in the CI; the backend will error out if an invalid date is provided but the error message isn't particularly user friendly (it's the generic 'invalid_args' response). This at least highlights which arg is invalid

vegeris: Actually, I think I'll just include validation for 'date cannot be in the past' for the `--archive-date` flag and that's it. It's not possible to provide a date in the past for `--archive-ttl` anyway. This just leaves out any extra validation for max archive date (backend will return invalid_args)

cmd/sandbox/create.go
vegeris: Maybe we [want](https://github.com/slackapi/slack-cli/pull/389#discussion_r2942232556) prompts instead though

cmd/sandbox/create.go
vegeris: I went ahead and added those prompts 👀

cmd/sandbox/create.go
vegeris: I added the prompts so we're no longer using `MarkFlagRequired` here

cmd/sandbox/delete.go
vegeris: We could prompt them instead of erroring though; that would be consistent with the `create` command now

vegeris: I updated this to a prompt instead

cmd/sandbox/create_test.go
vegeris: Updated!

PR #392, opened 2026-03-12
test: increase test coverage in cmd/ and internal/[api | config | deputil]

mwbrooks:
### Changelog

- N/A

### Summary

This pull request increases our test coverage +0.8% by adding tests to untested functions in:

- `cmd/`
- `internal/api`
- `internal/config`
- `internal/deputil`

### Requirements

### DIFF
--- a/internal/config/flags_test.go
+++ b/internal/config/flags_test.go
@@ -46,20 +46,67 @@ func Test_InitializeGlobalFlags(t *testing.T) {
 
 	config.InitializeGlobalFlags(cmd)
 
-	// Verify that key persistent flags were registered
-	flagNames := []string{
-		"apihost", "app", "config-dir", "experiment",
-		"force", "no-color", "skip-update", "slackdev",
-		"runtime", "team", "token", "verbose",
+	tests := map[string]struct {
+		longform  string
+		shorthand string
+		hidden    bool
+	}{
+		"apihost": {
+			longform: "apihost",
+			hidden:   true,
+		},
+		"app": {
+			longform:  "app",
+			shorthand: "a",
+		},
+		"config-dir": {
+			longform: "config-dir",
+		},
+		"experiment": {
+			longform: "experiment",
+		},
+		"force": {
+			longform:  "force",
+			shorthand: "f",
+		},
+		"no-color": {
+			longform: "no-color",
+		},
+		"runtime": {
+			longform:  "runtime",
+			shorthand: "r",
+			hidden:    true,
+		},
+		"skip-update": {
+			longform:  "skip-update",
+			shorthand: "s",
+		},
+		"slackdev": {
+			longform: "slackdev",
+			hidden:   true,
+		},
+		"team": {
+			longform:  "team",
+			shorthand: "w",
+		},
+		"token": {
+			longform: "token",
+		},
+		"verbose": {
+			longform:  "verbose",
+			shorthand: "v",
+		},
 	}
-	for _, name := range flagNames {
-		f := cmd.PersistentFlags().Lookup(name)
-		assert.NotNil(t, f, "flag %s should be registered", name)
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			f := cmd.PersistentFlags().Lookup(tc.longform)
+			assert.NotNil(t, f, "flag %s should be registered", tc.longform)
+			if tc.shorthand != "" {
+				assert.Equal(t, tc.shorthand, f.Shorthand, "flag %s shorthand mismatch", tc.longform)
+			}
+			assert.Equal(t, tc.hidden, f.Hidden, "flag %s hidden mismatch", tc.longform)
+		})
 	}
-
-	// Verify hidden flags
-	assert.True(t, cmd.PersistentFlags().Lookup("apihost").Hidden)
-	assert.True(t, cmd.PersistentFlags().Lookup("slackdev").Hidden)
 }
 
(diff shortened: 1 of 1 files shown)


internal/config/flags_test.go
```diff
@@ -24,6 +24,44 @@ import (
...
+		"runtime", "team", "token", "verbose",
+	}
+	for _, name := range flagNames {
+		f := cmd.PersistentFlags().Lookup(name)
+		assert.NotNil(t, f, "flag %s should be registered", name)
+	}
+
+	// Verify hidden flags
+	assert.True(t, cmd.PersistentFlags().Lookup("apihost").Hidden)
+	assert.True(t, cmd.PersistentFlags().Lookup("slackdev").Hidden)
```
mwbrooks: note: I went back-and-forth on whether to have this test, but it's nice to have a test that locks-in and confirms that our expected global flags and hidden flags haven't changed without intention.

cmd/datastore/query_test.go
```diff
@@ -465,3 +465,55 @@ func prepareExportMockData(cm *shared.ClientsMock, numberOfItems int, maxItemsTo
...
+			result, err := mapAttributeFlag(tc.flag)
+			if tc.wantErr {
+				assert.Error(t, err)
+			} else {
+				assert.NoError(t, err)
+				assert.NotNil(t, result)
+			}
+		})
+	}
+}
```
zimeg, line 519: 💡 thought: Unit tests matching each function is a kind pattern since IMHO it also encourages tests at the `cmd` level is preferred?

internal/api/app_test.go
```diff
@@ -519,3 +519,77 @@ func TestClient_DeveloperAppInstall_RequestAppApproval(t *testing.T) {
...
+				require.Contains(t, err.Error(), tc.errMessage)
+			} else {
+				require.NoError(t, err)
+				require.NotNil(t, result)
+			}
+		})
+	}
+}
+
+func Test_Client_ConnectionsOpen(t *testing.T) {
```
zimeg: 🧮 quibble(non-blocking): Forgive me for not searching too much up, but if tests have alphabetical order in this file?

👾 ramble: I'd be curious if separating API methods into different files gives us the most confidence toward focused tests. Not something to change now of course, but it'd be curious pattern for ongoing additions!

internal/api/app_test.go
zimeg, line 1: ⭐ praise: More praises for API tests! These are so good!

internal/config/flags_test.go
```diff
@@ -24,6 +24,44 @@ import (
...
+	cmd := &cobra.Command{}
+
+	config.InitializeGlobalFlags(cmd)
+
+	// Verify that key persistent flags were registered
+	flagNames := []string{
+		"apihost", "app", "config-dir", "experiment",
+		"force", "no-color", "skip-update", "slackdev",
+		"runtime", "team", "token", "verbose",
+	}
```
zimeg: 🧮 quibble: Separating these to newlines might be nice for fast reading. I assumed this was a map at first! Perhaps this can be a table test?

```go
tests := map[string]struct{
    shorthand string
    longform  string
}{
    "apihost": {
        longform: "apihost",
    },
    "app": {
        longform:  "app",
        shortform: "a",
    },
}
```

IIRC? Forgive me if `shortform` is too much. I'm hopeful that we'll soon revisit some of these options like `app` and `team` too!

internal/deputil/url_test.go
zimeg, line 1: 📠 thought: A `slackhttp` package might be easier to find this ongoing? TIL about these modules too-

cmd/datastore/query_test.go
mwbrooks: I agree, I think we could actually reduce our overall tests and have the same coverage by testing through the command level. We've had success with some of the simpler, newer commands and hopefully we can bring that to the original, complex commands!

internal/api/app_test.go
mwbrooks: Unfortunately, it's not alphabetized. The file looks like it started off being alphabetical, but after line 200, it's a free-for-all. Commit e2e0ed50452feed54c2f984b8c7bc21dda103e4d attempts to move these functions where they should be.

A challenge with alphabetizing is that logically we want to group getters/setters but alphabetically they are separated. Something for us to think about.

mwbrooks: > 👾 ramble: I'd be curious if separating API methods into different files gives us the most confidence toward focused tests. Not something to change now of course, but it'd be curious pattern for ongoing additions!

Personally, I prefer small files would be open to a single file for each API method. But Golang tends to prefer fewer, longer files. That's probably why we are taking the current approach.

internal/config/flags_test.go
mwbrooks: @zimeg Great suggestion, I think a list with explicit variables for short and hidden make these tests more readable. Commit c4dbb36198510be2aecf6094bd00f19509984cfc

internal/deputil/url_test.go
mwbrooks: 🧠 Makes sense that we can merge these two eventually! I'll add a note.

mwbrooks:
@zimeg Thanks for the thorough review! I'll try to keep things alphabetized and maybe we can find a common standard for how to order getter/settings (e.g. Getters don't include "Get" and Setter always follows the Getter → `AppName()` and `SetAppName()`). 🚀 rockets away, onto the next PR!

PR #395, opened 2026-03-12
feat: default manifest source to manifest.json for Bolt Framework projects

mwbrooks:
### Changelog

> Bolt JavaScript and Bolt Python projects now default to using the local `manifest.json` as the app manifest.
> - Previously, it was a remote manifest from App Settings (https://api.slack.com/apps).
> - Changes to the app manifest on App Settings will be detected by the Slack CLI and a prompt will be displayed before any manifest update action (e.g. `slack run`).
> - The manifest source is set after creating a new project with `slack create` or initializing an existing project with `slack init` and the setting is stored as `"manifest.source": "local"` in `.slack/config.json`.

### Summary

This pull request defaults Bolt Framework projects to use a `manifest.source: "local"` when creating and initializing a project.

**Reason behind the change:** We're making this change after receiving feedback from multiple developers who are confused by the `remote` setting and would prefer `local` for Bolt projects. The rationale is that developers using the CLI would prefer to default to managing the manifest with the CLI/Project.

Related #396 that will update `app link` to not change the manifest source.

### Open Question

- Is this considered `feat: enhancement` or `fix: bug fix`? 

### Preview

https://github.com/user-attachments/assets/2cea0cc1-0f4c-4e56-afa5-29e3d0f9b3b3

### Test Steps

**Test Creating a Bolt Framework Project:**

```bash
# Create a new Bolt Project
$ lack create my-app -t https://github.com/slack-samples/bolt-js-starter-template
$ cd my-app/

# Confirm local manifest
$ cat .slack/config.json
# → Confirm manifest.source: "local"

# Create and run the app
$ lack run
# → Press CTRL+C to stop the server

# Change Something on App Setting to Introduce a Conflict
$ lack app settings
# → Change App Name
# → Set Description and Background Color
# → Save

# Test Manifest Change Warning
$ lack run
# → Confirm warning that manifest is out-of-sync
# → Select Y

# Clean up
$ lack delete -f
$ cd ..
$ rm -rf my-app
```

**Test Initializing a Bolt Framework Project:**

```bash
# Clone a Bolt Framework Project
$ git clone https://github.com/slack-samples/bolt-js-starter-template.git
$ cd bolt-js-starter-template/

# Initialize it
$ lack init

# Confirm local manifest
$ cat .slack/config.json
# → Confirm manifest.source: "local"

# Clean up
$ cd ..
$ rm -rf bolt-js-starter-template/
```

**Test a Deno Project:**

```bash
# Create a new Deno Project
[...]

PR #396, opened 2026-03-13
feat: update 'app link' to not change the app manifest source

mwbrooks:
### Changelog

> Updated `slack app link` command to support project's that use a manifest file as the manifest source (`manifest.json`).
> - Does not display a warning message when the manifest source is local (project file).
> - Does not change the manifest source - now uses the project's current manifest source.
> - Displays the current manifest source for the developer's reference.
> - Displays a tip on where to change the manifest source (`.slack/config.json`).
>
> Updated the app manifest overwrite warning prompt wording and appearance.
> - Clarifies that the app settings manifest will be overwritten by the project's app manifest file.
> - Displays a yellow warning that changes were detected on the app settings manifest that are not in the project's app manifest file.

### Summary

This pull request is a little spicy 🌶️ because remove speed bumps and reduces friction for developers to allows apps from App setting to be managed by the Slack CLI.

See **Changelog** section for a detailed breakdown of the changes.

### Motivation

The motivation for these changes is 💬 developer feedback. A consistent theme of questions from developers is how to have the Slack CLI manage and use the `manifest.json` that ships with all of our sample apps (_sensible question_). We often need to explain how to manually update `.slack/config.json → manifest.source: "local"` for developers to get the experience they want.

Our concern for embracing the project's manifest file is that it can overwrite the manifest on App Settings. This could be a shocking and difficult experience, if it happens unexpectedly. Thankfully, the Slack CLI's overwrite protection has been reliable for the past year and this change leans more heavily on it.

#### Linking an App ID from App Setting

Currently, when Linking an App ID from App Setting, the developer is forced to switch the manifest source to be App Setting (remote). If they decline the confirmation prompt, then they cannot link the app. A workaround that developers have taken is to accept the confirmation and then manually change the manifest source back to the manifest file: `.slack/config → manifest.source: "local"`.

Now, the manifest source is not changed - it can remain local (manifest file) or remote (app settings). We rely on the manifest overwrite protection to warn developers when there are changes on App Settings that are not in the manifest file.

<details>
  <summary>Expand to see a video of the current behaviour</summary>

https://github.com/user-attachments/assets/b8183922-ff84-42da-b675-21d71b504c4a

</details>

https://github.com/user-attachments/assets/407133a4-4ac6-4350-aa9d-2345049f2af6

#### Overwriting App Settings Warning

We made minor wording and appearance changes to clarify that the app settings manifest will be overwritten by the manifest file.

<detials>
  <summary>Expand to see a video of the current behaviour</summary>

https://github.com/user-attachments/assets/c5134142-ce21-4503-922a-a215fd48a855

</details>

https://github.com/user-attachments/assets/47b02bc7-acb1-4366-b9f5-2e7b3eda9fb1

### Test Steps

```bash
# Create a Bolt Framework project
$ lack create my-app -t https://github.com/slack-samples/bolt-js-starter-template.git
$ cd my-app/

# Start & Stop the dev server to create an App ID 
$ lack run
# → Create an app
# → Stop the server once it's running
# CTRL+C

# Copy the App ID
$ lack app list
# → Copy the App ID

# Unlink the App ID
$ lack app unlink
# → Select your App ID

# Test: App Link
$ lack app link
# → Confirm: "Manifest source is "project" (local)"
# → Confirm: "Manifest source is configured in .slack/config.json"
# → App ID: Paste the App ID
# → Environment: Local

# Test: Change App Settings
$ lack app settings
# → Add a short description
# → Add a background color
# → Save

# Test: Overwrite Warning
$ lack run
# → Confirm: Overwrite warning
# → Select "Y"
# → CTRL+C after dev server starts

# Test: Confirm No Overwrite Warning when Manifests In-Sync
$ lack run
# → Confirm: No warning and dev server starts
# → CTRL+C

# Clean up
$ lack delete -f
$ cd ..
$ rm -rf my-app/
```

### Requirements


cmd/app/link.go
```diff
@@ -156,67 +150,20 @@ func LinkExistingApp(ctx context.Context, clients *shared.ClientFactory, app *ty
...
-					filepath.Join(config.ProjectConfigDirName, config.ProjectConfigJSONFilename),
-					fmt.Sprintf(`manifest.source: "%s"`, config.ManifestSourceRemote),
-				).
-				WithRootCause(err)
-			clients.IO.PrintError(ctx, "%s", slackErr.Error())
-		}
+	// App Manifest section
+	manifestSource, _ := clients.Config.ProjectConfig.GetManifestSource(ctx)
+	if !manifestSource.Exists() {
+		manifestSource = config.ManifestSourceLocal
```
mwbrooks: note(out-of-scope): We should have the default manifest source return by a common method instead of manually setting it in each spot.

cmd/app/link_test.go
```diff
@@ -431,135 +431,18 @@ func Test_Apps_Link(t *testing.T) {
 			CmdArgs:       []string{},
 			ExpectedError: slackerror.New(slackerror.ErrAppNotFound),
 		},
-		"accepting manifest source prompt should save information about the provided deployed app": {
+		"links app when manifest source is local": {
```
mwbrooks, line 434: note: Test linking an app when the manifest source is local (manifest file).

cmd/app/link_test.go
```diff
@@ -600,42 +483,33 @@ func Test_Apps_Link(t *testing.T) {
...
-					mock.Anything,
-				)
+				// Assert manifest source info is displayed
+				output := cm.GetCombinedOutput()
+				assert.Contains(t, output, "App Manifest")
+				assert.Contains(t, output, `"project" (local)`)
 			},
 		},
-		"manifest source prompt should display for GBP apps with local manifest source": {
+		"links app when manifest source is remote": {
```
mwbrooks, line 501: note: Test linking an app when the manifest source is remote (app settings).

internal/pkg/apps/install.go
```diff
@@ -752,7 +752,7 @@ func shouldUpdateManifest(ctx context.Context, clients *shared.ClientFactory, ap
 	case saved.Equals(""):
 		notice = "Manifest values for this app are overwritten on reinstall"
 	default:
-		notice = "The manifest on app settings has been changed since last update!"
+		notice = style.Yellow("The manifest on app settings has been changed since last update")
```
mwbrooks, line 755: note: I'm open to other formatting (Red?) but I think we should make this pop and stand out in some way.

cmd/app/link.go
zimeg: 🪬 question: I notice that without a `source` attribute in configs we default to a local manifest. Is this something we can favor instead of writing settings to projects?

👾 note: I realize a saved value in each project settings file means less confusion toward defaults, but IMHO we might have paths toward extending the "local" manifest experience adjacent.

https://github.com/slackapi/slack-cli/blob/8e4c0aa11d2814ba516faa97b6c6f35511c0edb1/internal/config/project.go#L147-L171

cmd/app/link.go
```diff
@@ -156,67 +150,20 @@ func LinkExistingApp(ctx context.Context, clients *shared.ClientFactory, app *ty
...
 	}
```
zimeg:
```suggestion
```

🪬 quibble: Related to the comment above, this might be alright to remove?

internal/pkg/apps/install.go
zimeg: 🌟 praise: I like yellow: it matches error codes and isn't confusing with green! But we might continue these discussions in a style guide soonish-

cmd/app/link.go
```diff
@@ -156,67 +150,20 @@ func LinkExistingApp(ctx context.Context, clients *shared.ClientFactory, app *ty
...
+	configPath := filepath.Join(config.ProjectConfigDirName, config.ProjectConfigJSONFilename)
+	clients.IO.PrintInfo(ctx, false, "%s", style.Sectionf(style.TextSection{
+		Emoji: "books",
+		Text:  "App Manifest",
+		Secondary: []string{
+			"Manifest source is " + style.Highlight(manifestSource.Human()),
+			"Manifest source is configured in " + style.Highlight(configPath),
+		},
+	}))
```
zimeg, line 167: 🧭 praise: This is a nice callout to new learnings in this experience! I'm not certain how familiar the local|remote manifest concept is to exploring developers.

cmd/app/link.go
```diff
@@ -156,67 +150,20 @@ func LinkExistingApp(ctx context.Context, clients *shared.ClientFactory, app *ty
...
+	manifestSource, _ := clients.Config.ProjectConfig.GetManifestSource(ctx)
+	if !manifestSource.Exists() {
+		manifestSource = config.ManifestSourceLocal
 	}
+	configPath := filepath.Join(config.ProjectConfigDirName, config.ProjectConfigJSONFilename)
+	clients.IO.PrintInfo(ctx, false, "%s", style.Sectionf(style.TextSection{
+		Emoji: "books",
+		Text:  "App Manifest",
+		Secondary: []string{
+			"Manifest source is " + style.Highlight(manifestSource.Human()),
```
zimeg:
```suggestion
			"Manifest source is gathered from " + style.Highlight(manifestSource.Human()),
```

📸 suggestion(non-blocking): To give less definition to what "project" and "app settings" mean for the manifest source:

```
📚 App Manifest
   Manifest source is gathered from "project" (local)
   Manifest source is configured in .slack/config.json
```

cmd/app/link.go
mwbrooks: Good question. My feeling is that - for now - we should explicitly write to the `config.json` to ensure that a project functions the way it's expected even if the CLI changes it's defaults. 

However, I see where you're coming from. I also know we juggle some challenges with `config.json` in templates (e.g. Project ID). Perhaps your ideas would have address those?

cmd/app/link.go
mwbrooks: Great suggestion! Commit b1b26788c5cb005af3f28e9aa6ed336acfaf39a6 in PR #395 now leverages the defaults provided by the project config getter. I think we can do the same here. I'll look into it now 🔍

mwbrooks: Commit 0f2e6e1659b24a5b754c512de301334f88cc518f removes this check 🎉

cmd/app/link.go
mwbrooks: It crossed my mind as well and we don't have formal documentation that we can point toward. I figured I'd leave it as a future task for now.

cmd/app/link.go
mwbrooks: Thanks for the suggestion. This works for me - personally I'd prefer to use fewer words - but we're dealing with a weird concept (manifest source) so anything that adds clarify is probably a good choice right now! 🙇🏻

PR #398, opened 2026-03-14
feat: migrate experiments config from array to bools

mwbrooks:
### Changelog

- N/A

### Summary

This pull request migrates the experiments config format from an array to a map of boolean values. It includes backwards-compatible JSON unmarshaling that will auto-migrate `config.json` files.

### Example

**Before:**

```json
{
  "experiments": ["sandboxes"],
  "manifest": {
    "source": "local"
  },
  "project_id": "..."
}
```

**After:**

```json
{
  "experiments": {
    "sandboxes": true,
    "charm" :false
  },
  "manifest": {
    "source": "local"
  },
  "project_id": "..."
}
```

### Motivation

**_Why on earth are you doing this?_** 👨🏻‍🎨 🎨 🚲  🏠 

The motivation is to unblock adding a `slack config` command to get and set system-level and project-level config files. 

The config command will use a dot-notation to get and set key values: 

```bash
$ slack config get manifest.source       # Getter
$ slack config set manifest.source local # Setter
```

When the `<value>` is a simple type, it works well, such as string, number, or boolean

``` bash
$ slack config set manifest.source "local"  # String
$ slack config set some.timeout 10000       # Number
$ slack config set some.toggle false        # Boolean
```

When the `<value>` is a complex type, it's not intuitive or simple, such as arrays

```bash
# Which format do we support? It will require regex parsing in most cases
$ slack config set experiments experiment1, experiment2, experiment3
$ slack config set experiments ["experiment1", "experiment2", "experiment3"]
$ slack config set experiments ['experiment1', 'experiment2', 'experiment3']

# Does this append experiment1 to the array or replace the array with experiment1?
$ slack config set experiments experiment1
```

**In order for setting `[value]` to be simple and intuitive, we should restrict the `config.json` files to use simple value types that can be expressed on the command-line (e.g. string, number, and boolean).** Arrays are complex and error-prone to express (`["value1", "value2"]` or `value1, value2`) and all elements must be included each time. 

This is why this pull request migrates the experiments to an object of `key: boolean` pairs. 

An added bonus is that the new format naturally supports disabling experiments from the config file. 🧠 

### Requirements


internal/config/experiments_test.go
```diff
@@ -40,7 +39,7 @@ func Test_Config_WithExperimentOn(t *testing.T) {
 		mockOutput, mockPrintDebug := setupMockPrintDebug()
 
 		// Write our test script to the memory file system used by the mock
-		jsonContents := []byte(fmt.Sprintf("{\"last_update_checked_at\":\"2023-05-11T15:41:07.799619-07:00\",\"experiments\":[\"%s\"]}", validExperiment))
+		jsonContents := []byte(fmt.Sprintf(`{"last_update_checked_at":"2023-05-11T15:41:07.799619-07:00","experiments":{"%s":true}}`, validExperiment))
```
mwbrooks, line 42: note: Just updating the formatting to make the tests more readable with literals and no escapes.

internal/config/experiments_test.go
```diff
@@ -244,7 +253,7 @@ func Test_Config_WithExperimentOn(t *testing.T) {
 		assert.True(t, config.WithExperimentOn(experiment.Placeholder))
 		assert.Contains(t, mockOutput.String(), "active project experiments: []")
 		assert.Contains(t, mockOutput.String(), fmt.Sprintf("active permanently enabled experiments: [%s]", experiment.Placeholder))
-		assert.Equal(t, []experiment.Experiment{experiment.Placeholder}, config.experiments)
+		assert.ElementsMatch(t, []experiment.Experiment{experiment.Placeholder}, config.GetExperiments())
```
mwbrooks, line 268: note: Ordering now varies because it's a map instead of slice.

internal/config/config.go
```diff
@@ -63,7 +62,7 @@ type Config struct {
 
 	// Feature experiments
 	ExperimentsFlag []string
-	experiments     []experiment.Experiment
```
zimeg, line 66: 🪬 question: Can we continue to use strict types of `experiment` or is this not allowed as a key?

```go
	experiments     map[experiment.Experiment]bool
```

internal/config/experiments.go
```diff
@@ -29,47 +31,73 @@ func (c *Config) LoadExperiments(
 	ctx context.Context,
 	printDebug func(ctx context.Context, format string, a ...interface{}),
 ) {
-	experiments := []experiment.Experiment{}
+	experiments := map[string]bool{}
 	// Load from flags
 	for _, flagValue := range c.ExperimentsFlag {
-		experiments = append(experiments, experiment.Experiment(flagValue))
+		experiments[flagValue] = true
 	}
```
zimeg: 👁️‍🗨️ thought: The order of experiment parsing might need to be reversed so that project and user configurations can be overridden with a flag if the lower settings use a "false" value?

internal/config/experiments_test.go
zimeg, line 42: ⭐ praise: Thanks for improving both the string formatting and `experiments` formatting here!

internal/config/experiments_test.go
```diff
@@ -50,13 +49,33 @@ func Test_Config_WithExperimentOn(t *testing.T) {
...
+				validExperiment,
+			),
+		)
+		_ = afero.WriteFile(fs, pathToConfigJSON, jsonContents, 0600)
+
+		config.LoadExperiments(ctx, mockPrintDebug)
+		experimentOn := config.WithExperimentOn(validExperiment)
+		assert.Equal(t, true, experimentOn)
+	})
+
```
zimeg:
```suggestion
```

🪓 quibble(non-blocking): We might not fear breaking experimental configuration setups since this can become burdensome to keep in mind ongoing?

internal/config/config.go
mwbrooks: @zimeg Great suggestion and I should have caught that myself. I think we can use `map[experiment.Experiment]bool`, so let me look into refactoring it 💻

internal/config/experiments_test.go
mwbrooks: thought: Good question. Here's what I've found from a little searching:

- [No documentation exists for configuring experiments with `config.json`.](https://docs.slack.dev/tools/slack-cli/reference/experiments/)
- [No sample apps use experiments.](https://github.com/search?q=org%3Aslack-samples%20experiments&type=code)

So, if a developer has an experiment enabled in the `config.json` then it's at their own risk.

My concern only concern is that removing the backwards compatibility will cause apps with `config.json` experiments to break on any CLI command:

```bash
Failed to parse contents of project-level config file (unable_to_parse_json)
json: cannot unmarshal array into Go struct field ProjectConfig.experiments of type map[string]bool
```

This is a very poor experience, but we can be good hosts by displaying a remediation message:

```bash
💡 Suggestion
   Check that ~/path/to/my-app/.slack/config.json is valid JSON. If the "experiments" field is an array (e.g. ["exp1"]), update it to an object (e.g. {"exp1": true})
```

Commit c500bdeecb8120a04e4c4db07b66e00b4edddb7d does the following:
- Removes `internal/config/experiments_unmarshal.go`
- Removed `UnmarshalJSON` from `internal/config/project.go`
- Removed `UnmarshalJSON` from `internal/config/system.go`
- Displays the remediation hint about "experiments" format when there is a JSON parse error for `config.json`

internal/config/experiments.go
mwbrooks: @zimeg 🙇🏻 Very good catch! Commit f0c8bc3bacf768e4896225516b78cd563405a5ff updates the order to be:

1. Permanently enabled
2. System config
3. Project config
4. Flags

question: Not sure where Permanently enabled should be. I feel we would be better served to rename it to "Enabled by Default" and allow developers to opt-out in case there is a known bug. If we keep it as Permanently enabled, then we may want it to be after Flags to override all other settings.

internal/config/config.go
mwbrooks: Commit a70d3ee5f5bbbfb66c8b26c1a78f286a357095a6 uses the strict types:
```golang
experiments     map[experiment.Experiment]bool
```

A helper method had to be created to copy from the JSON format to the strict type.

internal/config/experiments.go
zimeg: > ...rename it to "Enabled by Default" and allow developers to opt-out in case there is a known bug.

@mwbrooks Thanks! I'm a big fan of this approach! I understand flag experimental flag values to be `true` or not found, but that doesn't take from the configuration files 🚢

PR #401, opened 2026-03-15
revert(experiment): separate create questions with flag parsings

zimeg:
### Changelog

> Experimental changes - N/A 🍀 

### Summary

This PR reverts #362 to separate `create` questions with flag parsing handled with the `clients.IO` package. This hopes to fix an issue with the `--template` flag.

### Notes

- 🧪 Toward conclusion of the `charm` experiment I'm thinking to cover edge cases of the initial `survey` to `huh` package switch before extending specific prompts

### Requirements

- [x] I've read and understood the [Contributing Guidelines](https://github.com/slackapi/slack-cli/blob/main/.github/CONTRIBUTING.md) and have done my best effort to follow them.
- [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).

PR #400, opened 2026-03-15
feat(experiment): inline sample descriptions in plain text

zimeg:
### Changelog

> Nothing for this time 🍀 ✨ 

### Summary

This PR inlines sample descriptions in plain text for the `samples` command and moves docs reference to help text and also removes emojis from descriptions to avoid output issues.

### Preview

<details>
<summary>Before</summary>

before

</details>

After

after

### Requirements

- [x] I've read and understood the [Contributing Guidelines](https://github.com/slackapi/slack-cli/blob/main/.github/CONTRIBUTING.md) and have done my best effort to follow them.
- [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).

PR #87, opened 2025-05-10
feat: Add flags for prompt values to upgrade command

kkemple:
## Summary of changes

- Added a new boolean flag `--auto-approve` to the `upgrade` command
- Added a new method `InstallUpdatesWithoutPrompt` to the `UpdateNotification` type for handling auto-approved updates
- Modified the dependencies' `PrintUpdateNotification` methods to check for the flag and bypass confirmation
- Updated command examples and help text to document the new flag

This PR enables:
- Automated CLI upgrade workflows in CI/CD pipelines
- Programmatic upgrades through the MCP server tools API
- Simplified upgrades for users who always want the latest version
- Better integration with scripts and other automation tools

The flag follows established patterns in the codebase for command options, with appropriate documentation and test coverage to ensure reliability.

---

### Tests Added
- Updated `TestUpgradeCommand` to verify behavior with and without the flag
- Added a test case specifically for the auto-approve flag path
- Ensured all existing tests continue to pass

### Docs Updated
- Added the flag to the command documentation in `docs/reference/commands/slack_upgrade.md`
- Added an example demonstrating usage of the flag
- Updated the options section to include the new flag with its description

---

### Reasoning for implementation
The upgrade command needed a way to be run non-interactively, especially for automation and scripting use cases. The `--auto-approve` flag provides this functionality by skipping the confirmation prompt when installing updates.

While we could have modified the existing confirmation flow directly, creating a dedicated method provides better separation of concerns:
1. Keeps the standard interactive flow clean and focused
2. Makes the auto-approve path more explicit and maintainable
3. Reduces the risk of regression in either flow when changes are made

---

### Requirements

* [x] I've read and understood the [Contributing Guidelines](/blob/main/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).

PR #404, opened 2026-03-16
refactor(experiment): separate iostreams prompts from survey implementations

zimeg:
### Changelog

> N/A - "Nothing 'appening" 👻 

### Summary

This PR separates `iostreams` **prompts** from `survey` implementations for improvements to test cases and planned changes toward standard flag handling.

### Reviewers

No change to logic with this change I hope!

### Requirements

### DIFF
--- a/internal/iostreams/prompts.go
+++ b/internal/iostreams/prompts.go
@@ -144,13 +144,6 @@ type SelectPromptResponse struct {
 	Prompt bool // If a survey selection was made
 }
 
-// DefaultSelectPromptConfig returns default config object for a selection prompt
-func DefaultSelectPromptConfig() SelectPromptConfig {
-	return SelectPromptConfig{
-		Required: true,
-	}
-}
-
 // retrieveFlagValue returns the only changed flag in the flagset
 func (io *IOStreams) retrieveFlagValue(flagset []*pflag.Flag) (*pflag.Flag, error) {
 	var flag *pflag.Flag
--- a/internal/iostreams/prompts_mock.go
+++ b/internal/iostreams/prompts_mock.go
@@ -43,6 +43,13 @@ func MatchPromptConfig(match PromptConfig) interface{} {
 	})
 }
 
+// DefaultSelectPromptConfig returns default config object for a selection prompt
+func DefaultSelectPromptConfig() SelectPromptConfig {
+	return SelectPromptConfig{
+		Required: true,
+	}
+}
+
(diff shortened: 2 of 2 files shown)


cmd/datastore/count_test.go
```diff
@@ -251,7 +251,7 @@ func TestCountCommand(t *testing.T) {
 				}, nil)
 				cm.IO.On("InputPrompt", mock.Anything, "Enter an expression", iostreams.InputPromptConfig{
 					Required: false,
-				}).Return("")
```
zimeg, line 254: 🧪 note: Lots of changes add the expected error value to mock prompts!

internal/iostreams/charm.go
```diff
@@ -83,10 +83,6 @@ func buildSelectForm(msg string, options []string, cfg SelectPromptConfig, selec
 		Options(opts...).
 		Value(selected)
 
-	if cfg.PageSize > 0 {
-		field.Height(cfg.PageSize + 2)
-	}
```
zimeg, line 90: 🪓 note: Alongside changes of #400 and #401 we're hoping to use default form heights while the experiment for `charm` prompts remains active.

internal/iostreams/prompts.go
```diff
@@ -0,0 +1,263 @@
...
+}
+
+// MultiSelectPrompt prompts the user to select multiple values in a list and
+// returns the selected values
+func (io *IOStreams) MultiSelectPrompt(ctx context.Context, message string, options []string) ([]string, error) {
+	if io.config.WithExperimentOn(experiment.Charm) {
+		return charmMultiSelectPrompt(io, ctx, message, options)
+	}
+	return surveyMultiSelectPrompt(io, ctx, message, options)
+}
```
zimeg, line 218: 👾 note: IIRC callsites often guard against flag inputs, but we might follow up with changes here to accept these in configurations so these prompts guard against scripting fallthrough!

🐌 ramble: I'm curious about the default behavior for `confirm` prompts too when interactive prompts aren't available since we might use the default choice instead of requiring flag alternatives to perhaps revisit the changes of #87?

mwbrooks: I agree, in non-TTY it should use the default choice and we should offer a flag for each confirm to override the choice.

zimeg: Thanks for a check here! Let's follow up with defaults and also options as changes to prompts become stable. We're not regressing with current behavior, but it's a known gap now!

PR #409, opened 2026-03-18
test: increase test coverage for pkg/version and slackdeps

mwbrooks:
### Changelog

- N/A

### Summary

This pull request increases our test coverage in `internal/pkg/version` and `internal/slackdeps`.

I skipped updating a few other `internal/pkg` since there are some active refactoring going on in those packages and I want to avoid overcomplicating that work.

Tests are ordered alphabetically, grouping getters/setters, and constructors are at the top.

### Requirements

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,7 +1,9 @@
 {
   "permissions": {
     "allow": [
+      "Bash(go build:*)",
       "Bash(go mod tidy:*)",
+      "Bash(go test:*)",
       "Bash(gofmt:*)",
       "Bash(gh issue view:*)",
       "Bash(gh label list:*)",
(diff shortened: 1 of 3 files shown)


internal/slackdeps/os_test.go
```diff
@@ -15,11 +15,146 @@
...
+		"returns empty for unset env var": {
+			key:      "SLACK_TEST_OS_GETENV_UNSET",
+			expected: "",
+		},
+	}
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			o := NewOs()
+			if tc.value != "" {
+				t.Setenv(tc.key, tc.value)
```
mwbrooks, line 106: note: `t.Setenv(...)` is a built-in that backs up the original environment variable value, sets the new value, and then restores the original when the test completes. I'd rather use a 100% mocked call, but this is a good and safe start.

PR #410, opened 2026-03-18
fix: unquote environment variable passed to hook commands

zimeg:
### Changelog

> Fixes environment variable values set for hook commands to no longer be surrounded with double quotes.

### Summary

This PR fixes an issue where environment variables were set with double quotes surrounding values:

https://github.com/slackapi/slack-cli/blob/8056444e300d229cfcf726bff6ea3794d678a070/internal/goutils/map.go#L23-L39

```
	// Env specifies the environment of the process.
	// Each entry is of the form "key=value".
```

📚 Reference: https://pkg.go.dev/os/exec#Cmd

### Requirements

### DIFF
--- a/internal/hooks/hook_executor_v2_test.go
+++ b/internal/hooks/hook_executor_v2_test.go
@@ -55,7 +55,8 @@ func Test_Hook_Execute_V2_Protocol(t *testing.T) {
 			opts: HookExecOpts{
 				Hook: HookScript{Name: "happypath", Command: "echo {}"},
 				Env: map[string]string{
-					"BATMAN": "robin hood",
+					"BATMAN": "robin",
+					"WHOAMI": "lumpy space princess",
 					"YIN":    "yang",
 				},
 				Exec: &MockExec{
@@ -68,7 +69,8 @@ func Test_Hook_Execute_V2_Protocol(t *testing.T) {
 			check: func(t *testing.T, response string, err error, mockExec ExecInterface) {
 				require.Equal(t, `{"message": "hello world"}`, response)
 				require.Equal(t, nil, err)
-				require.Contains(t, mockExec.(*MockExec).mockCommand.Env, `BATMAN=robin hood`)
+				require.Contains(t, mockExec.(*MockExec).mockCommand.Env, `BATMAN=robin`)
+				require.Contains(t, mockExec.(*MockExec).mockCommand.Env, `WHOAMI=lumpy space princess`)
 				require.Contains(t, mockExec.(*MockExec).mockCommand.Env, `YIN=yang`)
 			},
 		},
@@ -76,9 +78,8 @@ func Test_Hook_Execute_V2_Protocol(t *testing.T) {
 			opts: HookExecOpts{
 				Hook: HookScript{Name: "happypath", Command: "echo {}"},
 				Env: map[string]string{
-					"batman": "robin",
-"BATMAN": "robin",
-"YIN":    "yang",
+					"BATMAN": "robin",
+					"YIN":    "yang",
 				},
 				Exec: &MockExec{
 					mockCommand: &MockCommand{
--- a/internal/hooks/hook_executor_default_test.go
+++ b/internal/hooks/hook_executor_default_test.go
@@ -60,7 +60,8 @@ func Test_Hook_Execute_Default_Protocol(t *testing.T) {
 			opts: HookExecOpts{
 				Hook: HookScript{Name: "happypath", Command: "echo {}"},
 				Env: map[string]string{
-					"BATMAN": "robin hood",
+					"BATMAN": "robin",
+					"WHOAMI": "lumpy space princess",
 					"YIN":    "yang",
 				},
 				Exec: &MockExec{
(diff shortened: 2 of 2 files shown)


internal/hooks/hooks.go
```diff
@@ -57,7 +57,9 @@ func processExecOpts(opts HookExecOpts) ([]string, []string, []string, error) {
 	// To avoid removing any environment variables that are set in the current environment, we first set the cmd.Env to the current environment.
 	// before adding any new environment variables.
 	var cmdEnvVars = os.Environ()
-	cmdEnvVars = append(cmdEnvVars, goutils.MapToStringSlice(opts.Env, "")...)
+	for name, value := range opts.Env {
+		cmdEnvVars = append(cmdEnvVars, name+"="+value)
```
zimeg, line 61: 🔭 note: These values are set for both the **default** and **message-boundaries** protocols:

https://github.com/slackapi/slack-cli/blob/8056444e300d229cfcf726bff6ea3794d678a070/internal/hooks/hook_executor_default.go#L70

https://github.com/slackapi/slack-cli/blob/8056444e300d229cfcf726bff6ea3794d678a070/internal/hooks/hook_executor_v2.go#L87

https://github.com/slackapi/slack-cli/blob/8056444e300d229cfcf726bff6ea3794d678a070/internal/hooks/shell.go#L56-L59

internal/pkg/platform/localserver.go
```diff
@@ -307,7 +306,9 @@ func (r *LocalServer) StartDelegate(ctx context.Context) error {
 	// To avoid removing any environment variables that are set in the current environment, we first set the cmd.Env to the current environment.
 	// before adding any new environment variables.
 	var cmdEnvVars = os.Environ()
-	cmdEnvVars = append(cmdEnvVars, goutils.MapToStringSlice(sdkManagedConnectionStartHookOpts.Env, "")...)
```
zimeg, line 310: 🏁 note: I understand the `MapToStringSlice` utilities to be useful in escaping command line arguments but we might not need this with the direct environment variable value setting in commands.

🏆 ramble: This implementation in `localserver` is the other find of a direct call to the hooks command executor so shares the change!

internal/hooks/hook_executor_v2_test.go
mwbrooks: suggestion: Minor suggestion, but we could capitalize these to match the above test case format. If so, we will also need to update the expected values.

```suggestion
"BATMAN": "robin",
"YIN":    "yang",
```

internal/hooks/hook_executor_v2_test.go
```diff
@@ -89,8 +89,8 @@ func Test_Hook_Execute_V2_Protocol(t *testing.T) {
 			check: func(t *testing.T, response string, err error, mockExec ExecInterface) {
 				require.Equal(t, sixtyFourKBString, response)
 				require.Equal(t, nil, err)
-				require.Contains(t, mockExec.(*MockExec).mockCommand.Env, `batman="robin"`)
-				require.Contains(t, mockExec.(*MockExec).mockCommand.Env, `yin="yang"`)
+				require.Contains(t, mockExec.(*MockExec).mockCommand.Env, `batman=robin`)
+				require.Contains(t, mockExec.(*MockExec).mockCommand.Env, `yin=yang`)
```
mwbrooks: suggestion: Only merge if we're capitalizing the suggestion above on L80.
```suggestion
				require.Contains(t, mockExec.(*MockExec).mockCommand.Env, `BATMAN=robin`)
				require.Contains(t, mockExec.(*MockExec).mockCommand.Env, `YIN=yang`)
```

internal/hooks/hook_executor_v2_test.go
```diff
@@ -55,8 +55,8 @@ func Test_Hook_Execute_V2_Protocol(t *testing.T) {
 			opts: HookExecOpts{
 				Hook: HookScript{Name: "happypath", Command: "echo {}"},
 				Env: map[string]string{
-					"batman": "robin",
-					"yin":    "yang",
+					"BATMAN": "robin hood",
+					"YIN":    "yang",
```
mwbrooks: note: I think we want to keep it as Batman and Robin (the duo/sidekick to Batman), not Batman and Robinhood (two separate fictional characters).

zimeg: @mwbrooks I didn't want to take from the fun! 😉 

I'll follow up with an update to add another example that includes a space elsewhere, which is the intention of initial change I meant!

internal/hooks/hook_executor_v2_test.go
zimeg: @mwbrooks Amazing eye for details 👁️‍🗨️ ✨  Let's include this in https://github.com/slackapi/slack-cli/pull/410/commits/d2f878be60be0df1ae698d61e6f3b67ddd30ac80 for now 🎁

internal/hooks/hook_executor_v2_test.go
zimeg: @mwbrooks We now assert correctness with https://github.com/slackapi/slack-cli/pull/410/commits/d2f878be60be0df1ae698d61e6f3b67ddd30ac80 🧪 ✨

internal/hooks/hook_executor_v2_test.go
zimeg: @mwbrooks A favorite [LSP](https://www.youtube.com/watch?v=cjRGbOwpbtM) is added in https://github.com/slackapi/slack-cli/pull/410/commits/28ff7f1b11ab9d047d617b900345c4aac1a833e0!

PR #419, opened 2026-03-19
test: increase test coverage for activity package formatters

mwbrooks:
### Changelog

- N/A

### Summary

This pull request increases the test coverage of `internal/pkg/platform/activity`:

- Add unit tests for all activity log formatting helper functions
- Brings coverage for individual *ToString functions from 0% to 100%

The tests are separated into 2 PRs to keep things easier to review. While this is +534 lines, there are a lot of `[]string{}` slices that expand across multiple lines.

### Requirements

### DIFF
--- a/cmd/sandbox/create.go
+++ b/cmd/sandbox/create.go
@@ -1,0 +1,278 @@
+// Copyright 2022-2026 Salesforce, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package sandbox
+
+import (
+	"fmt"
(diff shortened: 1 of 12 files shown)


internal/pkg/platform/activity_test.go
```diff
@@ -384,6 +384,540 @@ func TestPlatformActivity_TriggerExecutedToString(t *testing.T) {
...
+		Level:       types.ERROR,
+		ComponentID: "comp1",
+		TraceID:     "trace1",
+		Payload: map[string]interface{}{
+			"code": "wf_abc",
+		},
+		Created: 1686939542000000,
+	}
+	result := externalAuthMissingSelectedAuthToString(activity)
+	assert.Contains(t, result, "Missing mapped token for workflow 'wf_abc'")
```
mwbrooks, line 414: note: Don't love the hardcoded string here, but matching a partial also didn't feel right in this situation.

zimeg: @mwbrooks I think this is a solid check to keep! 📠 

The complete log format might be interesting to match if it's templated, but I don't know if this is needed before merge?

https://github.com/slackapi/slack-cli/blob/55a4069dfd63060c96bbe4a1e9fd6668fe813bbd/internal/pkg/platform/activity.go#L40

mwbrooks: 🧠 Ah, that's a clever idea! For now just merge the simple test that focuses on the outcome, but if we find it's hard to maintain then perhaps we lean on the log template 🤔

PR #421, opened 2026-03-20
feat(experiment): apply similar theme to huh forms

srtaalej:
### Changelog
na

### Summary
- Add `ThemeSurvey()` huh theme that replicates the legacy survey prompt appearance (gray+bold titles, blue+bold cursors/chevrons, `[x]`/`[ ]` multi-select prefixes, red errors) so huh forms look identical to survey when only `experiment.Huh` is on without `experiment.Lipgloss`
- Apply `ThemeSurvey()` as default in `newForm()` — `ThemeSlack()` is used when Lipgloss is on, `ThemeSurvey()` otherwise
- Style select option descriptions with bold option name + gray description text (via `style.Bright`/`style.Secondary`) and em-dash separator consistently across both experiment paths

https://github.com/user-attachments/assets/bc51c791-aa5b-4e65-982d-c4aece4edc12

### Testing Steps
```bash
make build 
make test     # all tests should pass
slack samples -e huh
slack create -e huh
slack create my-app -e huh
slack create -t slack-samples/bolt-js-starter-template
```

### Requirements


internal/iostreams/forms.go
```diff
@@ -90,7 +92,7 @@ func buildSelectForm(io *IOStreams, msg string, options []string, cfg SelectProm
 		key := opt
 		if cfg.Description != nil {
 			if desc := style.RemoveEmoji(cfg.Description(opt, len(opts))); desc != "" {
-				key = opt + " - " + desc
+				key = style.Bright(opt) + " — " + style.Secondary(desc)
```
zimeg, line 95: 🔭 thought: If we're changing the "—" separator to a character outside of the [ASCII character sets](https://en.wikipedia.org/wiki/ASCII#Character_set) we can perhaps use a new function to fallback to the common dash without styles?

```
style.Separator
```

👾 quibble: This might be a change to discuss in a separate PR. IMHO #400 isn't an experimental change but it was related to prompts at the time of these updates and gets confusing...

internal/style/theme.go
```diff
@@ -126,6 +126,54 @@ func Chevron() string {
...
+		Foreground(ansiRed)
+
+	// Select styles
+	t.Focused.SelectSelector = lipgloss.NewStyle().
+		Foreground(ansiBlue).
+		Bold(true).
+		SetString(Chevron() + " ")
+	t.Focused.SelectedOption = lipgloss.NewStyle().
+		Foreground(ansiBlue).
+		Bold(true)
```
zimeg, line 160: 🌟 praise: These selections have a impressive similarities! The change to help text placement builds confidence in this next update I think!

zimeg: 📸 **Before**

Image

📸 **After**

Image

zimeg: 👾 issue(non-blocking): I notice the guided tutorials reference isn't included in the updated help description.

🪬 ramble: This might be a quick change to the `create` selection but we might want to check for similar custom templates and other deprecated fields being used in other prompts. I'd lean toward a separate PR since this doesn't seem related to the theme itself, if this is something you'd want to check out related?

internal/style/theme.go
```diff
@@ -126,6 +126,54 @@ func Chevron() string {
...
+func ThemeSurvey() huh.Theme {
+	return huh.ThemeFunc(themeSurvey)
+}
+
+// themeSurvey builds huh styles matching the survey library's appearance.
+func themeSurvey(isDark bool) *huh.Styles {
+	t := huh.ThemeBase(isDark)
+
+	ansiBlue := lipgloss.ANSIColor(blue)
+	ansiGray := lipgloss.ANSIColor(gray)
```
zimeg, line 140: 👽 praise: This does solid in foreground text where used!

zimeg: 🪐 **Before**

Image

💫 **After**

Image

internal/style/theme.go
```diff
@@ -126,6 +126,54 @@ func Chevron() string {
...
+	ansiGreen := lipgloss.ANSIColor(green)
+	ansiRed := lipgloss.ANSIColor(red)
+
+	t.Focused.Title = lipgloss.NewStyle().
+		Foreground(ansiGray).
+		Bold(true)
```
zimeg, line 146: 🌲 praise: Nice use of **bold** formatting once more.

zimeg: 🌈 **Before**

Image

🍯 **After**

Image

internal/style/theme.go
srtaalej: ah thanks for catching that! ill include it in a follow up PR

internal/iostreams/forms.go
srtaalej: I agree!

zimeg: @srtaalej Sweet! Could you link the PR here for more discussion?

internal/style/theme.go
srtaalej: do we prefer it at the top
Screenshot 2026-03-23 at 12 26 44 PM

or bottom of the prompt
Screenshot 2026-03-23 at 12 33 37 PM

zimeg: @srtaalej Hmmm... I might prefer the top since it doesn't remove instructions below 🍀 ✨

PR #425, opened 2026-03-20
feat: replace existing survey prompts with beautiful charm forms

srtaalej:
### Changelog

> The interactive prompts asking for input had a huge overhaul and now these selections and other inputs are collected in beautiful forms built with the [`huh`](https://github.com/charmbracelet/huh) package.
>
> Huge thanks to the @charmbracelet team for help and kind encouragement in these improvements! 🍀 ✨ 

### Summary
 This PR completes the migration from the survey library to Charm's huh package for interactive prompts. 
 The `experiment.Huh` flag and all survey-based prompt code have been removed, making huh the permanent and only prompt implementation. 
 The `ThemeSurvey()` theme (introduced in the parent PR) ensures visual continuity by matching the legacy survey styling when the Lipgloss experiment is off.

### Test Steps
```zsh
make test testdir=internal/iostreams   # and verify all prompt form tests pass including the ThemeSurvey and ThemeSlack theme assertions
make test testdir=internal/style   # and confirm TestThemeSlack, TestThemeSurvey, and TestChevron pass without the removed TestSurveyIcons
make test testdir=internal/experiment    # and confirm the Test_Includes assertion no longer expects "huh" to be a registered experiment
make test testdir=cmd/project    # to verify the sample selection flow works without experiment gating
make lint     # to confirm zero issues
slack create    # interactively and verify prompts render with the expected cursor, selection indicators, and color scheme
slack env add MY_VAR    # without a --value flag and verify the mimicked input line still appears before the password prompt
```
### Requirements

### DIFF
--- a/docs/reference/experiments.md
+++ b/docs/reference/experiments.md
@@ -6,7 +6,6 @@ The Slack CLI has an experiment (`-e`) flag behind which we put features current
 
 The following is a list of currently available experiments. We'll remove experiments from this page if we decide they are no longer needed or once they are released, in which case we'll make an announcement about the feature's general availability in the [developer changelog](https://docs.slack.dev/changelog).
 
-- `huh`: shows beautiful prompts.
 - `lipgloss`: shows pretty styles.
 - `sandboxes`: enables users who have joined the Slack Developer Program to manage their sandboxes ([PR#379](https://github.com/slackapi/slack-cli/pull/379)).
 - `templates`: brings more agent templates to the `create` command.
@@ -14,7 +13,7 @@ The following is a list of currently available experiments. We'll remove experim
 ## Experiments changelog
 
 Below is a list of updates related to experiments.
-
+- **April 2026**: Concluded the `huh` experiment with full support now enabled by default in the Slack CLI. 
 - **March 2026**: Split the `charm` experiment into more beautiful `huh` prompts and prettier `lipgloss` styles for ongoing change.
 - **March 2026**: Concluded the `bolt` and `bolt-install` experiments with full Bolt framework support now enabled by default in the Slack CLI. All Bolt project features including remote manifest management are now standard functionality. See the announcement [here](https://slack.dev/slackcli-supports-bolt-apps/).
 - **February 2026**: Added the `charm` experiment.


internal/iostreams/prompts.go
```diff
@@ -21,11 +21,21 @@ import (
...
+// MimicInputPrompt formats a message and value to appear as a prompted input
+func MimicInputPrompt(message string, value string) string {
+	return fmt.Sprintf(
+		"%s %s %s",
+		style.Darken("?"),
+		style.Highlight(message),
+		style.Input(value),
+	)
+}
+
```
mwbrooks: question: Why was this added? A quick search doesn't show that it's called anywhere. 

Since this PR is focused on removing survey, if we need to add `MimicInputPrompt` then we should probably do it in a separate PR to keep this one focused on removal.

srtaalej: its used in [@add.go](https://github.com/slackapi/slack-cli/blob/main/cmd/env/add.go) `env add` command to display variable names. since it was in survey.go i figured i should move it to prompt.go but perhaps it makes more sense to have it directly in add.go...

cmd/project/create_samples_test.go
```diff
@@ -128,7 +128,12 @@ func TestSamples_PromptSampleSelection(t *testing.T) {
...
 					Flag: clientsMock.Config.Flags.Lookup("template"),
 				}),
-			).Return(tc.mockSelection, nil)
+			).Run(func(args mock.Arguments) {
+				cfg := args.Get(3).(iostreams.SelectPromptConfig)
+				if cfg.Description != nil {
+					cfg.Description("", 0)
+				}
+			}).Return(tc.mockSelection, nil)
 			clients := shared.NewClientFactory(clientsMock.MockClientFactory())
```
zimeg, line 132: 👁️‍🗨️ question: Can this be checked within `iostreams.MatchPromptConfig` or does this assert something about the orderings?

internal/iostreams/prompts.go
zimeg: 🪓 @srtaalej This is alright to remove IIRC - it was included with **survey** as a placeholder output for the following command in particular:

Image

👾 ramble: We wanted this to mimic the input form with an argument, but now that forms disappear we might be alright to remove this too:

https://github.com/slackapi/slack-cli/blob/0c3712813155a659fbde3b16e071ed065fd987d5/cmd/env/add.go#L106-L112

internal/iostreams/survey.go
zimeg, line 1: 🪬 praise: Kind thanks to the good [`survey`](https://github.com/AlecAivazis/survey/) package-

zimeg (commented):
🌈 Awesome awesome awesome! Thank you for getting this started @srtaalej!

I'm leaving comments from some adjacent bug hunts with findings and ideas toward this PR. Those checks make this so promising I think! 🍀

cmd/project/create_samples_test.go
srtaalej: this is going by order!

zimeg: @srtaalej Forgive my confusion, but can you explain what L134 is doing? I haven't seen this mock pattern before...

zimeg (approved):
@srtaalej This is looking fantastic! 🍀 ✨

Before merging let's note that the `huh` experiment is concluding in docs:

📚 https://docs.slack.dev/tools/slack-cli/reference/experiments

Also I'd like to address the comment I left earlier about testing mocks. Adding this pattern feels odd to me as part of these changes since I understand this to be a clean replacement in otherwise?

internal/iostreams/prompts.go
zimeg: @mwbrooks @srtaalej It looks like this was addressed in https://github.com/slackapi/slack-cli/pull/425/commits/c9fb8016c5f59ac9d2397b655eddb9179ebf6bc2 commit

cmd/project/create_samples_test.go
srtaalej: ah i see its calling Description so that codecov doesnt flag it as uncovered. i can remove it!

PR #427, opened 2026-03-20
test: add units tests for untested functions in style, api, types, and create packages

mwbrooks:
### Changelog

- N/A

### Summary

This pull request is that **last** in my effort to add test coverage to our untested functions. The remaining, untested functions are part of `package style` that I'm trying to leave alone while we have the `huh` and `lipgloss` experiments.

While not hugely impactful, this should increase our test coverage by about +0.5%.

### Requirements

### DIFF
--- a/cmd/platform/run.go
+++ b/cmd/platform/run.go
@@ -125,7 +125,7 @@ func RunRunCommand(clients *shared.ClientFactory, cmd *cobra.Command, args []str
 		}
 	}
 
-	runFlags.orgGrantWorkspaceID, err = prompts.ValidateGetOrgWorkspaceGrant(ctx, clients, &selection, runFlags.orgGrantWorkspaceID, false /* top prompt option should be 'all workspaces' */)
+	runFlags.orgGrantWorkspaceID, err = prompts.ValidateGetOrgWorkspaceGrant(ctx, clients, &selection, runFlags.orgGrantWorkspaceID, true /* top prompt option should be 'all workspaces' */)
 	if err != nil {
 		return err
 	}


internal/pkg/create/template_test.go
```diff
@@ -23,6 +23,24 @@ import (
...
+		assert.Equal(t, "subpath", tmpl.GetSubdir())
+	})
+}
+
+func Test_Template_SetSubdir(t *testing.T) {
+	tmpl := Template{}
+	tmpl.SetSubdir("custom/dir")
+	assert.Equal(t, "custom/dir", tmpl.GetSubdir())
+}
+
```
mwbrooks, line 43: note: It makes me feel good to see these have some tests added.

internal/shared/types/app_manifest_test.go
```diff
@@ -48,6 +48,34 @@ func Test_RawJSON_UnmarshalJSON(t *testing.T) {
...
+func Test_RawJSON_MarshalJSON(t *testing.T) {
+	tests := map[string]struct {
+		rawJSON  RawJSON
+		expected string
+	}{
+		"marshals JSONData when present": {
+			rawJSON: func() RawJSON {
+				raw := json.RawMessage(`{"name":"foo"}`)
+				return RawJSON{JSONData: &raw}
+			}(),
```
mwbrooks, line 60: note: TIL about an anonymous function as a way to populate a variable in our test tables. It's actually handy, although it can be difficult to read when overused.

internal/shared/types/slack_yaml_test.go
```diff
@@ -0,0 +1,112 @@
...
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func Test_SlackYaml_hasValidIconPath(t *testing.T) {
```
mwbrooks, line 26: threat: I'm coming for you `SlackYaml`. Not today. Not tomorrow. But one day, I'm coming for you and giving you a new name.

internal/style/format_test.go
```diff
@@ -158,6 +197,38 @@ func TestCommandf(t *testing.T) {
...
+			defer func() {
+				os.Args[0] = processTemp
+				isColorShown = globalColorShown
+			}()
+
+			formatted := Commandf(tc.command, tc.isPrimary)
+			assert.NotContains(t, formatted, "`")
+			assert.Contains(t, formatted, tc.command)
+		})
+	}
```
mwbrooks, line 229: note: A bit scary how we change the `os.Args[0]` but I feel okay with the `defer func()` to reset it.

internal/style/format_test.go
```diff
@@ -255,6 +326,50 @@ func TestStyleFlags(t *testing.T) {
...
+		expected string
+	}{
+		"path under home directory gets tilde": {
+			input: homeDir + "/Documents/project",
+			expected: func() string {
+				if isWindows {
+					return homeDir + "/Documents/project"
+				}
+				return "~/Documents/project"
+			}(),
```
mwbrooks, line 346: note: Again, I thought this was a nice way to return a `string` for the one test case that needs a bit of logic.

internal/api/activity_test.go
```diff
@@ -211,3 +213,11 @@ func Test_APIClient_ActivityInvalidJSON(t *testing.T) {
 	require.Error(t, err)
 	require.Contains(t, err.Error(), slackerror.ErrUnableToParseJSON)
 }
+
+func Test_Activity_CreatedPretty(t *testing.T) {
+	// Created is in microseconds
+	createdMicroseconds := int64(1700000000000000)
+	activity := Activity{Created: createdMicroseconds}
+	expected := time.Unix(createdMicroseconds/1000000, 0).Format("2006-01-02 15:04:05")
+	assert.Equal(t, expected, activity.CreatedPretty())
```
zimeg, line 222: 🐞 praise: Pretty time at `L222`!

internal/shared/types/slack_yaml_test.go
```diff
@@ -0,0 +1,112 @@
...
+	tests := map[string]struct {
+		icon     string
+		setup    func(t *testing.T, dir string)
+		expected bool
+	}{
+		"valid custom icon path returns true": {
+			icon: "custom/icon.png",
+			setup: func(t *testing.T, dir string) {
+				require.NoError(t, os.MkdirAll(filepath.Join(dir, "custom"), 0o755))
+				require.NoError(t, os.WriteFile(filepath.Join(dir, "custom", "icon.png"), []byte("img"), 0o644))
```
zimeg, line 36: 🎁 praise: Exciting tests to find for this YAML package to get in meantimes!

internal/shared/types/slack_yaml_test.go
```diff
@@ -0,0 +1,112 @@
...
+		},
+		"no icon and no default returns true": {
+			icon:     "",
+			setup:    func(t *testing.T, dir string) {},
+			expected: true,
+		},
+	}
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			dir := t.TempDir()
```
zimeg, line 61: 🤖 thought: It stinks we aren't using `aefero` here... I'm curious if this might be check better kept for callsites in upcoming change?

internal/style/format_test.go
```diff
@@ -52,6 +54,43 @@ func TestGetKeyLength(t *testing.T) {
...
+			input: map[string]string{},
+			validate: func(t *testing.T, result string) {
+				assert.Equal(t, "", result)
+			},
+		},
+		"single entry formats correctly": {
+			input: map[string]string{"Name": "my-app"},
+			validate: func(t *testing.T, result string) {
+				assert.Contains(t, result, "Name")
+				assert.Contains(t, result, "my-app")
```
zimeg, line 74: 🧪 thought: Testing formats might have good findings in exact strings, but no blocker for new coverage!

internal/style/format_test.go
zimeg: 🌟 praise: Nice catch to keep tests confident across builds.

internal/style/format_test.go
```diff
@@ -408,3 +523,73 @@ func TestLocalRunDisplayNamePlain(t *testing.T) {
...
+	now := int(time.Now().Unix())
+	tests := map[string]struct {
+		datetime     int
+		containsUnit string
+		containsDir  string
+	}{
+		"seconds ago": {
+			datetime:     now - 30,
+			containsUnit: "second",
+			containsDir:  "ago",
```
zimeg, line 545:
```suggestion
			datetime: now - 30,
			expected: "30 seconds ago",
```

🧪 thought: Is asserting the expected difference meaningful to include now?

🎙️ ramble: I believe this logic hasn't caused known issues so we might save this for found regression of course!

internal/shared/types/slack_yaml_test.go
mwbrooks: Agreed, we have a few places where we use `t.TempDir()` instead of `afero`. FWIW I have a personal TODO to loopback and switch these to `afero`, if possible, after we hit 70% test coverage 😄

internal/style/format_test.go
mwbrooks: 🧠 Both are fair points @zimeg.

I like the `expected` design choice, but I think there are practical challenges to using it. 

Our tests use `time.Now()`, which could lead to timing issues. By the time `TimeAgo` runs, a fraction of a second may have elapsed causing the match to fail (e.g. we might get "31 seconds ago" instead of "30 seconds ago").

For now, I think we should keep the looser tests so that we avoid flaky test results.

PR #428, opened 2026-03-21
refactor: move internal/pkg/version to internal/version

mwbrooks:
### Changelog

- N/A

### Summary

This pull request refactors `internal/pkg/version` to `internal/version`. The motivation is to reduce the usage of `internal/pkg` - all of these packages should eventually find a home outside of `internal/pkg`.

While this is a small change, it does touch our build configuration.

### Test Steps

```bash
$ lack --version
# → Confirm the correct SHA is in the version: `-g<SHA>`
#   → "v3.15.0-26-gc082ce3"
```

### Requirements

- [x] I've read and understood the [Contributing Guidelines](https://github.com/slackapi/slack-cli/blob/main/.github/CONTRIBUTING.md) and have done my best effort to follow them.
- [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).

PR #429, opened 2026-03-21
refactor: remove legacy CLIVersion field and SetVersion() function

mwbrooks:
### Changelog

- N/A

### Summary

This pull request removes the legacy version references from `ClientFactory`, including the `ClientFactory.CLIVersion` field and `shared.SetVersion()` function.

- Removes the `CLIVersion` field from `ClientFactory`.
- Removes the `SetVersion()` functional from `ClientFactory`.
- Removes stale `TODO` comment referencing a circular dependency that no longer exists. This was solved by `slackcontext.Version(ctx)`.
- Replaces indirect version access (`clients.CLIVersion`) with direct `version.Raw()` calls. The `clients.CLIVersion` was previously set to the raw version value, so the functionality is unchanged.

**Context:**

The `ClientFactory.CLIVersion` and `shared.SetVersion()` were a workaround for a circular dependency between `internal/api/client.go` and `package version`. That dependency no longer exists - `internal/api/client.go` does not import the version package and the API client now reads the CLI version from context with `slackcontext.Version(ctx)`.

**Open Questions:**

- In a previous PR, we had concerns that this may have unexpected impacts around the login:
  - https://github.com/slackapi/slack-cli/pull/428#discussion_r2969823381
  - Solution is to test the `login` command in our **Test Steps**.

### Test Steps

**1. Test Version:**

```bash
$ lack version
# → Expect: "Using slack v3.15.0-26-ga164d25"

$ lack --version
# → Expect: "Using slack v3.15.0-26-ga164d25"
```

**2. Test Version Override with Environment Variable:**

```bash
$ SLACK_TEST_VERSION=v0.0.1 lack version
# → Expect: "Using slack v0.0.1"
```

**3. Test Doctor Command (confirms `Config.Version` is set):**

```bash
$ lack doctor
# → Expect:
#     ✔ CLI (this tool for building Slack apps)
#      Version: v3.15.0-26-ga164d25
```

**4. Test Debug Output (confirms version reaches context)**

```bash
$ lack auth list --verbose --skip-update 2>&1 | grep -i "version"
# → Expect: "cli_version":"3.15.0-26-ga164d25"
```

**5. Test Login (confirms version accepted by API)**

```bash
$ lack login
# → Login to a new or existing auth
# → Expect: Login successful
```

### Requirements

### DIFF
--- a/internal/iostreams/forms.go
+++ b/internal/iostreams/forms.go
@@ -34,6 +34,8 @@ func newForm(io *IOStreams, field huh.Field) *huh.Form {
 	form := huh.NewForm(huh.NewGroup(field))
 	if io != nil && io.config.WithExperimentOn(experiment.Lipgloss) {
 		form = form.WithTheme(style.ThemeSlack())
+	} else {
+		form = form.WithTheme(style.ThemeSurvey())
 	}
 	return form
 }
@@ -90,7 +92,7 @@ func buildSelectForm(io *IOStreams, msg string, options []string, cfg SelectProm
 		key := opt
 		if cfg.Description != nil {
 			if desc := style.RemoveEmoji(cfg.Description(opt, len(opts))); desc != "" {
-				key = opt + " - " + desc
+				key = style.Bright(opt) + " — " + style.Secondary(desc)
 			}
 		}
 		opts = append(opts, huh.NewOption(key, opt))
--- a/internal/iostreams/forms_test.go
+++ b/internal/iostreams/forms_test.go
@@ -195,6 +195,57 @@ func TestSelectForm(t *testing.T) {
 		assert.Contains(t, view, "First letter")
 	})
 
+	t.Run("descriptions use em-dash separator with lipgloss enabled", func(t *testing.T) {
+		style.ToggleLipgloss(true)
+		style.ToggleStyles(true)
+		t.Cleanup(func() {
+			style.ToggleLipgloss(false)
+			style.ToggleStyles(false)
+		})
+
+		fsMock := slackdeps.NewFsMock()
+		osMock := slackdeps.NewOsMock()
+		osMock.AddDefaultMocks()
+		cfg := config.NewConfig(fsMock, osMock)
+		cfg.ExperimentsFlag = []string{"lipgloss"}
+		cfg.LoadExperiments(context.Background(), func(_ context.Context, _ string, _ ...any) {})
+		io := NewIOStreams(cfg, fsMock, osMock)
+
+		var selected string
+		options := []string{"Alpha", "Beta"}
+		selectCfg := SelectPromptConfig{
+			Description: func(opt string, _ int) string {
+				if opt == "Alpha" {
+					return "First letter"
+				}
+				return ""
+			},
+		}
+		f := buildSelectForm(io, "Choose", options, selectCfg, &selected)
+		f.Update(f.Init())
+
+		view := ansi.Strip(f.View())
+		assert.Contains(t, view, " — First letter")
+	})
+
+	t.Run("descriptions use em-dash separator without lipgloss", func(t *testing.T) {
+		var selected string
+		options := []string{"Alpha", "Beta"}
+		selectCfg := SelectPromptConfig{
+			Description: func(opt string, _ int) string {
+				if opt == "Alpha" {
+					return "First letter"
+				}
+				return ""
+			},
+		}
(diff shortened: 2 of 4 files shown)


cmd/root.go
```diff
@@ -265,7 +265,7 @@ func InitConfig(ctx context.Context, clients *shared.ClientFactory, rootCmd *cob
 
 	// Init clients that use flags
 	clients.Config.APIHostResolved = clients.Auth().ResolveAPIHost(ctx, clients.Config.APIHostFlag, nil)
-	clients.Config.LogstashHostResolved = clients.Auth().ResolveLogstashHost(ctx, clients.Config.APIHostResolved, clients.CLIVersion)
```
mwbrooks, line 268: note: `clients.CLIVersion` was set to `version.Raw()`.

cmd/root.go
```diff
@@ -297,7 +297,7 @@ func InitConfig(ctx context.Context, clients *shared.ClientFactory, rootCmd *cob
 	clients.Config.LoadExperiments(ctx, clients.IO.PrintDebug)
 	style.ToggleLipgloss(clients.Config.WithExperimentOn(experiment.Lipgloss))
 	// TODO(slackcontext) Consolidate storing CLI version to slackcontext
-	clients.Config.Version = clients.CLIVersion
```
mwbrooks, line 300: note: `clients.CLIVersion` was set to `version.Raw()`.

internal/shared/clients.go
```diff
@@ -92,14 +91,6 @@ func NewClientFactory(options ...func(*ClientFactory)) *ClientFactory {
...
 	clients.Browser = clients.defaultBrowserFunc
 
-	// TODO: Temporary hack to get around circular dependency in internal/api/client.go since that imports version
-	// Follows pattern demonstrated by the GitHub CLI here https://github.com/cli/cli/blob/5a46c1cab601a3394caa8de85adb14f909b811e9/pkg/cmd/factory/default.go#L29
-	// Used by the APIClient for its userAgent
-	// Currently needed because trying to get the version of the CLI from pkg/version/version.go would cause a circular dependency
-	// We can get rid of this once we refactor the code relationship between pkg/ and internal/
-	// userAgent can get Slack CLI version from context which is defined in main.go, this approach bypass circular dependency. The clients.CLIVersion is retained for future code refactor purpose and serve SetVersion function
-	clients.CLIVersion = ""
-
```
mwbrooks, line 101: 🪓

main.go
```diff
@@ -80,7 +80,8 @@ func main() {
 func recoveryFunc() {
 	// in the event of a panic, log panic
 	if r := recover(); r != nil {
-		var clients = shared.NewClientFactory(shared.SetVersion(version.Raw()))
+		var clients = shared.NewClientFactory()
+		clients.Config.Version = version.Raw()
```
mwbrooks, line 84: thought: I still don't like that we're setting the version in multiple places, instead of just referencing the source of truth (`version.Raw()` or `version.Get()`). However, I think this PR is a step toward simpler code with fewer indirect references. We can follow-up further in more PRs if we want.

zimeg: 🔭 thought: For this `panic` catching I think it's alright to forgive some duplication and strange logic.

🌟 praise: Thanks for catching this in these updates still!

cmd/root.go
```diff
@@ -297,7 +297,7 @@ func InitConfig(ctx context.Context, clients *shared.ClientFactory, rootCmd *cob
...
+	clients.Config.Version = version.Raw()
```
zimeg, line 300: 🔍 thought(non-blocking): Would setting this version above - before `L268` above - let us reuse the configuration version instead of calling to the `version` package twice?

👾 ramble: I'm unsure if this value is set here due to dependencies above, but I might expect the version to remain stable during execution here!

zimeg: 🗣️ ramble: Forgive me if the `TODO` of "slackcontext" above is meant to capture similar changes!

internal/shared/clients.go
zimeg: 🌲 praise: This has bothered me more than I understood. Thanks for making change that improves the code @mwbrooks!

cmd/root.go
mwbrooks: thought: I have a branch that removes `clients.Config.Version` entirely. I'll give it a second look, but my first impression is that it simplifies how we pass version around.

PR #430, opened 2026-03-23
feat(style): add style.Separator() for hyphen consistency

srtaalej:
### Changelog

n/a

### Summary
This PR adds style.Separator() function that returns an hyphen (" - ") for use as a consistent delimiter between option names and descriptions in prompts. Replace hardcoded em-dash string in forms.go with style.Separator() for consistency with other style functions  

Screenshot 2026-03-23 at 12 12 41 PM

### Test plan      
```bash
make test testdir=internal/iostreams                                
make test testdir=internal/style
make lint 
``` 
### Requirements

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -148,7 +148,7 @@ func Init(ctx context.Context) (*cobra.Command, *shared.ClientFactory) {
 	// updateNotification will check for an update in the background and print a message after the command runs
 	var updateNotification *update.UpdateNotification
 
-	clients = shared.NewClientFactory(shared.SetVersion(version.Raw()))
+	clients = shared.NewClientFactory()
 	rootCmd := NewRootCommand(clients, updateNotification)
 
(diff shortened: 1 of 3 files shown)


internal/style/style.go
```diff
@@ -273,6 +273,10 @@ func Underline(text string) string {
 	return render(lipgloss.NewStyle().Underline(true), text)
 }
 
+func Separator() string {
+	return " — "
```
mwbrooks: suggestion: Can we use the dash instead of the em-dash? The em-dash is a less common character on the terminal and it's becoming a bit of a AI-artifact-meme. I'd prefer that developers who copy & paste the terminal output receive a regular, plain, old dash 👴🏻 
```suggestion
	return " - "
```

mwbrooks: request: Could we add a unit test for this new `Separator()` function please?

PR #431, opened 2026-03-23
refactor: consolidate version access to slackcontext.Version(ctx)

mwbrooks:
### Changelog

- N/A

### Summary

This pull request consolidates version access to `slackcontext.Version(ctx)`.

Currently, the version is accessed with `version.Raw()`, `version.Get()`, `clients.config.Version`, and `slackcontext.Version(ctx)`. This is brittle, because it requires all version fields to be set to the same version. Additionally, the version value is passed as a parameter through some functions, which creates more risk of the incorrect version bring used.

- Migrates `version.Raw()` callers to `slackcontext.Version(ctx)`.
- Standardizes on `version.Raw()`, previously it was a mix - most used `.Raw()`, but the `recoveryFunc` used `.Get()`. This could lead to difficulty associating log traces.
- Removes `clients.config.Version`.
- Replaces `version string` parameter with `ctx` parameter.
- Simplifies mocks that required the version parameter.
- Version now set once in `main.go`.

### Open Questions

- [Please weigh-in on this question around whether to use `.Get()` or `.Raw()`.](https://github.com/slackapi/slack-cli/pull/431#discussion_r2978029180)

#### 1. Version output

```bash
$ ./bin/slack --version
# → Confirm: the version string is displayed correctly (e.g. v0.x.x or a dev build SHA)
```

#### 2. Version in debug output                                                                                                                             

```bash
$ ./bin/slack auth list --verbose --skip-update 2>&1 | grep -i version
# → Confirm: the CLI version appears in "Resolving logstash host"
# → Confirm: the CLI version appears in "FlushToLogstash"
# → Confirm no warnings about "version could not be found in context"
```

#### 3.. User-Agent in API debug output

```bash
$ ./bin/slack auth login --verbose --skip-update 2>&1 | grep "User-Agent"
# - Confirm: the CLI version appears in "HTTP Request User-Agent"
# CTRL+C
```

#### 4.. Log file version

```bash
$ ./bin/slack auth list
$ cat ~/.slack/logs/slack-debug-*.log | grep "Slack-CLI Version"
# → Confirm: last line in the log file contains the correct Slack-CLI Version value (not empty)
# → Example: "Slack-CLI Version: [v3.15.0]"
```

### Requirements

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -36,7 +36,7 @@ jobs:
       - name: Install Tests
         run: make test-install
       - name: Upload coverage to Codecov
-        uses: codecov/codecov-action@1af58845a975a7985b0beb0cbe6fbbb71a41dbad # v5.5.3
+        uses: codecov/codecov-action@57e3a136b779b570ffcdbf80b3bdc90e7fab3de2 # v6.0.0
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
           files: ./coverage.out


cmd/app/add_test.go
```diff
@@ -894,7 +894,7 @@ func prepareAddMocks(t *testing.T, clients *shared.ClientFactory, clientsMock *s
 
 	clientsMock.Auth.On("ResolveAPIHost", mock.Anything, mock.Anything, mock.Anything).
 		Return("api host")
-	clientsMock.Auth.On("ResolveLogstashHost", mock.Anything, mock.Anything, mock.Anything).
```
mwbrooks, line 897: note: This is the bulk of the changes - the version parameter was removed from the `ResolveLogstashHost`, which simplifies our testing mocks.

cmd/root.go
```diff
@@ -296,8 +296,6 @@ func InitConfig(ctx context.Context, clients *shared.ClientFactory, rootCmd *cob
 	// Init configurations
 	clients.Config.LoadExperiments(ctx, clients.IO.PrintDebug)
 	style.ToggleLipgloss(clients.Config.WithExperimentOn(experiment.Lipgloss))
-	// TODO(slackcontext) Consolidate storing CLI version to slackcontext
-	clients.Config.Version = version.Raw()
```
mwbrooks, line 300: 🪓 Removing the `TODO` to consolidate the version to `slackcontext`.

internal/auth/auth.go
```diff
@@ -508,7 +509,11 @@ func (c *Client) ResolveAPIHost(ctx context.Context, apiHostFlag string, customA
 
 // TODO: how does this play together with ResolveAPIHost above?
 // ResolveLogstashHost returns the error log stash host based on API Host and CLI version
-func (c *Client) ResolveLogstashHost(ctx context.Context, apiHost string, cliVersion string) string {
+func (c *Client) ResolveLogstashHost(ctx context.Context, apiHost string) string {
+	cliVersion, err := slackcontext.Version(ctx)
+	if err != nil {
+		c.io.PrintDebug(ctx, "Warning: %s", err.Error())
+	}
```
mwbrooks, line 516: note: `slackcontext.Version(ctx)` will return an error when `ctx` doesn't include the version field. This could happen if the function is called with `context.Background()` or `t.Context()`, which doesn't have the Slack CLI context properties. 

When an error happens, `slackcontext.Version(ctx)` falls back on `version.Raw()`, which is set by the compiler. So, it should be very reliable at returning the correct version even if the context was incorrectly passed into the function. However, still want to handle the error by printing an warning to the debug logs, so that it can be known and fixed.

internal/config/config.go
```diff
@@ -59,7 +59,6 @@ type Config struct {
 	TeamFlag                string
 	TokenFlag               string
 	NoColor                 bool
-	Version                 string
```
mwbrooks, line 62: 🪓 `shared.Config.Version`

internal/slackcontext/slackcontext.go
```diff
@@ -170,15 +171,16 @@ func SetSystemID(ctx context.Context, systemID string) context.Context {
 	return ctx
 }
 
-// Version returns the CLI version associated with `ctx`, or
-// `""` and `slackerror.ErrContextValueNotFound` if no version could be found.
+// Version returns the CLI version associated with `ctx`. If the version is not
+// found in the context, it falls back to version.Raw() and returns an error to
+// signal that the context was not properly initialized.
```
mwbrooks, line 176: note: The important design decision is written in this comment.

internal/tracking/tracking_test.go
```diff
@@ -128,8 +130,8 @@ func Test_Tracking_FlushToLogstash(t *testing.T) {
 			shouldNotSendRequest: true,
 		},
 		"should strip 'v' prefix from version string": {
-			setup: func(cfg *config.Config) {
-				cfg.Version = "v4.2.0"
+			ctxSetup: func(ctx context.Context) context.Context {
+				return slackcontext.SetVersion(ctx, "v4.2.0")
```
mwbrooks, line 134: note: This is a safer mock now :)

main.go
```diff
@@ -49,7 +49,7 @@ func main() {
 
 	// Set context values
 	sessionID := uuid.New().String()
-	cliVersion := version.Get()
+	cliVersion := version.Raw()
```
mwbrooks, line 52: 💬 discussion: This is an important decision that we should feel confident in.

Originally, the version used by `slackcontext` and log traces was `version.Get()` (always prefixes the 'v'). But, Logstash and API calls used `version.Raw()` (git tag, which has the 'v' by convention but not guaranteed).

This wasn't a problem because our git tags always have a prefixed 'v' (e.g. `v3.15.0`). However, if we created a git tag without the 'v' then version would be different in different places.

:two-cents: _Consistency is more important than format._ This PR does introduce the consistency, I hope, because we consolidate the version access through `slackcontext.Version(ctx)`. w.r.t the version format, I went with `version.Raw()` so that we know our git tag is the source of truth. If someone creates a git tag with a different format, at least we know it all lines up. But, I'm open to either.

Thoughts?

main.go
```diff
@@ -81,13 +81,13 @@ func recoveryFunc() {
 	// in the event of a panic, log panic
 	if r := recover(); r != nil {
 		var clients = shared.NewClientFactory()
-		clients.Config.Version = version.Raw()
 
 		var ctx = context.Background()
 		ctx = slackcontext.SetSessionID(ctx, uuid.New().String())
+		ctx = slackcontext.SetVersion(ctx, version.Raw())
```
mwbrooks, line 87: note: Cannot re-use the `cliVersion` above because this is the `recovery()` function.

mwbrooks (commented):
Comments for the wise reviewers!

cmd/app/add_test.go
zimeg: 🌟 praise: @mwbrooks Amazing use of the `slackcontext` values for gathering this now!

```diff
- ResolveLogstashHost(ctx, clients.Config.APIHostResolved, clients.Config.Version)
+ ResolveLogstashHost(ctx, clients.Config.APIHostResolved)
```

main.go
srtaalej: I agree with using `Raw()` and i like the idea of having git tag as our source of truth! it means its less likely for our logstash, api calls, etc to have the wrong version. also if we did mistakenly make a git tag w/out the 'v', that version would be consistent everywhere else so it would be easier to correct

internal/auth/auth.go
srtaalej: falliing back on `version.Raw()` is nice!

internal/tracking/tracking.go
```diff
@@ -168,7 +168,11 @@ func (e *EventTracker) FlushToLogstash(ctx context.Context, cfg *config.Config,
 		// In this case, the root command was not initialized, so none of the bootup routine executed (flags weren't parsed, config not initialized, etc.).
 		return nil
 	}
-	versionString, _ := strings.CutPrefix(cfg.Version, "v")
+	cliVersion, err := slackcontext.Version(ctx)
+	if err != nil {
+		ioStream.PrintDebug(ctx, "Warning: %s", err.Error())
+	}
+	versionString, _ := strings.CutPrefix(cliVersion, "v")
```
zimeg, line 175:
```suggestion
	versionString := strings.TrimPrefix(cliVersion, "v")
```

🔭 suggestion(non-blocking): The unhandled "found" value might be more confusing than a different method here? No blocker since that's unchanged from before but it might be a good time to improve this for later.

🔗 https://pkg.go.dev/strings#TrimPrefix

main.go
zimeg: 💡 praise: Thanks for calling this out! I'd been stumped on what might be best too.

👾 thought: I was leaning toward using `version.Get` to guarantee that releases use the "v" prefix to avoid errors with the `User-Agent` format:

```
slack-cli/v3.4.5-example-2-ga7e71fe (os: darwin)
```

But agree that consistent formats is better! We might lean on CI to package releases to avoid nuance as well 🎁 

🔭 question: Should we also remove `version.Get` in a follow up PR? It might be a chance to ask less from the `regex` package:

https://github.com/slackapi/slack-cli/blob/2cf691e1e0cc9b4ef09b22d57b6a2f56bfb7f60a/internal/version/version.go#L43-L50

mwbrooks: Thank you for sharing your thoughts @srtaalej and @zimeg! It makes me feel more confident that we are all agreeing that the git tag should be the source of truth.

> 🔭 question: Should we also remove version.Get in a follow up PR? It might be a chance to ask less from the regex package:

Mind-meld! I was going to suggest the same. I think the above discussion is a by-product of having 2 ways to "get" the version - `.Raw()` and `.Get()`. I was going to suggest that we either remove `.Get()` and make the git tag the true source of truth.

I think we will need a discussion on how to enforce our version format: semver with a 'v' prefix. But let's take that offline for now.

PR #433, opened 2026-03-24
feat: add slack docs search subcommand 

lukegalbraithrussell:
### Changelog

Adds `slack docs search` subcommand. It can output search results as plain text, json, or kick ya to the browser

### Summary

`slack docs search`

| Option | Type | Default | Description | Required |
|--------|------|---------|-------------|---|
| `<query>` | string |  | Search terms | Required
| `--output` | string | `text` | Output format: `text' , 'browser` or `json` | Optional
| `--limit` | int | `20` | Max number of results to return (JSON and text output only) | Optional

### Examples

```bash
# Search and return plain text results
$ slack-cli docs search "Block Kit"

# Search and return JSON results with custom result limit
$ slack-cli docs search "api" --output=json --limit=5

# Search and open results in browser
$ slack-cli docs search "webhooks" --output=browser
```

### Requirements

- [X] I've read and understood the [Contributing Guidelines](https://github.com/slackapi/slack-cli/blob/main/.github/CONTRIBUTING.md) and have done my best effort to follow them.
- [X] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).

### DIFF
--- a/cmd/project/create_template.go
+++ b/cmd/project/create_template.go
@@ -144,7 +144,7 @@ func promptTemplateSelection(cmd *cobra.Command, clients *shared.ClientFactory, 
 		}
 	} else {
 		// Prompt for the category
-		promptForCategory := "Select an app:"
+		promptForCategory := "Select a category:"
 		optionsForCategory := getSelectionOptionsForCategory(clients)
 		titlesForCategory := make([]string, len(optionsForCategory))
 		for i, m := range optionsForCategory {
@@ -196,7 +196,7 @@ func promptTemplateSelection(cmd *cobra.Command, clients *shared.ClientFactory, 
 	}
 
 	// Prompt for the example template
-	prompt := "Select a language:"
+	prompt := "Select a framework:"
 	if categoryID == "slack-cli#ai-apps" {
 		prompt = "Select a template:"
 	}
@@ -229,7 +229,7 @@ func promptTemplateSelection(cmd *cobra.Command, clients *shared.ClientFactory, 
 	for i, opt := range examples {
 		choices[i] = opt.Title
 	}
-	choice, err := clients.IO.SelectPrompt(ctx, "Select a language:", choices, iostreams.SelectPromptConfig{
+	choice, err := clients.IO.SelectPrompt(ctx, "Select a framework:", choices, iostreams.SelectPromptConfig{
 		Description: func(value string, index int) string {
 			return examples[index].Description
 		},
--- a/cmd/project/create_test.go
+++ b/cmd/project/create_test.go
@@ -46,7 +46,7 @@ func TestCreateCommand(t *testing.T) {
 		"creates a bolt application from prompts": {
 			Setup: func(t *testing.T, ctx context.Context, cm *shared.ClientsMock, cf *shared.ClientFactory) {
 				cm.IO.On("IsTTY").Return(true)
-				cm.IO.On("SelectPrompt", mock.Anything, "Select an app:", mock.Anything, mock.Anything).
+				cm.IO.On("SelectPrompt", mock.Anything, "Select a category:", mock.Anything, mock.Anything).
 					Return(
 						iostreams.SelectPromptResponse{
 							Prompt: true,
@@ -54,7 +54,7 @@ func TestCreateCommand(t *testing.T) {
 						},
 						nil,
 					)
-				cm.IO.On("SelectPrompt", mock.Anything, "Select a language:", mock.Anything, mock.Anything).
+				cm.IO.On("SelectPrompt", mock.Anything, "Select a framework:", mock.Anything, mock.Anything).
 					Return(
 						iostreams.SelectPromptResponse{
 							Prompt: true,
@@ -83,7 +83,7 @@ func TestCreateCommand(t *testing.T) {
 			CmdArgs: []string{"--template", "slack-samples/deno-starter-template"},
 			Setup: func(t *testing.T, ctx context.Context, cm *shared.ClientsMock, cf *shared.ClientFactory) {
 				cm.IO.On("IsTTY").Return(true)
-				cm.IO.On("SelectPrompt", mock.Anything, "Select an app:", mock.Anything, mock.Anything).
+				cm.IO.On("SelectPrompt", mock.Anything, "Select a category:", mock.Anything, mock.Anything).
 					Return(
 						iostreams.SelectPromptResponse{
 							Flag:   true,
@@ -91,7 +91,7 @@ func TestCreateCommand(t *testing.T) {
 						},
 						nil,
 					)
-				cm.IO.On("SelectPrompt", mock.Anything, "Select a language:", mock.Anything, mock.Anything).
+				cm.IO.On("SelectPrompt", mock.Anything, "Select a framework:", mock.Anything, mock.Anything).
 					Return(
 						iostreams.SelectPromptResponse{
 							Flag:   true,
@@ -129,7 +129,7 @@ func TestCreateCommand(t *testing.T) {
 						},
 						nil,
 					)
-				cm.IO.On("SelectPrompt", mock.Anything, "Select a language:", mock.Anything, mock.Anything).
+				cm.IO.On("SelectPrompt", mock.Anything, "Select a framework:", mock.Anything, mock.Anything).
 					Return(
 						iostreams.SelectPromptResponse{
 							Prompt: true,
@@ -152,7 +152,7 @@ func TestCreateCommand(t *testing.T) {
 				}
 				createClientMock.AssertCalled(t, "Create", mock.Anything, mock.Anything, expected)
 				// Verify that category prompt was NOT called
-				cm.IO.AssertNotCalled(t, "SelectPrompt", mock.Anything, "Select an app:", mock.Anything, mock.Anything)
+				cm.IO.AssertNotCalled(t, "SelectPrompt", mock.Anything, "Select a category:", mock.Anything, mock.Anything)
 				cm.IO.AssertCalled(t, "InputPrompt", mock.Anything, "Name your app:", mock.Anything)
 			},
 		},
@@ -168,7 +168,7 @@ func TestCreateCommand(t *testing.T) {
 						},
 						nil,
 					)
-				cm.IO.On("SelectPrompt", mock.Anything, "Select a language:", mock.Anything, mock.Anything).
+				cm.IO.On("SelectPrompt", mock.Anything, "Select a framework:", mock.Anything, mock.Anything).
 					Return(
 						iostreams.SelectPromptResponse{
 							Prompt: true,
@@ -189,7 +189,7 @@ func TestCreateCommand(t *testing.T) {
 				}
 				createClientMock.AssertCalled(t, "Create", mock.Anything, mock.Anything, expected)
 				// Verify that category prompt was NOT called
-				cm.IO.AssertNotCalled(t, "SelectPrompt", mock.Anything, "Select an app:", mock.Anything, mock.Anything)
+				cm.IO.AssertNotCalled(t, "SelectPrompt", mock.Anything, "Select a category:", mock.Anything, mock.Anything)
 				// Verify that name prompt was NOT called since name was provided as arg
 				cm.IO.AssertNotCalled(t, "InputPrompt", mock.Anything, "Name your app:", mock.Anything)
 			},
@@ -197,11 +197,11 @@ func TestCreateCommand(t *testing.T) {
 		"creates a pydantic ai agent app": {
 			CmdArgs: []string{"my-pydantic-app"},
 			Setup: func(t *testing.T, ctx context.Context, cm *shared.ClientsMock, cf *shared.ClientFactory) {
-				cm.IO.On("SelectPrompt", mock.Anything, "Select an app:", mock.Anything, mock.Anything).
+				cm.IO.On("SelectPrompt", mock.Anything, "Select a category:", mock.Anything, mock.Anything).
 					Return(iostreams.SelectPromptResponse{Prompt: true, Index: 1}, nil)
 				cm.IO.On("SelectPrompt", mock.Anything, "Select a template:", mock.Anything, mock.Anything).
 					Return(iostreams.SelectPromptResponse{Prompt: true, Index: 0}, nil)
-				cm.IO.On("SelectPrompt", mock.Anything, "Select a language:", mock.Anything, mock.Anything).
+				cm.IO.On("SelectPrompt", mock.Anything, "Select a framework:", mock.Anything, mock.Anything).
 					Return(iostreams.SelectPromptResponse{Prompt: true, Index: 2}, nil)
 				createClientMock = new(CreateClientMock)
 				createClientMock.On("Create", mock.Anything, mock.Anything, mock.Anything).Return("", nil)
@@ -222,7 +222,7 @@ func TestCreateCommand(t *testing.T) {
 		"creates an app named agent when template flag is provided": {
 			CmdArgs: []string{"agent", "--template", "slack-samples/bolt-js-starter-template"},
 			Setup: func(t *testing.T, ctx context.Context, cm *shared.ClientsMock, cf *shared.ClientFactory) {
-				cm.IO.On("SelectPrompt", mock.Anything, "Select an app:", mock.Anything, mock.Anything).
+				cm.IO.On("SelectPrompt", mock.Anything, "Select a category:", mock.Anything, mock.Anything).
 					Return(
 						iostreams.SelectPromptResponse{
 							Flag:   true,
@@ -230,7 +230,7 @@ func TestCreateCommand(t *testing.T) {
 						},
 						nil,
 					)
-				cm.IO.On("SelectPrompt", mock.Anything, "Select a language:", mock.Anything, mock.Anything).
+				cm.IO.On("SelectPrompt", mock.Anything, "Select a framework:", mock.Anything, mock.Anything).
 					Return(
 						iostreams.SelectPromptResponse{
 							Flag:   true,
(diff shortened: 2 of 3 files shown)


cmd/docs/search.go
```diff
@@ -0,0 +1,136 @@
...
+	"os"
+	"strings"
+
+	"github.com/slackapi/slack-cli/internal/shared"
+	"github.com/slackapi/slack-cli/internal/slacktrace"
+	"github.com/slackapi/slack-cli/internal/style"
+	"github.com/spf13/cobra"
+)
+
+const docsSearchAPIURL = "https://docs-slack-d-search-api-duu9zr.herokuapp.com/api/search"
```
lukegalbraithrussell: This will be updated to docs.slack.dev once [api endpoint PR is merged](https://github.com/slackapi/docs/pull/462) in private docs repo

cmd/docs/search.go
```diff
@@ -0,0 +1,136 @@
...
+		},
+	}))
+
+	clients.Browser().OpenURL(docsURL)
+	clients.IO.PrintTrace(ctx, slacktrace.DocsSearchSuccess, query)
+
+	return nil
+}
+
+func fetchAndOutputSearchResults(ctx context.Context, clients *shared.ClientFactory, query string, limit int, httpClient *http.Client) error {
```
srtaalej: it'd be nice if URLs were full https://docs.slack.dev URLs so they're `cmd+clickable` in the terminal
```json
{
      "url": "https://docs.slack.dev/block-kit/",
      "title": "Block Kit"
    },
```

right now the json outputs as
```json
 {
      "url": "/block-kit/",
      "title": "Block Kit"
    },
```

cmd/docs/search_test.go
```diff
@@ -0,0 +1,260 @@
...
+// Verifies that various query formats are properly URL encoded and API parameters are correctly passed.
+func Test_Docs_SearchCommand_JSONOutput_QueryFormats(t *testing.T) {
+	mockResponse := `{
+		"total_results": 2,
+		"limit": 20,
+		"results": [
+			{
+				"title": "Block Kit",
+				"url": "https://docs.slack.dev/block-kit"
+			},
```
srtaalej: these tests are using full URLs in mocks but the real API returns relative paths —  so the tests aren't catching this. a simple fix would be to append to the api output. 
something like `"https://docs.slack.dev" + result.URL`

cmd/docs/docs.go
```diff
@@ -32,27 +32,33 @@ func NewCommand(clients *shared.ClientFactory) *cobra.Command {
...
+		Args: cobra.ArbitraryArgs, // Allow any arguments
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return runDocsCommand(clients, cmd, args)
 		},
+		// Disable automatic suggestions for unknown commands
+		DisableSuggestions: true,
 	}
 
-	cmd.Flags().BoolVar(&searchMode, "search", false, "open Slack docs search page or search with query")
+	cmd.Flags().BoolVar(&searchMode, "search", false, "[DEPRECATED] open Slack docs search page or search with query (use 'docs search' subcommand instead)")
```
zimeg:
```suggestion
	// DEPRECATED(semver:major): Remove in favor of "search" command
	cmd.Flags().BoolVar(&searchMode, "search", false, "open Slack docs search page or search with query")
	cmd.Flag("search").Hidden = true
```

🔭 suggestion: Let's prefer to hide deprecated features!

cmd/docs/search.go
zimeg: 📚 thought: I agree listing entire links is ideal and might suggest we add a `text` format with this?

```sh
$ slack docs search "Block Kit" --output=text
```

👾 ramble: If this format is supported it seems awkward to write that flag instead of defaulting to it but this is personal preference I think!

cmd/docs/search.go
```diff
@@ -0,0 +1,136 @@
...
+
+func runDocsSearchCommand(clients *shared.ClientFactory, cmd *cobra.Command, args []string, cfg *searchConfig, httpClient *http.Client) error {
+	ctx := cmd.Context()
+
+	query := strings.Join(args, " ")
+
+	if cfg.output == "json" {
+		return fetchAndOutputSearchResults(ctx, clients, query, cfg.limit, httpClient)
+	}
+
```
zimeg:
```suggestion
	if cfg.output != "browser" {
		return slackerror.New(slackerror.ErrInvalidFlag)
	}

```

👻 issue: We should error if an unexpected flag is provided instead of the next steps!

📣 ramble: We have formatting to change the message and remediation too IIRC with `WithMessage` or with `WithRemediation` attached to the `New` error, if wording more might be right? This is rambles.

cmd/docs/search.go
zimeg:
```suggestion
const docsURL = "https://docs-slack-d-search-api-duu9zr.herokuapp.com/"
```

🪬 suggestion: We might want to abstract this to the `docs/docs.go` file since we target it in multiple places? It might be useful for testing changes too?

cmd/docs/search.go
```diff
@@ -0,0 +1,136 @@
...
+	return nil
+}
+
+func fetchAndOutputSearchResults(ctx context.Context, clients *shared.ClientFactory, query string, limit int, httpClient *http.Client) error {
+	apiURL := fmt.Sprintf("%s?q=%s&limit=%d", docsSearchAPIURL, url.QueryEscape(query), limit)
+
+	resp, err := httpClient.Get(apiURL)
+	if err != nil {
+		return fmt.Errorf("failed to fetch search results: %w", err)
```
zimeg:
```suggestion
		return slackerror.New(slackerror.ErrHTTPRequestFailed).WithRootCause(err)
```

⚠️ suggestion: We can explain errors with existing error codes!

cmd/docs/search.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("API returned status %d", resp.StatusCode)
```
zimeg:
```suggestion
		return slackerror.New(slackerror.ErrHTTPRequestFailed).WithMessage("API returned status %d", resp.StatusCode)
```

🧰 suggestion: This might be the right format for similar changes to use error codes?

cmd/docs/search.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	var searchResponse DocsSearchResponse
+	if err := json.NewDecoder(resp.Body).Decode(&searchResponse); err != nil {
+		return fmt.Errorf("failed to parse search results: %w", err)
```
zimeg:
```suggestion
		return slackerror.New(slackerror.ErrHTTPResponseInvalid).WithRootCause(err)
```

👁️‍🗨️ suggestion: This might be alright but let's hope against such case!

cmd/docs/search.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	encoder := json.NewEncoder(os.Stdout)
+	encoder.SetIndent("", "  ")
+	if err := encoder.Encode(searchResponse); err != nil {
+		return fmt.Errorf("failed to output search results: %w", err)
+	}
```
zimeg:
```suggestion
	output, err := json.MarshalIndent(searchResponse, "", "  ")
	if err != nil {
		return slackerror.JSONUnmarshalError(err, searchResponse)
	}
	clients.IO.PrintInfo(ctx, false, "%s", string(output))
```

🔍 suggestion: Forbid we find this happen but let's output in formats that include response detail-

cmd/docs/search.go
```diff
@@ -0,0 +1,136 @@
...
+				Command: "docs search \"webhooks\" --output=browser",
+			},
+			{
+				Meaning: "Search docs with limited JSON results",
+				Command: "docs search \"api\" --output=json --limit=5",
+			},
+		}),
+		Args: cobra.MinimumNArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runDocsSearchCommand(clients, cmd, args, cfg, http.DefaultClient)
```
zimeg: 🌟 praise: I'm a fan of the `cfg` approach for stable flag values. I'm curious if this can be explored more in a `slackcommand` follow up. This is cool!

cmd/docs/search.go
```diff
@@ -0,0 +1,136 @@
...
+
+	cmd.Flags().StringVar(&cfg.output, "output", "json", "output format: browser, json")
+	cmd.Flags().IntVar(&cfg.limit, "limit", 20, "maximum number of search results to return (only applies with --output=json)")
+
+	return cmd
+}
+
+func runDocsSearchCommand(clients *shared.ClientFactory, cmd *cobra.Command, args []string, cfg *searchConfig, httpClient *http.Client) error {
+	ctx := cmd.Context()
+
```
zimeg, line 85:
```suggestion
	client := api.NewClient(httpClient, docsURL, clients.IO)
```

📺 suggestion: We might want to use the `internal/api/client` and adjacent paths for these **search** methods to take advantage of the error handling. A lot of comments are related to this and I realize the API client does much more:

https://github.com/slackapi/slack-cli/blob/2cf691e1e0cc9b4ef09b22d57b6a2f56bfb7f60a/internal/api/client.go#L209-L261

zimeg: 🧪 ramble: These patterns might also make testing more standard to avoid mixing API and command checks I hope!

zimeg (changes requested):
@lukegalbraithrussell This is an awesome feature to be bringing 📚 ✨ 

I'm leaving a handful of comments that do request changes, but I'm hoping we can work through these:

- The API method logic is included with command: I understand we reach a different host but we might still find `api` package useful.
- Defaulting outputs for the person reading: I'm a fan of JSON but find this jarring when hoping for a quick search in terminal. I instead suggest a default `text` output to use section formatting:
```sh
$ slack docs search "Block Kit"
```

⚠️ Edit: I submit too soon! WIP! WIP!

cmd/docs/search.go
lukegalbraithrussell: like?:

```
default:
	return slackerror.New(slackerror.ErrInvalidFlag).WithMessage(
		fmt.Sprintf("Invalid output format: %s", cfg.output),
	).WithRemediation(
		"Use one of: text, json, browser",
	)
```

zimeg: @lukegalbraithrussell Hmm I think the `if` might be alright for now but if changed to `cases` this seems solid!

🧵 FWIW the `fmt.Sprintf` wouldn't be needed IIRC

cmd/docs/docs.go
```diff
@@ -26,42 +26,55 @@ import (
...
 )
 
+const docsURL = "https://docs.slack.dev"
+
 var searchMode bool
 
+func buildDocsSearchURL(query string) string {
+	encodedQuery := url.QueryEscape(query)
+	return fmt.Sprintf("%s/search/?q=%s", docsURL, encodedQuery)
+}
```
zimeg: 🎁 suggestion(non-blocking): I understand this is shared between this file and `search.go` but perhaps it'd be alright to move to the search file?

cmd/docs/docs.go
```diff
@@ -26,42 +26,55 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const docsURL = "https://docs.slack.dev"
```
zimeg, line 24: 🔗 praise: This is a nice constant to share I'll claim! Global variables sometimes can give me stress...

cmd/docs/search.go
```diff
@@ -0,0 +1,144 @@
...
+	return docsURL + relativeURL
+}
+
+func NewSearchCommand(clients *shared.ClientFactory) *cobra.Command {
+	cfg := &searchConfig{}
+
+	cmd := &cobra.Command{
+		Use:   "search <query>",
+		Short: "Search Slack developer docs",
+		Long:  "Search the Slack developer docs and return results in text, JSON, or browser format",
```
zimeg:
```suggestion
		Long:  strings.Join([]string{
			"Search the Slack developer docs and return results in text, JSON, or browser",
			"format.",
		}, "\n"),
```

📚 suggestion: We're still earliest times for the formatting guide but let's start making these `long` descriptions full sentence?

cmd/docs/search.go
zimeg: 🔬 note: Thanks for also including these as now separate variables:

https://github.com/slackapi/slack-cli/blob/671b4c8d348391b7c56f3841614c8e93b4fc7e30/cmd/docs/docs.go#L29

https://github.com/slackapi/slack-cli/blob/671b4c8d348391b7c56f3841614c8e93b4fc7e30/internal/api/docs.go#L27

cmd/docs/docs.go
zimeg: 🪓 note: We might now consider breaking this for the next update. I'm starting to think removing these fallbacks can be alright in this PR? @lukegalbraithrussell

cmd/docs/docs.go
zimeg: 👾 ramble: If we do decide to remove the "--search" flag let's also perhaps inline this!

zimeg: 🔬 note: With these changes let's update remediation for an invalid `docs` command to suggest the "search" command also:

```
$ slack docs apps

🚫 Invalid docs command. Did you mean to search? (docs_search_flag_required)

💡 Suggestion
   Use --search flag: slack docs --search "apps"

```

cmd/docs/search.go
```diff
@@ -0,0 +1,144 @@
...
+		return relativeURL
+	}
+	return docsURL + relativeURL
+}
+
+func NewSearchCommand(clients *shared.ClientFactory) *cobra.Command {
+	cfg := &searchConfig{}
+
+	cmd := &cobra.Command{
+		Use:   "search <query>",
```
zimeg:
```suggestion
		Use:   "search [query]",
```

🚕 suggestion: For optional arguments we can use square braces!

cmd/docs/search.go
```diff
@@ -0,0 +1,144 @@
...
+func runDocsSearchCommand(clients *shared.ClientFactory, cmd *cobra.Command, args []string, cfg *searchConfig) error {
+	ctx := cmd.Context()
+
+	query := strings.Join(args, " ")
+
+	switch cfg.output {
+	case "json":
+		return fetchAndOutputSearchResults(ctx, clients, query, cfg.limit)
+	case "text":
+		return fetchAndOutputTextResults(ctx, clients, query, cfg.limit)
```
zimeg: 🪩 quibble: I'd support inlining these functions to avoid jumping around the page but not a blocker!

cmd/docs/search.go
```diff
@@ -0,0 +1,144 @@
...
+
+func fetchAndOutputTextResults(ctx context.Context, clients *shared.ClientFactory, query string, limit int) error {
+	searchResponse, err := clients.API().DocsSearch(ctx, query, limit)
+	if err != nil {
+		return err
+	}
+
+	for _, result := range searchResponse.Results {
+		absoluteURL := makeAbsoluteURL(result.URL)
+		fmt.Fprintf(clients.IO.WriteOut(), "%s\n%s\n\n", result.Title, absoluteURL)
```
zimeg:
```suggestion
		clients.IO.PrintInfo(ctx, false, style.Sectionf(style.TextSection{
			Emoji:     "books",
			Text:      result.Title,
			Secondary: []string{absoluteURL},
		}))
```

🎨 suggestion: Let's use sections for these outputs! I'm unsure of this format but it'd be similar to the output of `samples` command perhaps?

Image

cmd/docs/search.go
```diff
@@ -0,0 +1,144 @@
...
+			{
+				Meaning: "Search docs and open results in browser",
+				Command: "docs search \"webhooks\" --output=browser",
+			},
+			{
+				Meaning: "Search docs with limited JSON results",
+				Command: "docs search \"api\" --output=json --limit=5",
+			},
+		}),
+		Args: cobra.MinimumNArgs(1),
```
zimeg, line 71:
```suggestion
		Args: cobra.MinimumNArgs(0),
```

👾 issue: We might want to default to opening the browser if no arguments are provided?

cmd/docs/search.go
```diff
@@ -0,0 +1,144 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runDocsSearchCommand(clients, cmd, args, cfg)
+		},
+	}
+
+	cmd.Flags().StringVar(&cfg.output, "output", "text", "output format: text, json, browser")
+	cmd.Flags().IntVar(&cfg.limit, "limit", 20, "maximum number of search results to return (only applies with --output=json and --output=text)")
```
zimeg:
```suggestion
	cmd.Flags().IntVar(&cfg.limit, "limit", 20, "maximum number of text or json search results to return")
```

🔭 suggestion: Since this is output alongside `output` we might inline nuance?

cmd/docs/search.go
lukegalbraithrussell: @zimeg I'm 50/50 on this. if you forget the argument you get yanked into a browser which feels maybe rude?

zimeg: @lukegalbraithrussell Apologies for a slow response back! I expected the browser when testing but agree that it's jarring and that we should keep this as an error for now.

👾 ramble: Avoiding errors with different remediation does keep me curious but let's follow what's giving us such friction next!

internal/api/docs.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func buildDocsSearchURL(baseURL, query string, limit int) (string, error) {
+	endpoint := fmt.Sprintf("%s?query=%s&limit=%d", docsSearchMethod, url.QueryEscape(query), limit)
+	sURL, err := url.Parse(baseURL + "/" + endpoint)
+	if err != nil {
+		return "", err
+	}
+	return sURL.String(), nil
+}
```
zimeg, line 55: 🪬 thought: Would it make downstream calls easier if we return a URL with this function?

```diff
- string
+ *url.URL
```

zimeg: 👾 ramble: I'm hesitant now to separate these functions for sake of testing specific errors in HTTP constructions.

thought: This was alright but I notice now that the `buildDocsSearchURL` function is duplicated and different between this package and the command too - it's unclear what's preferred so I think the inlined approach keeps it simple.

internal/api/docs.go
```diff
@@ -0,0 +1,103 @@
...
+	"net/url"
+	"runtime"
+
+	"github.com/opentracing/opentracing-go"
+	"github.com/slackapi/slack-cli/internal/slackcontext"
+)
+
+var docsBaseURL = "https://docs.slack.dev"
+
+const docsSearchMethod = "api/v1/search"
```
zimeg, line 31: 🌟

internal/api/docs.go
lukegalbraithrussell: @zimeg if i'm understanding this, ultimately not worth pulling out rn? we can revisit in the future as well

PR #436, opened 2026-03-24
feat: support loading a dotenv '.env' file for your app

zimeg:
### Changelog

> Environment variables saved to a `.env` file are now loaded into the process before hook scripts run. This brings meaningful improvement to the `slack run` command foremost as additional dependencies are no longer required for loading environment variables into the runtime context.
>
> Other commands, such as the `slack deploy` command and the `slack manifest` command, also benefit from this since the underlying hooks can be more expressive in customizations and configured to custom environments.

### Summary

This PR loads environment variables from the `.env` file for hook commands. These variables are loaded before each hook command executes to ensure correctness 🌲 ✨ 

We also fix orderings of environment variable precedence within hooks to be:

1. Provided variables to hooks - Ex: The `SLACK_CLI_XOXP` token provided with the `run` command.
2. Saved ".env" file - Ex: The developer's project keeps a `.env`. file.
3. Existing shell environment - Ex: Prior `EXPORT` or set variables otherwise. This is most important!

### Preview

https://github.com/user-attachments/assets/c0e25104-dd5c-4ef5-95ef-b097948c3270

### Reviewers

A few test cases might be interesting to saved variables:

```sh
$ slack create asdf -t slack-samples/bolt-js-assistant-template
$ cd asdf
$ vim .env
OPENAI_API_KEY=sk-proj-example-123  # Replace please!
$ npm uninstall dotenv
$ diff app.js
- import 'dotenv/config';           # Delete this line
$ slack run                         # File is read
$ OPENAI_API_KEY=oopsies slack run  # Shell overrides
```

### Notes

We might follow up with similar `.env` patterns and I'm hoping these changes guide decent direction to placement of loading environment variables for hooks!

- Earlier exploration attempted to read environment variables during the CLI setup but that required configuration changes that seemed incorrect to make. I understand now the [`internal/config/dotenv`](https://github.com/slackapi/slack-cli/blob/main/internal/config/dotenv.go) package is for internal configurations instead of developer application variables.
- The existing [`clients.Config.ManifestEnv`](https://github.com/slackapi/slack-cli/blob/a3187a9bfe0c35f84009a6a2ed71f7824d230932/internal/config/config.go#L70) attribute is used for `manifest` and `trigger` commands but not `run` or the `deploy` commands for **Bolt** apps. I'm unsure that we should continue to support [this](https://github.com/slackapi/slack-cli/blob/a3187a9bfe0c35f84009a6a2ed71f7824d230932/internal/app/manifest.go#L41-L56) as we bring enhancement to a project ".env" file itself? Regardless, it wasn't the right rabbit hole...

### Requirements


internal/pkg/platform/localserver.go
zimeg, line 1: 🗣️ note: The `run` command uses a separate `hook` process to handle automatic restarts with file watching so we duplicate some logic here!

cmd/platform/deploy.go
```diff
@@ -181,6 +181,7 @@ func deployHook(ctx context.Context, clients *shared.ClientFactory) error {
 	// so we instantiate the default here.
 	shell := hooks.HookExecutorDefaultProtocol{
 		IO: clients.IO,
+		Fs: clients.Fs,
```
zimeg, line 184: 🔭 note: Standalone protocol setups must define `fs` to access `.env` but we don't error otherwise. AFAICT this is required for just the `deploy` and `run` commands.

internal/hooks/hook_executor_default_test.go
```diff
@@ -80,6 +81,31 @@ func Test_Hook_Execute_Default_Protocol(t *testing.T) {
 				require.Contains(t, opts.Exec.(*MockExec).mockCommand.Env, `YIN=yang`)
 			},
 		},
+		"dotenv vars are loaded": {
```
mwbrooks: suggestion: Very nice test!

With my head in the weeds of how this all works, I appreciate that you're testing both session environment variables (`Env:`) and dotenv variables (`.env`).

My concern is that our future selves (or teammates) but not pick up on this nuance.

Perhaps a comment would help explain that we're testing both sources of environment variables?

internal/hooks/hook_executor_v2_test.go
```diff
@@ -132,6 +134,29 @@ func Test_Hook_Execute_V2_Protocol(t *testing.T) {
 				)
 			},
 		},
+		"dotenv vars are loaded": {
```
mwbrooks: note: If we decide to add a comment to the table tests for the default hook executor, I suppose we should do it here as well.

internal/hooks/hooks.go
```diff
@@ -53,13 +73,39 @@ func processExecOpts(opts HookExecOpts) ([]string, []string, []string, error) {
...
+	// Order of precedence from lowest to highest:
+	// 1. Provided "opts.Env" variables
+	// 2. Saved ".env" file
+	// 3. Existing shell environment
+	//
+	// > Each entry is of the form "key=value".
+	// > ...
+	// > If Env contains duplicate environment keys, only the last value in the slice for each duplicate key is used.
+	//
+	// https://pkg.go.dev/os/exec#Cmd.Env
```
mwbrooks: praise: ❤️ 🖊️  Love the detailed comment for future readers!

internal/hooks/hooks_test.go
mwbrooks: suggestion: We improve `processExecOpts` with the loading order of the env vars (opts → .env → session). I think we should add tests that verify that each of this overrides work and that `.env` does not override the session env variables.

internal/pkg/platform/localserver.go
```diff
@@ -308,13 +308,39 @@ func (r *LocalServer) StartDelegate(ctx context.Context) error {
...
+	// Order of precedence from lowest to highest:
+	// 1. Provided "opts.Env" variables
+	// 2. Saved ".env" file
+	// 3. Existing shell environment
+	//
+	// > Each entry is of the form "key=value".
+	// > ...
+	// > If Env contains duplicate environment keys, only the last value in the slice for each duplicate key is used.
+	//
+	// https://pkg.go.dev/os/exec#Cmd.Env
```
mwbrooks: suggestion: This logic feels important and since it's used in 2 places, I think we should consider DRY'ing up the code and putting the logic into 1 place. For example, `internal/config/dotenv.go` (existing) or rename the file to `internal/slackdotenv` if we want something that doesn't name collide with the dotenv dependency.

Additionally, a single function would make future improvements easier. For example, we may want to introduce a `--dotenv-overwrite` flag and `{ "dotenv-overwrite": true` config value that allow the `.env` to overwrite session variables. This seems to be a common use-case because most dotenv libraries support it, including our package with `godotenv.Overload()`.

Note: This would also allow us to unit test the scenarios in `internal/config/dotenv_test.go` instead of in the `localserver_test.go` and `hooks_test.go`.

internal/hooks/hooks.go
```diff
@@ -53,13 +73,39 @@ func processExecOpts(opts HookExecOpts) ([]string, []string, []string, error) {
 	var cmdArgVars = cmdArgs[1:] // omit the first item because that is the command name
 	cmdArgVars = append(cmdArgVars, goutils.MapToStringSlice(opts.Args, "--")...)
 
+	// Load .env file variables
+	dotEnv, err := LoadDotEnv(fs)
+	if err != nil {
+		io.PrintDebug(ctx, "Warning: failed to parse .env file: %s", err)
```
mwbrooks: question: Should we be more noisy about this warning? 

The scenario that I think about is that I have a malformed `.env`. The warning is printed into debug, which I don't see. Then I spend a long time debugging my app, wondering why it's not working until I realize that the environment variables are never loaded.

internal/hooks/hooks.go
```diff
@@ -53,13 +73,39 @@ func processExecOpts(opts HookExecOpts) ([]string, []string, []string, error) {
...
+	}
+	if len(dotEnv) > 0 {
+		keys := make([]string, 0, len(dotEnv))
+		for k := range dotEnv {
+			keys = append(keys, k)
+		}
+		io.PrintDebug(ctx, "loaded variables from .env file: %s", strings.Join(keys, ", "))
```
mwbrooks: nit: [We use capital "Loaded" in localserver.go](https://github.com/slackapi/slack-cli/pull/436/changes#diff-a5672d47ecf2c90f6b50fe73eea0880d4d9df7a6cab12ad914495e7ad21d46beR321). Perhaps capitalize this as well?
```suggestion
		io.PrintDebug(ctx, "Loaded variables from .env file: %s", strings.Join(keys, ", "))
```

internal/hooks/hooks_test.go
```diff
@@ -20,9 +20,86 @@ import (
 	"github.com/slackapi/slack-cli/internal/config"
 	"github.com/slackapi/slack-cli/internal/iostreams"
 	"github.com/slackapi/slack-cli/internal/slackdeps"
+	"github.com/spf13/afero"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
+func Test_Hooks_LoadDotEnv(t *testing.T) {
```
mwbrooks: suggestion: All of these tests are great, but we should add one for malformed `.env` files as well, I think. Since it's a user-edited file, we're guaranteed to have some malformed files.

internal/hooks/hooks.go
```diff
@@ -19,29 +19,49 @@ import (
...
+	}
+	file, err := afero.ReadFile(fs, ".env")
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil, nil
+		}
+		return nil, err
+	}
+	return godotenv.UnmarshalBytes(file)
+}
```
mwbrooks: question: Should we move this logic to `internal/config/dotenv.go` to consolidate dotenv handling to one place?

internal/hooks/hook_executor_default_test.go
zimeg: @mwbrooks I appreciate the fresh eye to these changes! I realize now this title isn't clear! 👻 

In https://github.com/slackapi/slack-cli/pull/436/changes/72828ecd29897241bfcdd07be0aae4a5ea1a30e8 we change this as:

```diff
- dotenv vars are loaded
+ dotenv vars and hook vars are loaded into the environment
```

internal/hooks/hook_executor_v2_test.go
zimeg: @mwbrooks Much appreciated call as always - this found https://github.com/slackapi/slack-cli/commit/72828ecd29897241bfcdd07be0aae4a5ea1a30e8 changes ☎️ ✨

internal/hooks/hooks.go
zimeg: @mwbrooks I'm open to this change but find we're inconsistent elsewhere with debug logs and this hopes to match a current pattern:

- Lowercase loggings during "init" sequences
- Sentence cased output for the "run" command

If we ought start standardizing to either of these I'd love to start now! 🐢 💨

internal/config/dotenv.go
```diff
@@ -17,24 +17,10 @@ package config
...
 
-// GetDotEnvFileVariables collects only the variables in the .env file
-func (c *Config) GetDotEnvFileVariables() (map[string]string, error) {
-	variables := map[string]string{}
-	file, err := afero.ReadFile(c.fs, ".env")
-	if err != nil && !c.os.IsNotExist(err) {
-		return variables, err
-	}
-	return godotenv.UnmarshalBytes(file)
-}
```
zimeg, line 33: 📦 note: This is moved to our own `slackdotenv` package to avoid odd dependencies of `config` and to avoid duplicate logic!

internal/slackdotenv/slackdotenv.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+	file, err := afero.ReadFile(fs, ".env")
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil, nil
+		}
+		return nil, err
+	}
+	return godotenv.UnmarshalBytes(file)
+}
```
zimeg, line 51: 🪬 note: I'm not confident on the `Read` vs `Load` term but for now this is perhaps clear. We might want to revisit this in changes to the actual ".env" file ongoing?

internal/hooks/hooks.go
zimeg: @mwbrooks So glad you picked up on this! We now have a `slackdotenv` package that both hooks and config and commands can use with https://github.com/slackapi/slack-cli/pull/436/commits/f63d0ffd4bef49dce33b6896b18baee81a05220d 🎁

internal/hooks/hooks_test.go
zimeg: 🧠 @mwbrooks Let's hope this isn't often! But we now assert that at least no error appears if so:

https://github.com/slackapi/slack-cli/blob/c1382dfec911518c2eb49fe20496bf9a3b38018d/internal/hooks/shell_test.go#L71-L79

internal/pkg/platform/localserver.go
zimeg: @mwbrooks Super appreciate that you noticed this! I agree it's meaningful logic that felt fragile in my initial implementation 👁️‍🗨️ 

We refactored our environment variable management into a few places:

- `internal/slackdotenv`: This reads the ".env" file.
- `internal/hooks/shell.go#HookExecOpts:ShellEnv()`: This orders variables for hook commands with precedence.

This change takes place in two commits https://github.com/slackapi/slack-cli/pull/436/changes/f63d0ffd4bef49dce33b6896b18baee81a05220d and https://github.com/slackapi/slack-cli/pull/436/changes/9160bcc5a69cb2a1600e3d6a98f6bbb87698f046 with some adjacent tests.

zimeg: @mwbrooks Not meaning to send before addressing the `Overload` options 👾 

These refactors make this and similar changes to loading from environment variable files much more safe I feel. I look forward to these changes!

internal/hooks/hooks.go
zimeg: @mwbrooks Great eye 👁️‍🗨️ ✨ We now have https://github.com/slackapi/slack-cli/pull/436/commits/525f6932b05d0cdb7d269906a8f9e3421a2799bb for this and an example as:

> ⚠️ unexpected character "\n" in variable name near "X\n"

For files that might match:

```
EXAMPLE_VARIABLE=12 # nice
X
```

.claude/CLAUDE.md
```diff
@@ -164,6 +164,12 @@ for name, tc := range tests {
 }
 ```
 
+### Error Handling
+
+- Wrap errors returned across package boundaries with `slackerror.Wrap(err, slackerror.ErrCode)` so they carry a structured error code
```
zimeg, line 169: 📚 note: This is motivated in ongoing changes with encouragement and optimism for checks from the wonderful `wrapcheck` lint:

> Checks that errors returned from external packages are wrapped.


🔗 https://golangci-lint.run/docs/linters/configuration/#wrapcheck

internal/pkg/platform/localserver.go
mwbrooks: Woo woo, thanks @zimeg. The `internal/slackdotenv` organization makes a lot more sense to me. Glad to see `internal/config` is gradually becoming an interface for `[system|project]/.slack/config.json`

internal/hooks/hooks.go
mwbrooks: This is perfect, thanks!

internal/hooks/hooks.go
mwbrooks: Good point. 🤔 I imagine we should choose and document in it as a style in the `MAINTAINERS_GUIDE.md` and `CLAUDE.md`.

internal/slackdotenv/slackdotenv.go
mwbrooks: thought: I like the `Read` right now. While it's personal, to me `Load` implies that the content is "loaded into some variable or context" while `Read` implies it'll read the file and return the contents.

This method reads the file and returns the contents. So, `Read` make sense to me.

PR #437, opened 2026-03-25
feat: list available ".env" variable names for non-hosted apps

zimeg:
### Changelog

> The `slack env list` command will now list the environment variables available from the `.env` file for apps that aren't run on Slack infrastructure.

### Summary

This PR lists available `.env` variable names for non-hosted apps to match what's available to hooks and the `run` command.

### Preview

demo

### Reviewers

Your check is so appreciated. The following commands might recommend testing:

```sh
$ slack create asdf -t slack-samples/bolt-js-starter-template
$ cd asdf
$ echo "EXAMPLE_TOKEN=12" > .env
$ slack install
$ slack env list
```

```sh
$ slack create asdf -t slack-samples/deno-starter-template
$ cd asdf
$ slack deploy
$ slack env add   # Perhaps odd examples can be used
$ slack env list
```

When done please do delete these apps! This is the environment command kindplace 🌲 ✨ 

### Requirements


cmd/env/list.go
```diff
@@ -34,11 +35,13 @@ func NewEnvListCommand(clients *shared.ClientFactory) *cobra.Command {
...
-			"infrastructure.",
+			"List environment variables available to the app at runtime.",
 			"",
-			"This command is supported for apps deployed to Slack managed infrastructure but",
-			"other apps can attempt to run the command with the --force flag.",
+			"Commands that run in the context of a project source environment variables from",
+			"the \".env\" file. This includes the \"run\" command.",
+			"",
+			"The \"deploy\" command gathers environment variables from the \".env\" file as well",
+			"unless the app is using ROSI features.",
```
zimeg, line 44: 📸 **Preview**

Image

cmd/env/list.go
```diff
@@ -81,20 +76,34 @@ func runEnvListCommandFunc(
...
+		variableNames, err = clients.API().ListVariables(
+			ctx,
+			selection.Auth.Token,
+			selection.App.AppID,
+		)
+		if err != nil {
+			return err
+		}
+	} else {
+		dotEnv, err := hooks.LoadDotEnv(clients.Fs)
```
zimeg: 🪝 note: We reuse the logic that `hooks` has for gathering environment variables for confidence in matching!

🐟 ramble: We don't **require** app selection before this at this point, but we might consider keeping it to avoid breaking change if it becomes required later. FWIW it's required for hosted apps so this is familiar experience I think!

cmd/env/list.go
```diff
@@ -112,22 +121,23 @@ func runEnvListCommandFunc(
 		},
 	}))
 
-	if len(variableNames) <= 0 {
+	if count <= 0 {
```
srtaalej, line 124: nice catch i wonder why count was not used here before 🤔

PR #445, opened 2026-03-27
refactor: remove 'version.Get()' and only use 'version.Raw()'

mwbrooks:
### Changelog

- N/A

### Summary

This pull request removes `version.Get()` and only uses `version.Raw()`. The motivation is the simplify version handling to be 1 method instead of 2 methods.

**Context:**

PR #429 consolidated version handling to prefer `version.Raw()` and `slackcontext.Version(ctx)`. A follow-up task was to remove `version.Get()` entirely since it creates confusion about which function to use. The only difference is that `.Get()` prepends a `v` prefix if missing - but we can guarantee the `v` prefix is always present without requiring `.Get()`.

**Why it's Safe to Remove:**

The `v` prefix is already guaranteed in all code paths that set Version:

1. **Build-Time (Makefile):** `git describe --tags --match 'v*.*.*'` - the `--match` pattern requires a v-prefixed tag, and git describe outputs the matching tag name, so the result always starts with `v`.
2. **Default Value (version.go):** `"v0.0.0-dev"` - already has the prefix.
3. **Env Var Override (version.go):** `SLACK_TEST_VERSION` - this is the one gap. QA testers could set it without a `v`. This PR adds validation by moving the v-prefix enforcement from `.Get()` into `init()`, ensuring Version is normalized at startup.

### Open Question 🧠 

- Do we want to rename `version.Raw()` to `version.Version()`?
  - Technically, it's no longer the "raw" version because it may have a v-prefix added.
  - In the future, we may want a `version.Major()`, `version.Minor()`, etc to get specific elements of the version.

### Requirements

### DIFF
--- a/scripts/install-windows-dev.ps1
+++ b/scripts/install-windows-dev.ps1
@@ -66,7 +66,8 @@ function check_slack_binary_exist() {
         Write-Host "Halting the install to avoid accidentally overwriting it."
 
         Write-Host "`nTry using an alias when installing to avoid name conflicts:"
-        Write-Host "`nirm https://downloads.slack-edge.com/slack-cli/install-windows.ps1 -Alias your-preferred-alias | iex"
+        Write-Host "`nInvoke-WebRequest -Uri https://downloads.slack-edge.com/slack-cli/install-windows.ps1 -OutFile 'install-windows.ps1'"
+        Write-Host ".\install-windows.ps1 -Alias your-preferred-alias"
         throw
       }
     }
--- a/scripts/install-windows.ps1
+++ b/scripts/install-windows.ps1
@@ -63,7 +63,8 @@ function check_slack_binary_exist() {
         Write-Host "Halting the install to avoid accidentally overwriting it."
 
         Write-Host "`nTry using an alias when installing to avoid name conflicts:"
-        Write-Host "`nirm https://downloads.slack-edge.com/slack-cli/install-windows.ps1 -Alias your-preferred-alias | iex"
+        Write-Host "`nInvoke-WebRequest -Uri https://downloads.slack-edge.com/slack-cli/install-windows.ps1 -OutFile 'install-windows.ps1'"
+        Write-Host ".\install-windows.ps1 -Alias your-preferred-alias"
         throw
       }
     }


internal/version/version.go
```diff
@@ -33,23 +33,23 @@ func init() {
 	if envVersion := getVersionFromEnv(); envVersion != "" {
 		Version = envVersion
 	}
+	Version = ensurePrefix(Version)
```
mwbrooks, line 35: note: We now enforce the `v` prefix on `init()` so that all version-pathways will now have a `v` prefix.

cmd/upgrade/upgrade.go
```diff
@@ -55,7 +55,7 @@ func NewCommand(clients *shared.ClientFactory) *cobra.Command {
 // When there are updates, the function will *not* print a message because the root command handles printing update notifications.
 func checkForUpdates(clients *shared.ClientFactory, cmd *cobra.Command) error {
 	ctx := cmd.Context()
-	updateNotification := update.New(clients, version.Get(), "SLACK_SKIP_UPDATE")
+	updateNotification := update.New(clients, version.Raw(), "SLACK_SKIP_UPDATE")
```
zimeg, line 58: 🎁 suggestion(non-blocking): In later changes to `upgrade` do we want to prefer `slackcontext` packages instead of importing `version` here?

https://github.com/slackapi/slack-cli/blob/e114959c41ccaa83d1749eac5ab1bad73583bfe5/internal/slackcontext/slackcontext.go#L174-L184

internal/version/version.go
```diff
@@ -33,23 +33,23 @@ func init() {
...
 }
 
-// Get the version and format it (e.g. `v1.0.0`)
-func Get() string {
-	version := Version
-	if match, _ := regexp.MatchString(`^[^v]`, version); match {
-		version = "v" + version
+// ensurePrefix ensures that the version string has a "v" prefix.
+func ensurePrefix(v string) string {
+	if match, _ := regexp.MatchString(`^[^v]`, v); match {
```
zimeg:
```suggestion
	if !strings.HasPrefix(v, "v") {
```

🌚 suggestion: While we're improving this we might import from `string` instead?

📚 https://pkg.go.dev/strings#HasPrefix

internal/version/version_test.go
```diff
@@ -52,10 +51,7 @@ func Test_Get(t *testing.T) {
 	}
 	for name, tc := range tests {
 		t.Run(name, func(t *testing.T) {
-			original := Version
-			defer func() { Version = original }()
-			Version = tc.version
-			assert.Equal(t, tc.expected, Get())
```
zimeg, line 58: 🪓 praise: Huge improvement IMHO!

internal/version/version.go
zimeg: 🦠 thought: I like this approach! It keeps the `ctx` corrent and commands are well structured for this I believe!

cmd/upgrade/upgrade.go
mwbrooks: Good suggestion. During this PR, I was surprised to see so many `version.Raw()` calls still remain. In the follow-up PR, I'll aim to replace these - it's less dangerous now that `slackcontext.Version(ctx)` fallsback on the raw version. 👍🏻

internal/version/version.go
mwbrooks: Good suggestion! Thanks for keeping an eye out for opportunities to reduce our regex-foo!

Commit 218fa43d14c0622ab5562c5410b2b431152e5750 replaces `regexp.MatchString` with `strings.HasPrefix`. The `ensurePrefix` method must also now handle the empty string use-case, which our unit tests already cover.

PR #451, opened 2026-03-30
feat: `env add` command supports the ".env" file for Bolt Framework apps

zimeg:
### Changelog

> The `slack env add` command can now add and update environment variables in the `.env` file for apps that aren't built or run on Slack infrastructure. This makes setting up and iterating on a Bolt app from the command line faster.

### Summary

This PR uses the `env add` command to add variables to the `.env` file for non-hosted apps. Follows #437 🌲

### Preview

demo

### Reviewers

Attempt to write or update variables from this example and branch:

```
$ vim .env
# Database configuration
DB_HOST=localhost
DB_PORT=5432 # Pretend default
export DB_KEY="---START---
password
---END---"

# App secrets
SECRET_KEY=original_secret
export API_TOKEN="xoxb-new-token" # https://api.slack.com/apps/example


# Feature flags
DEBUG=true
$ slack env list
$ slack env add API_TOKEN
$ slack env add LOG_FORMAT
$ slack env list
```

### Notes

- We require app selection to determine if an app is hosted or not. Perhaps we can workaround that for apps with the `remote` function runtime but this app selection might remain useful in later enhancements:

```
.env
.env.A0123456789
``` 

### Requirements

### DIFF
--- a/cmd/env/add.go
+++ b/cmd/env/add.go
@@ -26,6 +26,7 @@ import (
 	"github.com/slackapi/slack-cli/internal/slackdotenv"
 	"github.com/slackapi/slack-cli/internal/slacktrace"
 	"github.com/slackapi/slack-cli/internal/style"
+	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
 )
 
@@ -146,17 +147,24 @@ func runEnvAddCommandFunc(clients *shared.ClientFactory, cmd *cobra.Command, arg
 			},
 		}))
 	} else {
+		exists, err := afero.Exists(clients.Fs, ".env")
+		if err != nil {
+			return err
+		}
 		err = slackdotenv.Set(clients.Fs, variableName, variableValue)
 		if err != nil {
 			return err
 		}
 		clients.IO.PrintTrace(ctx, slacktrace.EnvAddSuccess)
+		var details []string
+		if !exists {
+			details = append(details, "Created a project .env file that shouldn't be added to version control")
+		}
+		details = append(details, fmt.Sprintf("Successfully added \"%s\" as a project environment variable", variableName))
 		clients.IO.PrintInfo(ctx, false, "\n%s", style.Sectionf(style.TextSection{
-			Emoji: "evergreen_tree",
-			Text:  "App Environment",
-			Secondary: []string{
-				fmt.Sprintf("Successfully added \"%s\" as a project environment variable", variableName),
-			},
+			Emoji:     "evergreen_tree",
+			Text:      "App Environment",
+			Secondary: details,
 		}))
 	}
 	return nil


cmd/env/add.go
```diff
@@ -127,27 +124,40 @@ func runEnvAddCommandFunc(clients *shared.ClientFactory, cmd *cobra.Command, arg
...
+			variableName,
+			variableValue,
+		)
+		if err != nil {
+			return err
+		}
+		clients.IO.PrintTrace(ctx, slacktrace.EnvAddSuccess)
+		clients.IO.PrintInfo(ctx, false, "\n%s", style.Sectionf(style.TextSection{
+			Emoji: "evergreen_tree",
+			Text:  "App Environment",
```
zimeg, line 144: 🎨 note: We might follow up to change the section text for these `env` commands for outputs that match:

> - 🌲 Environment Add
> - 🌲 Environment List
> - 🌲 Environment Remove

👾 ramble: I avoided this change for now because this PR is so large!

cmd/env/add.go
```diff
@@ -127,27 +124,40 @@ func runEnvAddCommandFunc(clients *shared.ClientFactory, cmd *cobra.Command, arg
...
+		err = slackdotenv.Set(clients.Fs, variableName, variableValue)
+		if err != nil {
+			return err
+		}
+		clients.IO.PrintTrace(ctx, slacktrace.EnvAddSuccess)
+		clients.IO.PrintInfo(ctx, false, "\n%s", style.Sectionf(style.TextSection{
+			Emoji: "evergreen_tree",
+			Text:  "App Environment",
+			Secondary: []string{
+				fmt.Sprintf("Successfully added \"%s\" as a project environment variable", variableName),
```
zimeg: 📚 thought: This might behave in unexpected fashion for now since we prompt to select an app but save variables to the project, where all apps have access. I understand we have plans for more nuanced environment variable files and also that these `.env` file behaviors are understood when developing so I don't know if this requires more of a callout!

internal/slackdotenv/slackdotenv.go
```diff
@@ -49,3 +50,103 @@ func Read(fs afero.Fs) (map[string]string, error) {
...
+	// Check for an existing .env file and parse it to detect existing keys.
+	existing, err := Read(fs)
+	if err != nil {
+		return err
+	}
+
+	// If the file does not exist, create it with the new entry.
+	if existing == nil {
+		return writeFile(fs, []byte(newEntry+"\n"))
+	}
```
zimeg, line 82: 🔬 question: From the command logic, would it be meaningful to output a note that the `.env` file shouldn't be checked into version control? Should we check if it's ignored already?

cmd/env/add.go
```diff
@@ -33,12 +34,15 @@ func NewEnvAddCommand(clients *shared.ClientFactory) *cobra.Command {
 		Use:   "add <name> <value> [flags]",
 		Short: "Add an environment variable to the app",
 		Long: strings.Join([]string{
-			"Add an environment variable to an app deployed to Slack managed infrastructure.",
+			"Add an environment variable to the app.",
```
mwbrooks: suggestion: Perhaps we should say "project" instead of "app"? For ROSI apps, it was actually the app (server-side) but for all other apps including locally run Deno apps, it's actually the "project's" .env file.
```suggestion
			"Add an environment variable to the project.",
```

cmd/env/add.go
```diff
@@ -33,12 +34,15 @@ func NewEnvAddCommand(clients *shared.ClientFactory) *cobra.Command {
 		Use:   "add <name> <value> [flags]",
 		Short: "Add an environment variable to the app",
```
mwbrooks: suggestion(non-blocking): See the suggestion below. You may want to merge this change if you accept the one below.
```suggestion
		Short: "Add an environment variable to the project",
```

cmd/env/add.go
```diff
@@ -33,12 +34,15 @@ func NewEnvAddCommand(clients *shared.ClientFactory) *cobra.Command {
 		Use:   "add <name> <value> [flags]",
```
mwbrooks: suggestion: Seems like a nice time to update the Usage to be more accurate?
```suggestion
		Use:   "add <name> [value] [flags]",
```

cmd/env/add.go
```diff
@@ -33,12 +34,15 @@ func NewEnvAddCommand(clients *shared.ClientFactory) *cobra.Command {
...
+			"Add an environment variable to the app.",
 			"",
 			"If a name or value is not provided, you will be prompted to provide these.",
 			"",
-			"This command is supported for apps deployed to Slack managed infrastructure but",
-			"other apps can attempt to run the command with the --force flag.",
+			"Commands that run in the context of a project source environment variables from",
+			"the \".env\" file. This includes the \"run\" command.",
+			"",
+			"The \"deploy\" command gathers environment variables from the \".env\" file as well",
```
mwbrooks: suggestion(non-blocking): Please ignore if you want, but thought this may be a chance to keep the code simpler.
```suggestion
			`the ".env" file. This includes the "run" command.`,
			"",
			`The "deploy" command gathers environment variables from the ".env" file as well`,
```

cmd/env/add.go
```diff
@@ -69,26 +73,19 @@ func NewEnvAddCommand(clients *shared.ClientFactory) *cobra.Command {
...
 func preRunEnvAddCommandFunc(ctx context.Context, clients *shared.ClientFactory, cmd *cobra.Command) error {
 	clients.Config.SetFlags(cmd)
-	err := cmdutil.IsValidProjectDirectory(clients)
-	if err != nil {
-		return err
-	}
-	if clients.Config.ForceFlag {
-		return nil
-	}
-	return cmdutil.IsSlackHostedProject(ctx, clients)
```
mwbrooks, line 83: praise: Bolt is coming for you `env`! ⚡

cmd/env/add.go
mwbrooks: praise: Good call! I like that this PR is keeping ROSI untouched, but we should make a note to follow-up on it adding parity between the two experiences.

cmd/env/add.go
mwbrooks: Interesting, this is something that I overlooked when we first proposed the command changes.

suggestion(non-blocking): I'm happy with the current approach as phase 1, if you'd like to land the PR, avoid it becoming too large, and iterate.

However, I think we need to improve this experience. One of the main value props of the `env add` command is to set your app's third-party API Tokens. However, these tokens need to be set before you can `run` the app.

But... if you can't `env add` until you've created an app, then you're in a bit of a pickle. 🐣 You'd need to use the (currently) awkward `install` command before `env add`.

Would checking for the Deno runtime be a way to determine if we need to prompt for an app? If it's Deno then we prompt (local = .env, deploy = API), otherwise we skip the prompt and just use `.env`.

cmd/env/add_test.go
```diff
@@ -269,24 +204,73 @@ func Test_Env_AddCommand(t *testing.T) {
 				)
 			},
 		},
+		"add a variable to the .env file for non-hosted app": {
```
mwbrooks, line 242: praise: Love seeing the new test use-case!

internal/slackdotenv/slackdotenv.go
```diff
@@ -49,3 +50,103 @@ func Read(fs afero.Fs) (map[string]string, error) {
 	}
 	return vars, nil
 }
+
+// Set sets a single environment variable in the .env file, preserving
+// comments, blank lines, and other formatting. If the key already exists its
+// value is replaced in-place. Otherwise the entry is appended. The file is
+// created if it does not exist.
+func Set(fs afero.Fs, name string, value string) error {
```
mwbrooks, line 59: praise: 🧑‍🍳 💋 👌🏻 What a great Godoc description!

internal/slackdotenv/slackdotenv.go
mwbrooks: Good question. It feels like the CLI is overstepping to add the `.env` to the `.gitignore` but perhaps that's just my own feeling. Others may consider that a delightful feature.

I think outputting a message is a very good start and allows us to improve upon it later (e.g. editing the `.gitignore`).

This approach has worked well for us in the past:
1. Output the manual steps
2. Let it bake and gauge how it feels
3. Automate the manual steps

internal/slackdotenv/slackdotenv.go
```diff
@@ -49,3 +50,103 @@ func Read(fs afero.Fs) (map[string]string, error) {
...
+			WithMessage("Failed to marshal the .env variable: %s", err)
+	}
+
+	// Check for an existing .env file and parse it to detect existing keys.
+	existing, err := Read(fs)
+	if err != nil {
+		return err
+	}
+
+	// If the file does not exist, create it with the new entry.
```
mwbrooks, line 79: question(non-blocking): This is out-of-scope of this PR, but our original discussion included a `env init` that would copy a `.env.sample` or `.env.example` to be `.env`. 

If we eventually add this ability, I can see the init being called here instead of creating a blank `.env`, since the app may include a `.env.sample` or `.env.example`.

internal/slackdotenv/slackdotenv.go
```diff
@@ -49,3 +50,103 @@ func Read(fs afero.Fs) (map[string]string, error) {
...
+	entries := []string{
+		"export " + name + "=" + oldMarshaledValue,
+		"export " + name + "=" + `"` + oldValue + `"`,
+		"export " + name + "=" + oldValue,
+		"export " + name + "=" + "'" + oldValue + "'",
+		name + "=" + oldMarshaledValue,
+		name + "=" + `"` + oldValue + `"`,
+		name + "=" + oldValue,
+		name + "=" + "'" + oldValue + "'",
+	}
```
mwbrooks: thought: This feels very brittle to me. 🤔 

For example, the following format fails as a test case, but I'd expect it to be a valid entry - spaces around the equals sign, which works with Python's dotenv package:

```bash
FOO = "bar"
```

Here is the test case:

```go
		"some poorly formatted .env file": {
			name:         "FOO",
			value:        "bar",
			expectedFile: "FOO = \"bar\"\n",
		},
```

In production, if I use the above `.env` file and execute the `env add` then it duplicates the entries:

```bash
$ lack env add FOO banana

$ cat .env
FOO = "bar"
FOO="banana"
```

I feel like there's a hesitation to use regex or string matches, but it may be the more flexible approach here.

cmd/env/add.go
zimeg: @mwbrooks Thanks for calling out this awkward experience 👾 🔭 

I'm confident we can decide to prompt for Deno apps and not Bolt apps, but we might want to follow up to mirror this in the `env list` command.

I was hesitant earlier to avoid blocking possible follow ups requiring app selection, but perhaps those explorations make app selection optional! 🚢

mwbrooks: 🫡 Sounds good! Let's follow-up with exploring how to improve this.

internal/slackdotenv/slackdotenv.go
```diff
@@ -49,3 +51,91 @@ func Read(fs afero.Fs) (map[string]string, error) {
...
+	// double-quoted values with embedded newlines.
+	re := regexp.MustCompile(
+		`(?m)(^[^\S\n]*export[^\S\n]+|^[^\S\n]*)` + regexp.QuoteMeta(name) + `[^\S\n]*=[^\S\n]*` +
+			`(?:` +
+			`"(?:[^"\\]|\\.)*"` + // double-quoted (with escapes)
+			`|'[^']*'` + // single-quoted
+			"|`[^`]*`" + // backtick-quoted
+			`|[^\n]*` + // unquoted to end of line
+			`)`,
+	)
```
mwbrooks, line 115: praise: Monster rewrite! 🐺 Thanks for the comments!

internal/slackdotenv/slackdotenv_test.go
```diff
@@ -111,3 +111,167 @@ func Test_Read(t *testing.T) {
...
+			expectedFile:  "FOO=\"bar\"\n",
+		},
+		"adds a variable preserving existing variables": {
+			existingEnv:   "EXISTING=value\n",
+			writeExisting: true,
+			name:          "NEW_VAR",
+			value:         "new_value",
+			expectedFile:  "EXISTING=value\nNEW_VAR=\"new_value\"\n",
+		},
+		"adds a variable preserving comments and blank lines": {
```
mwbrooks: suggestion: Perhaps we should clarify that it preserves newline comments (since inline comments are not preserved yet).

cmd/env/add.go
zimeg: @mwbrooks Oh wow I agree it's more clear after https://github.com/slackapi/slack-cli/pull/451/commits/51641d2e7a9e7213534541208a85ba0c3e9a5066! Thanks for suggesting this improvement 📚

cmd/env/add.go
zimeg: @mwbrooks Both suggestions landed in https://github.com/slackapi/slack-cli/pull/451/commits/51641d2e7a9e7213534541208a85ba0c3e9a5066! 🎁

cmd/env/add.go
zimeg: @mwbrooks Ahha at the time of writing I didn't have access to the backtick ``` character! Thanks 💡 ✨

cmd/env/add.go
zimeg: @mwbrooks This is included in https://github.com/slackapi/slack-cli/pull/451/commits/aa718e0aff11caea2dd7d72dc9e3c5e465b75098 but I'm unsure if we should encourage a required `name` argument too?

For now I'll keep recommending this documented use case for clear shell commands 📚

internal/slackdotenv/slackdotenv.go
zimeg: @mwbrooks Let's keep note of this! I feel like prompt options for starting from a template file will be useful 🍀

internal/slackdotenv/slackdotenv.go
zimeg: @mwbrooks Thank you for testing edges of these formats. I'm bummed it's not a standard but let's support these findings and related edges in https://github.com/slackapi/slack-cli/pull/451/commits/9d1972258e3e41f1e6a72e62e7850cc3c209e55d 🤓

internal/slackdotenv/slackdotenv_test.go
zimeg: @mwbrooks Good suggestions as always - thank you for sharing your good eye in https://github.com/slackapi/slack-cli/pull/451/commits/7f314926c7d0b433633b9d72d5001af1ec8c2eb4 👁️‍🗨️ ✨

internal/slackdotenv/slackdotenv.go
zimeg: @mwbrooks Super appreciative of these shared thoughts 🧠

We're landing https://github.com/slackapi/slack-cli/pull/451/commits/f708074b00149d228d1261c17572e744b20f4137 with these changes to get a feel for things, and I'd love your feedback more if something is obvious 🎨 🐢

📸 This demo shows the disclaimer for new `.env` file creations:

output

internal/slackdotenv/slackdotenv.go
zimeg: @mwbrooks The test cases from prior were so helpful in confidence 🧪 ✨

PR #454, opened 2026-03-31
fix: apply --no-color flag to huh interactive prompts

srtaalej:
### Changelog

> N/A - Fixes for unreleased changes.

### Summary

The --no-color flag stripped styles from CLI output but not from huh interactive prompts (select, confirm, input, etc.), because huh uses its own theme/rendering pipeline. This PR added ThemePlain() — a colorless huh theme with no ANSI styling so that when --no-color is set, forms now use ThemePlain + huh's accessible mode, rendering as plain numbered lists instead of a styled TUI

### Test plan                                                 

Run `slack create --no-color` and verify prompts render as plain numbered lists with no color
Run `slack login --no-color` and verify confirmation prompt has no color
Run `slack create` (without --no-color) and verify themed prompts still work
`go test ./internal/style/ ./internal/iostreams/`

### Requirements

### DIFF
--- a/cmd/feedback/feedback.go
+++ b/cmd/feedback/feedback.go
@@ -68,9 +68,8 @@ const (
 
 // Supported survey names
 const (
-	SlackCLIFeedback                = "slack-cli"
-	SlackPlatformFeedback           = "slack-platform"
-	SlackPlatformFeedbackDeprecated = "platform-improvements" // DEPRECATED(semver:major)
+	SlackCLIFeedback      = "slack-cli"
+	SlackPlatformFeedback = "slack-platform"
 )
 
 type SurveyConfigInterface interface {
@@ -239,12 +238,6 @@ func runFeedbackCommand(ctx context.Context, clients *shared.ClientFactory, cmd 
 	if len(SurveyStore) == 0 {
 		clients.IO.PrintInfo(ctx, false, "No feedback options currently available; please try again later")
 		return nil
-	}
-
-	// DEPRECATED(semver:major): Support the deprecated survey name for backwards compatibility
-	if surveyNameFlag == SlackPlatformFeedbackDeprecated {
-		surveyNameFlag = SlackPlatformFeedback
-		clients.IO.PrintWarning(ctx, "DEPRECATED: The '--name %s' flag is deprecated; use '--name %s' instead", SlackPlatformFeedbackDeprecated, SlackPlatformFeedback)
 	}
 
 	surveyNames, surveyPromptOptions := initSurveyOpts(ctx, clients, SurveyStore)
--- a/cmd/feedback/feedback_test.go
+++ b/cmd/feedback/feedback_test.go
@@ -263,17 +263,6 @@ func TestShowSurveyMessages(t *testing.T) {
 
 func Test_Feedback_FeedbackCommand(t *testing.T) {
 	testutil.TableTestCommand(t, testutil.CommandTests{
-		// DEPRECATED(semver:major): Support the deprecated survey name for backwards compatibility
-		"supports deprecated --name platform-improvements": {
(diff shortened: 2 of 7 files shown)


internal/style/theme.go
```diff
@@ -126,6 +126,36 @@ func Chevron() string {
...
+		t.Focused.ErrorMessage = lipgloss.NewStyle()
+		t.Focused.SelectSelector = lipgloss.NewStyle().SetString("> ")
+		t.Focused.Option = lipgloss.NewStyle()
+		t.Focused.MultiSelectSelector = lipgloss.NewStyle().SetString("> ")
+		t.Focused.SelectedOption = lipgloss.NewStyle()
+		t.Focused.SelectedPrefix = lipgloss.NewStyle().SetString("[x] ")
+		t.Focused.UnselectedOption = lipgloss.NewStyle()
+		t.Focused.UnselectedPrefix = lipgloss.NewStyle().SetString("[ ] ")
+		t.Focused.FocusedButton = lipgloss.NewStyle().Padding(0, 2).MarginRight(1)
+		t.Focused.BlurredButton = lipgloss.NewStyle().Padding(0, 2).MarginRight(1)
```
zimeg: 📸 issue: I'm not sure if these are the correct options but I find the confirm default isn't highlighted now. I'll include a screenshot for other reviewers:

Image

zimeg:
👾 ramble: I'm now uncertain if the comment about "hint" formatting is too nuanced but I do find that the Charm team offers a "default" theme that might be useful to use here for confidence in edge cases?

image

🔗 https://github.com/charmbracelet/huh?tab=readme-ov-file#themes

internal/style/theme.go
```diff
@@ -122,6 +122,36 @@ func Chevron() string {
...
+		t.Focused.ErrorIndicator = lipgloss.NewStyle().SetString(" *")
+		t.Focused.ErrorMessage = lipgloss.NewStyle()
+		t.Focused.SelectSelector = lipgloss.NewStyle().SetString("> ")
+		t.Focused.Option = lipgloss.NewStyle()
+		t.Focused.MultiSelectSelector = lipgloss.NewStyle().SetString("> ")
+		t.Focused.SelectedOption = lipgloss.NewStyle()
+		t.Focused.SelectedPrefix = lipgloss.NewStyle().SetString("[x] ")
+		t.Focused.UnselectedOption = lipgloss.NewStyle()
+		t.Focused.UnselectedPrefix = lipgloss.NewStyle().SetString("[ ] ")
+		t.Focused.FocusedButton = lipgloss.NewStyle().Padding(0, 2).MarginRight(1).Bold(true).Underline(true)
```
zimeg: 🌟 praise: This is a clever approach @srtaalej!

👾 note: I made changes in 2c514837a57db5ae24b33aadb32743174d4ddb5e that adjust this to remove the **bold** and **underline** formatting when using this `--no-color` flag. We'll want to avoid ANSI altogether:

🎨 https://en.wikipedia.org/wiki/ANSI_escape_code#Select_Graphic_Rendition_parameters

📸 note: Here's a move recent preview, with also changes of 784f9f200fb3bd7cef0b7ccd7b26aaba935daf02 added:

https://github.com/user-attachments/assets/ae1d8cde-01fc-43b8-aeab-f1261674ac5c

zimeg (approved):
@srtaalej LGTM! And thanks for exploring what's possible with these packages 📺 ✨ 

I'm approving this after a bit more testing and some changes noted below. If this all looks alright I think we can merge it! One note we might want to follow up with support for the variables:

```
$ NO_COLOR=1 slack run
```

Which right now removes format from `huh` but not the select options! No blocker for this PR 🚢 

I hope the note on ANSI limitations makes sense too 📚 We might find that unicode and other box drawing characters are alright but not escaped sequences for this flag 🤓

PR #455, opened 2026-03-31
feat: add accessible flag and environment variable for screen reader friendly forms

srtaalej:
### Changelog

Interactive forms become screen reader friendly when the `ACCESSIBLE` environment variable is set or when using the `--accessible` flag.    

### Summary

Related: #454  
This PR adds a global --accessible flag that switches huh interactive prompts to accessible mode. In accessible mode, select prompts render as numbered lists with "Enter a number between 1 and N" input, confirm prompts accept y/n text, and input prompts use plain line-by-line I/O. This makes the CLI usable with screen readers by avoiding the TUI-based rendering                                          

<img width="1496" height="1118" alt="Screenshot 2026-03-31 at 4 03 49 PM" src="https://github.com/user-attachments/assets/1ecd89e5-886c-4cd9-9e6c-eaef1f19976b" />
                     
                                                                                                            
### Test plan                                                                                                                      
                                                                                                                   
Run `slack create --accessible` and verify prompts render as numbered lists                                                    
Run `slack login --accessible` and verify confirm prompt accepts y/n                                                         
Run `slack create` (without flag) and verify normal TUI prompts still work                                                     
`go test ./internal/iostreams/ -run TestFormsAccessible  `                                                                     
                                                                                                                                 
### Requirements


cmd/root_test.go
```diff
@@ -194,6 +195,15 @@ func TestVersionFlags(t *testing.T) {
...
 
+func Test_AccessibleImpliesNoColor(t *testing.T) {
+	cfg := &config.Config{Accessible: true}
+	// Simulate the logic from PersistentPreRunE
+	if cfg.Accessible {
+		cfg.NoColor = true
+	}
+	assert.True(t, cfg.NoColor, "--accessible should imply --no-color")
+}
+
```
zimeg:
```suggestion
```

🪓 note: I'm not sure this test is so useful against regressions here?

internal/config/config.go
```diff
@@ -58,6 +58,7 @@ type Config struct {
 	SlackTestTraceFlag      bool
 	TeamFlag                string
 	TokenFlag               string
+	Accessible              bool
```
zimeg: 🌲 thought: As we're introducing this, should we include the environment variable similar?

```
ACCESSIBLE
```

> We recommend setting this through an environment variable or configuration option to allow the user to control accessibility.

🔗 https://github.com/charmbracelet/huh?tab=readme-ov-file#accessibility

zimeg: 🔍 ramble: We might find this adjacent file most useful!

https://github.com/slackapi/slack-cli/blob/f39d175b241d07ed6c47dde6b0045cbd531b3dba/internal/config/dotenv.go#L23-L24

zimeg (commented):
@srtaalej So amazing to find this feature arriving here! 🍀 ✨ 

Before approving I want to leave a few comments, but think we can follow up with some other improvements to the edges:

- **Environment Variable**: The `ACCESSIBLE` environment variable is standard I understand and we should include support alongside these changes! This makes preferring these prompts easier with an `export` added to the shell profile 🔬
- **Blank Selections**: The "select" prompts request a number to be entered but accept blank input for the first option! This might be an upstream issue 🐛 
- **Input Default**: Adjacent for the "input" prompt I find the default value of the `create` command name isn't showing 🧪 

The first note is most important for these changes IMHO! I'd be curious if the latter two can be proven in unit tests, but please don't consider those blocking 🙏

internal/iostreams/forms.go
```diff
@@ -39,13 +40,20 @@ func newForm(io *IOStreams, field huh.Field) *huh.Form {
...
+		form = form.WithAccessible(true)
+	}
 	return form
 }
 
 // buildInputForm constructs an interactive form for text input prompts.
 func buildInputForm(io *IOStreams, message string, cfg InputPromptConfig, input *string) *huh.Form {
+	title := message
+	if io != nil && io.config.Accessible && cfg.Placeholder != "" {
+		title = fmt.Sprintf("%s (default: %s)", message, cfg.Placeholder)
```
zimeg:
```suggestion
		title = fmt.Sprintf("%s (default: %s):", strings.TrimSuffix(message, ":"), cfg.Placeholder)
```

🔭 suggestion: This might not be the right syntax but I think formatting input ends best with a colon:

```diff
- Name your app: (default: flamboyant-salamander-784)
+ Name your app (default: flamboyant-salamander-784):
```

zimeg: 🎨 ramble: We don't have a comment on this in the `STYLE_GUIDE` at this time and we're not consistent on using a `:` at all or sometimes `?` instead. This might be something nice to align over time.

internal/iostreams/forms.go
```diff
@@ -100,8 +108,13 @@ func buildSelectForm(io *IOStreams, msg string, options []string, cfg SelectProm
 		opts = append(opts, huh.NewOption(key, opt))
 	}
 
+	title := msg
+	if io != nil && io.config.Accessible && len(options) > 0 {
+		title = fmt.Sprintf("%s (press Enter for 1)", msg)
```
zimeg:
```suggestion
		title = fmt.Sprintf("%s (press Enter for 1):", strings.TrimSuffix(msg, ":"))
```

🪬 suggestion: Similar to the `:` suggestion above - I'm more hesitant to this one because some titles might end with "?" but I don't think that formatting is incorrect...

zimeg (commented):
@srtaalej Thanks for these updates! I'm hoping we can land a few more to bring this to a polished state for release 🔬 ✨ 

I tested some more cases that I'll share here:

-  **One selection option**: The hint is solid IMHO! I share more comments on the title format later but I'm a fan of the hint.

Image

- **Multiple select options**: This highlights the ":" format a bit more for following comments.

Image

- **Invalid select options**: Super cool that this validates!

Image

- **Confirmation default**: This appears as a kind callout.

Image

- **Input prompts without default**: These might favor the trailing ":" also?

Image

The formatting issue I callout in comments below and some screenshots above is that I think form titles should end with a ":" to signal input! Some workarounds might be needed in code but this outputs well I think.

Let me know if more testing can be requested 🫡

zimeg:
🤖 CI error: https://github.com/slackapi/slack-cli/actions/runs/24351525876/job/71211294566#step:4:39

```
  Running [/Users/runner/golangci-lint-2.11.4-darwin-arm64/golangci-lint config path] in [/Users/runner/work/slack-cli/slack-cli] ...
  Running [/Users/runner/golangci-lint-2.11.4-darwin-arm64/golangci-lint config verify] in [/Users/runner/work/slack-cli/slack-cli] ...
  Running [/Users/runner/golangci-lint-2.11.4-darwin-arm64/golangci-lint run  --timeout=5m] in [/Users/runner/work/slack-cli/slack-cli] ...
  internal/iostreams/forms.go:1: : # github.com/slackapi/slack-cli/internal/iostreams [github.com/slackapi/slack-cli/internal/iostreams.test]
  Error: internal/iostreams/forms.go:53:44: undefined: strings
  Error: internal/iostreams/forms.go:113:50: undefined: strings (typecheck)
  // Copyright 2022-2026 Salesforce, Inc.
  1 issues:
  * typecheck: 1
  
  Error: issues found
```

internal/config/config.go
```diff
@@ -58,6 +59,7 @@ type Config struct {
...
```
zimeg:
```suggestion
	AccessibleFlag              bool
```

🧮 quibble: We might want to match adjacent configurations and perhaps alphabetical order this? The following `NoColor` option can perhaps find similar change in a follow up PR I think too

internal/iostreams/forms.go
```diff
@@ -39,13 +41,20 @@ func newForm(io *IOStreams, field huh.Field) *huh.Form {
...
+	}
 	return form
 }
 
 // buildInputForm constructs an interactive form for text input prompts.
 func buildInputForm(io *IOStreams, message string, cfg InputPromptConfig, input *string) *huh.Form {
+	title := message
+	if io != nil && io.config.Accessible && cfg.Placeholder != "" {
+		title = fmt.Sprintf("%s (default: %s):", strings.TrimSuffix(message, ":"), cfg.Placeholder)
+	}
```
zimeg: 📸 suggestion: Let's add a missing ":" to inputs without placeholder too:

**With interactive prompts** - Unsure if this requires a change but might be nice? If it doesn't feel right we can hold off on update here!

Image

**Without interactive prompts** - Feels more meaningful to have a split between output and input

Image

📣 I suggest the form following:

```
Variable name:
```

zimeg: 🔬 note: Similar comment might be applied to password forms and others?

zimeg (commented):
@srtaalej A few more comments of polish before this lands but it's feeling good in testing 💾 

The comments below hint at variable suffixes and placement and also standard line endings to these prompts. It'd be awesome to update the preview in the PR description and perhaps include test case also for the password prompt before merge 🙏

internal/config/config.go
```diff
@@ -25,6 +25,7 @@ import (
 const slackAutoRequestAAAEnv = "SLACK_AUTO_REQUEST_AAA"
 const slackConfigDirEnv = "SLACK_CONFIG_DIR"
 const slackDisableTelemetryEnv = "SLACK_DISABLE_TELEMETRY"
+const slackAccessibleEnv = "ACCESSIBLE"
```
zimeg: 🧮 quibble: These variables can be ordered too!

internal/config/flags.go
```diff
@@ -37,6 +37,7 @@ func (c *Config) InitializeGlobalFlags(cmd *cobra.Command) {
 	cmd.PersistentFlags().BoolVarP(&c.DeprecatedDevFlag, "dev", "d", false, "use dev apis")                                                // Can be removed after v0.25.0
 	cmd.PersistentFlags().StringVarP(&c.DeprecatedWorkspaceFlag, "workspace", "", "", "select workspace or organization by domain name or team ID")
 	cmd.PersistentFlags().StringSliceVarP(&c.ExperimentsFlag, "experiment", "e", nil, "use the experiment(s) in the command")
+	cmd.PersistentFlags().BoolVarP(&c.AccessibleFlag, "accessible", "", false, "use accessible prompts for screen readers")
```
zimeg: 🧮 quibble: Order!

zimeg (approved):
@srtaalej LGTM to merge when right times.

I comment about ordering here because later variables become less obvious where to add and changing this later becomes a hassle but let's think about how we avoid quibbles like this overall! 🤓

PR #460, opened 2026-04-02
feat: alias `env set` and `env unset` as primary commands

zimeg:
### Changelog

> The `env` commands were aliased to `env set` and `env unset` to match more common shell conventions. These commands were recommended as `add` and `remove` before, which became confusing when existing variables were updated.

### Summary

This PR aliases the `env set` and `env unset` commands to be the preferred command of setting and unsetting environment variables.

### Preview

env

### Reference

- https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html
- https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#index-unset
### Requirements

- [x] I've read and understood the [Contributing Guidelines](https://github.com/slackapi/slack-cli/blob/main/.github/CONTRIBUTING.md) and have done my best effort to follow them.
- [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).

PR #457, opened 2026-04-01
feat: `env remove` command supports the ".env" file for Bolt Framework apps

zimeg:
### Changelog

> The `slack env remove` command will remove environment variables from the `.env` file for apps that aren't built or run on Slack infrastructure.

### Summary

This PR brings support for Bolt frameworks removing variables from the `.env` file with the `env remove` command! 🧾 

```
$ slack env remove
```

Follows #437 and #451 and #456 recent.

### Preview

demo

### Reviewers

Use the various `env` commands to interact with this file:

```
$ cat .env
# Tests follow
EXAMPLE_VARIABLE=12
PLACEHOLDER="false"           # Leftover comments
DATABASE_SECRET="---start---
EOF
---end---"
DETAILS=
$ slack env list
$ slack env remove
$ slack env list
$ cat .env
```

### Notes

- ROSI continue to call the API. For local development the `.env` file is now supported.
- Inlined comments for a removed variable are also removed. This is different from `env add` but might make sense?

### Requirements

- [x] I've read and understood the [Contributing Guidelines](https://github.com/slackapi/slack-cli/blob/main/.github/CONTRIBUTING.md) and have done my best effort to follow them.
- [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).

PR #461, opened 2026-04-02
fix: format env command section headers together as titles

zimeg:
### Changelog

> The sections of `env` command outputs are now titled to match the command itself and are grouped under one section for each command.

### Summary

This PR formats the `env` command section headers together in a standard title format.

### Preview

demo

### Notes

- This change has adjacent changes #457 and #460 with changes to aliases and command features which might make review odd. I hope to keep these branches updated with the latest changes on `main` though!

### Requirements

### DIFF
--- a/cmd/env/set.go
+++ b/cmd/env/set.go
@@ -139,7 +139,7 @@ func runEnvSetCommandFunc(clients *shared.ClientFactory, cmd *cobra.Command, arg
 		if err != nil {
 			return err
 		}
-		details = append(details, fmt.Sprintf("Successfully added \"%s\" as an app environment variable", variableName))
+		details = append(details, fmt.Sprintf("Successfully set \"%s\" as an app environment variable", variableName))
 	} else {
 		exists, err := afero.Exists(clients.Fs, ".env")
 		if err != nil {
@@ -152,7 +152,7 @@ func runEnvSetCommandFunc(clients *shared.ClientFactory, cmd *cobra.Command, arg
 		if !exists {
 			details = append(details, "Created a project .env file that shouldn't be added to version control")
 		}
-		details = append(details, fmt.Sprintf("Successfully added \"%s\" as a project environment variable", variableName))
+		details = append(details, fmt.Sprintf("Successfully set \"%s\" as a project environment variable", variableName))
 	}
 
 	clients.IO.PrintTrace(ctx, slacktrace.EnvSetSuccess)
--- a/cmd/env/list.go
+++ b/cmd/env/list.go
@@ -118,7 +118,7 @@ func runEnvListCommandFunc(
 
 	details := []string{
 		fmt.Sprintf(
-			"There %s %d %s stored in this environment",
+			"There %s %d %s set in this environment",
 			style.Pluralize("is", "are", count),
 			count,
 			style.Pluralize("variable", "variables", count),
--- a/cmd/env/unset.go
+++ b/cmd/env/unset.go
@@ -119,7 +119,7 @@ func runEnvUnsetCommandFunc(clients *shared.ClientFactory, cmd *cobra.Command, a
 		}
 		selected, err := clients.IO.SelectPrompt(
 			ctx,
-			"Select a variable to remove",
+			"Select a variable to unset",
 			variables,
 			iostreams.SelectPromptConfig{
 				Flag:     clients.Config.Flags.Lookup("name"),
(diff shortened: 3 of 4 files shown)


cmd/env/set.go
```diff
@@ -137,14 +138,7 @@ func runEnvAddCommandFunc(clients *shared.ClientFactory, cmd *cobra.Command, arg
...
 		}
-		clients.IO.PrintTrace(ctx, slacktrace.EnvAddSuccess)
-		clients.IO.PrintInfo(ctx, false, "\n%s", style.Sectionf(style.TextSection{
-			Emoji: "evergreen_tree",
-			Text:  "App Environment",
-			Secondary: []string{
-				fmt.Sprintf("Successfully added \"%s\" as an app environment variable", variableName),
-			},
-		}))
+		details = append(details, fmt.Sprintf("Successfully added \"%s\" as an app environment variable", variableName))
```
mwbrooks: question: Should we start using the terms "set" and "unset" based on PR #460? While using another word like "added" might solidify what's happening, I think "set" is more accurate because it may be adding or updating the key.

cmd/env/add.go
```diff
@@ -154,17 +148,17 @@ func runEnvAddCommandFunc(clients *shared.ClientFactory, cmd *cobra.Command, arg
...
-			Emoji:     "evergreen_tree",
-			Text:      "App Environment",
-			Secondary: details,
-		}))
 	}
+
+	clients.IO.PrintTrace(ctx, slacktrace.EnvAddSuccess)
+	clients.IO.PrintInfo(ctx, false, "\n%s", style.Sectionf(style.TextSection{
+		Emoji:     "evergreen_tree",
+		Text:      "Environment Add",
```
mwbrooks: question: Based on #460 what about? 
* **Set Environment Variable**
* **Unset Environment Variable**
* **List Environment Variables** _(plural)_
```suggestion
		Text:      "Set Environment Variable",
```

cmd/env/remove.go
```diff
@@ -107,7 +107,7 @@ func runEnvRemoveCommandFunc(clients *shared.ClientFactory, cmd *cobra.Command,
 			clients.IO.PrintTrace(ctx, slacktrace.EnvRemoveSuccess)
 			clients.IO.PrintInfo(ctx, false, "\n%s", style.Sectionf(style.TextSection{
 				Emoji: "evergreen_tree",
-				Text:  "App Environment",
+				Text:  "Environment Remove",
```
srtaalej:
```suggestion
				Text:  "Environment unset", 
``` 
**edit**- Saw michaels suggestion above and I like "Unset environment variable"!

cmd/env/add.go
zimeg: @mwbrooks Haha I realize these branches overlap! Would we want to use:

- **Environment Set**
- **Environment List**
- **Environment Unset**

Instead? I agree this reads odd but matching the command name might let use expand more in details that follow I hope 📚

cmd/env/set.go
zimeg: @mwbrooks @srtaalej Noted in #460 also that "remove" might have a few places to change! I agree with both comments and mirror this here to not forget 👾 

> looking great!! i noticed there are still a few references to removing in the file 👁️‍🗨️
> line 120 ["Select a variable to remove"](https://github.com/slackapi/slack-cli/blob/7a888e0782bbee8c1cc3d9c6b047c980d767abc7/cmd/env/unset.go#L120)
> line 150 ["Successfully removed \"%s\" from the app's environment variables"](https://github.com/slackapi/slack-cli/blob/7a888e0782bbee8c1cc3d9c6b047c980d767abc7/cmd/env/unset.go#L150)
> might be nice to update these too 🚀

zimeg: 🧪 note: This will require updates to our E2E tests too!

cmd/env/remove.go
zimeg: @srtaalej Thanks for these callouts too! I'm leaning to keep this in command format as:

```
🌲 Environment Unset
```

Which will match our current style guide:

https://github.com/slackapi/slack-cli/blob/def332f1102db5a728a5cd969068f7b5db701d43/.github/STYLE_GUIDE.md#L65-L79

zimeg:
@mwbrooks @srtaalej Both amazing thanks for the reviews and share eyes shared 👁️‍🗨️ ✨ 

I'm merging this once tests pass with changes that:
- Use command headers
- Replace wording with "set" instead of "save" or "add" alternatives

Thanks again for bringing polish to this experience 🙏

PR #463, opened 2026-04-02
chore!: remove `workspace` and `team` aliases from the `app` command

zimeg:
### Changelog

> We removed the aliases `slack workspace` and `slack team` to now prefer the `slack app` command when managing or reviewing app installations of a project.

### Summary

🪓 This PR removes the `workspace` and `team` aliases from the `app` command to keep this common focused on app installations.

### Requirements

### DIFF
--- a/.github/STYLE_GUIDE.md
+++ b/.github/STYLE_GUIDE.md
@@ -5,7 +5,9 @@ A current suggestion for how Slack CLI inputs are handled and outputs are format
 - **Input**
   - [Prompts are Flags with Forms](#prompts-are-flags-with-forms)
 - **Output**
-  - [Format Sections with Command Details](#format-sections-with-command-details)
+  - [Help Arguments use Opinionated Brackets](#help-arguments-use-opinionated-brackets)
+  - [Help Descriptions find Complete Sentences](#help-descriptions-find-complete-sentences)
+  - [Section Formats with Command Headings](#section-formats-with-command-headings)
 
 ## Input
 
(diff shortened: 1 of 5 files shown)


cmd/app/app.go
```diff
@@ -25,7 +25,7 @@ import (
 func NewCommand(clients *shared.ClientFactory) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:     "app",
-		Aliases: []string{"workspace", "app", "apps", "team", "teams", "workspaces"},
```
zimeg, line 28: 👾 note: It was unclear if the "team" aliases are right to remove but I think it makes sense to remove these with this change!

mwbrooks: Long history but it kinda shows the evolution of this command from being designed for "workspaces" before supporting Enterprises with a generic "teams" before realizing that we need a way to manage the specific app and not just the team that the app is installed to with "apps".

In hindsight, the CLI has no use-case today to do CRUD operations on a Workspace or Enterprise Org. That's the admin world, which we haven't feature requests around.

I'm glad to see these aliases gone though, because they're so old that they've lost their meaning.

zimeg: @mwbrooks Thank you for sharing this! That's a most interesting start and use case considered 📚 

Now when I think about `team` or `workspace` the credentials I have saved is what I'd want to find. We have the command for showing an overview I prefer as:

```sh
$ slack auth list
```

Which can also be nice for collecting a service token:

```sh
$ slack auth token
```

Thanks also for confirming these aliases are alright for error at this time 🌚

zimeg:
@mwbrooks Thanks for the review and details of the command 🙏 ✨ 

I'm wanting to now merge this after tests complete and have noted this toward the next milestone **[Next Release](https://github.com/slackapi/slack-cli/milestone/24)** upcoming with the correct labels as I found - noting a `major` label included 🫡

Issue #462, opened 2026-04-02
Tests fail when using fish (and possibly other shells)


artlogic:
### Description

When attempting to run tests under fish, the following errors are incorrectly thrown:

```
--- FAIL: TestDeployCommand_DeployHook (11.06s)
    --- FAIL: TestDeployCommand_DeployHook/fails_to_execute_an_unknown_script_path (2.31s)
        deploy_test.go:294:
                Error Trace:    github.com/slackapi/slack-cli/cmd/platform/deploy_test.go:294
                Error:          Should be true
                Test:           TestDeployCommand_DeployHook/fails_to_execute_an_unknown_script_path
    --- FAIL: TestDeployCommand_DeployHook/echos_stderr_output_to_standard_err (2.20s)
        deploy_test.go:296:
                Error Trace:    github.com/slackapi/slack-cli/cmd/platform/deploy_test.go:296
                Error:          Received unexpected error:
                                Error running 'Deploy' command: exit status 127 (sdk_hook_invocation_failed)

                                Suggestion
                                   Run `platform.test doctor` to check that your system dependencies are up-to-date.
                Test:           TestDeployCommand_DeployHook/echos_stderr_output_to_standard_err
FAIL
FAIL    github.com/slackapi/slack-cli/cmd/platform      11.953s
```

This seems to be due to a dependency on particular shell syntax/messages in the tests.

### Version

Using slack-cli v3.15.0

### OS Info

Arch Linux (under WSL2)

#### Steps to reproduce:

Share the commands to run, source code, and project settings:

1. Build.
2. `go test ./...`

### Expected result:

I expected the tests to pass.

### Actual result:

The tests failed, as shown above.

### Requirements

* [x] I've read and understood the [Contributing guidelines](/blob/main/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).
* [x] I've searched for any related issues and avoided creating a duplicate issue.

artlogic:
> [@artlogic](https://github.com/artlogic) Ahhaha thanks for the note that the test command has to use the `fish` shell itself 🎣 ✨
> 
> I can confirm these errors and think the matrix of tests above might be added to! AFAICT, with the `SHELL` variable set, the `run` and the `deploy` command still work as expected so these errors might be limited to tests? 🧪

Yes, the errors seem to be limited to the tests. I would be happy to provide a potential fix in a PR if that would be acceptable?

PR #464, opened 2026-04-02
ci: run unit tests on supported os runners

zimeg:
### Changelog

> N/A - Testing changes.

### Summary

This PR attempts to resolve issues noted in #462 using logs of CI from these changes 🧪 

We hope to add unit test checks for all supported OS:

- Mac
- Linux (Ubuntu)
- Windows

### Requirements

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -40,7 +40,12 @@ jobs:
           version: latest
           args: --timeout=5m
       - name: Unit Tests
+        if: runner.os == 'Linux'
         run: make test
+      - name: Unit Tests
+        if: runner.os != 'Linux'
+        run: go test -v ./... -race -covermode=atomic -coverprofile=coverage.out
+        shell: bash
       - name: Install Tests
         if: runner.os != 'Windows'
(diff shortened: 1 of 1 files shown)


zimeg:
Adjacent change worth coordinating with: #652 (the Pscx `Expand-Archive` fix for #651) also edits `.github/workflows/tests.yml` — it adds an `install-tests` matrix across `macos-latest` / `ubuntu-latest` / `windows-latest` and splits it out of the `lint-test` job. Since this PR is adding OS-matrix coverage to the same file, the two will want to rebase on each other rather than land independently.

One concrete signal from #652 that's relevant here: the Windows runner does run CI fine, but the installer's post-install courtesy check hangs under pwsh 7 on the current `windows-latest` image (the native-command `_fingerprint | Tee-Object -Variable | Out-Null` pipeline blocks with no console). #652 works around it with a static AST guard rather than a live install. Worth keeping in mind if the Windows unit-test leg here ends up shelling anything similar — the runner is viable, but native-command pipelines that expect a console are the sharp edge.

PR #469, opened 2026-04-03
feat(experiment): add apps.icon.set support for non-hosted app icons

srtaalej:
### Changelog

> _No user-facing changes until `apps.icon.set` endpoint is live_

### Summary

This PR adds `IconSet()` API method targeting the `apps.icon.set` endpoint for non-hosted (Bolt) app icons. Gated behind the set-icon experiment flag (`--experiment=set-icon`) (see Note)
### Expected Behavior / Test Plan
`slack deploy` → updateIcon() uses Icon() (no experiment needed, hosted apps, existing behavior)  
`slack deploy --experiment set-icon` → updateIcon() uses IconSet() (both app types)             
`slack run --experiment set-icon` → InstallLocalApp icon block runs via IconSet() (both app types)
`slack run` (no experiment) → no icon upload 

### Migration plan                                                                                
When `apps.icon.set` is generally available:                                                      
1. Remove the `set-icon` experiment gate from `updateIcon` and `InstallLocalApp`                  
2. Replace all `Icon()` calls with `IconSet()`                                                    
3. Delete the deprecated `Icon()` method, its tests, mock, and the `apps.hosted.icon` constant    
       
### Note

> 🔴  The apps.icon.set endpoint is available under a toggle. No changes were made to the existing `apps.hosted.icon` flow.
> Required scope (`app_configurations:write`) may be subject to change before the endpoint goes public                                                   

### Requirements


internal/api/icon.go
```diff
@@ -35,6 +35,8 @@ import (
 
 const (
 	appIconMethod = "apps.hosted.icon"
+	// AppIconSetMethod is the API method for setting app icons for non-hosted apps.
+	AppIconSetMethod = "apps.icon.set"
```
zimeg:
```suggestion
	appIconSetMethod = "apps.icon.set"
```

🔭 suggestion: Let's keep this private to this package!

internal/api/icon.go
```diff
@@ -48,6 +50,16 @@ type iconResponse struct {
 
 // Icon updates a Slack App's icon
```
zimeg:
```suggestion
// Icon updates a hosted Slack app icon
// DEPRECATED: Prefer "IconSet" instead
```

📝 note: The "Icon" method is not so descriptive! I'm unsure if changing that is needed for this PR but we might want to note of the preferred method here.

internal/pkg/apps/install.go
zimeg:
```suggestion
	_, err = clients.API().SetIcon(ctx, clients.Fs, token, appID, iconPath)
```

🪬 question: Would we want to instead update this callsite to prefer the new method instead? I understand this is meant to replace the hosted icon upload altogether?

zimeg (commented):
🐣 A few quick comments and questions and I'm excited to test this next!

internal/pkg/apps/install.go
srtaalej: SetIcon doesnt exist did you mean IconSet()?

zimeg: @srtaalej Ah apologies! If that's the same for `apps.icon.set` I did! 🤓

zimeg: 🐏 note: The suggestion above was made quick too - we perhaps would want to guard against experimental changes in stable code still here too!

internal/api/icon.go
```diff
@@ -35,6 +35,8 @@ import (
 
 const (
 	appIconMethod = "apps.hosted.icon"
+	// appIconSetMethod is the API method for setting app icons for non-hosted apps.
+	appIconSetMethod = "apps.icon.set"
```
zimeg, line 39: 🔬 issue: This outputs a warning to verbose logs and it's not clear what I might not be including?

```
uploading icon
icon error: https://slack.com/api/apps.icon.set error: missing_arguments
   Error updating app icon: https://slack.com/api/apps.icon.set error: missing_arguments
   Finished in 0.7s
```

zimeg (changes requested):
@srtaalej Right now I'm not having luck with this when testing in a shared team sandbox and the error isn't so obvious to me...

Let me know if this is something perhaps with the image I use? I'll include it here for testing and other reviewers:

Image

internal/api/icon.go
srtaalej: are you uploading through the manifest or is the image just in the root of the project?

zimeg: @srtaalej I saved this file to the root of the project 🌲 ✨

srtaalej: the "file" fielf for the api request had the wrong name! can you try again ?

internal/pkg/apps/install.go
```diff
@@ -521,30 +522,25 @@ func InstallLocalApp(ctx context.Context, clients *shared.ClientFactory, orgGran
...
+	if clients.Config.WithExperimentOn(experiment.SetIcon) {
+		var iconPath = slackManifest.Icon
+		if iconPath == "" {
+			if _, err := os.Stat("icon.png"); !os.IsNotExist(err) {
+				iconPath = "icon.png"
+			}
+		}
+		if iconPath != "" {
+			clients.IO.PrintDebug(ctx, "uploading icon")
+			_, iconErr := clients.API().IconSet(ctx, clients.Fs, token, app.AppID, iconPath)
```
zimeg, line 535: 🔬 question: Could we match debug log outputs for the `apps.icon.set` method here? Right now I find this in verbose:

```
uploading icon
```

While other methods include various request and response details that are useful in debugging. I think edges in image upload might make verbose logging most important here!

internal/experiment/experiment.go
zimeg, line 1: 📚 note: Let's document this too!

https://docs.slack.dev/tools/slack-cli/reference/experiments

internal/api/icon_test.go
```diff
@@ -28,65 +28,221 @@ import (
...
-		ExpectedMethod: appIconMethod,
-		Response:       `{"ok":true}`,
-	})
-	defer teardown()
-	_, err = c.Icon(ctx, fs, "token", "12345", imgFile)
+	err = png.Encode(myfile, myimage)
 	require.NoError(t, err)
 }
+
+func TestClient_Icon(t *testing.T) {
```
zimeg, line 46: 🧪 praise: Thank you for updating existing tests to a familiar format!

internal/api/icon.go
```diff
@@ -81,7 +94,7 @@ func (c *Client) Icon(ctx context.Context, fs afero.Fs, token, appID, iconFilePa
 
 	var part io.Writer
 	h := make(textproto.MIMEHeader)
-	h.Set("Content-Disposition", fmt.Sprintf(`form-data; name="%s"; filename="%s"`, "file", iconStat.Name()))
+	h.Set("Content-Disposition", fmt.Sprintf(`form-data; name="%s"; filename="%s"`, fileFieldName, iconStat.Name()))
```
zimeg:
```suggestion
	h.Set("Content-Disposition", fmt.Sprintf(`form-data; name="%s"; filename="%s"`, iconStat.Name()))
```

🌚 suggestion: Let's revert the addition of `fileFieldName` here and in arguments too if this is static?

zimeg (approved):
@srtaalej Woot! Thanks for finding these fixes! 📸

I'm approving now with a few notes of follow up that might be interesting. But testing so far is alright for me and fast enough on initial upload and re-uploads at ~4 seconds and ~2 seconds! Caching might not be an immediate concern perhaps? Notes:

- The default "icon" field might be meaningful to add to the app manifest [schema](https://github.com/slackapi/manifest-schema) for developers with custom paths wanting editor hints?
- The `InstallLocalApp` and the `Install` functions are now so similar enough to perhaps collapse into one function? Not urgent but it'd be a meaningful follow up to remove more from `pkg` internal 🎁 

IMHO logging is important to land before this exits experiments but no blockers for this PR outside of note to document the experiment! 📚

internal/pkg/apps/install.go
srtaalej: since this is a gap in both apps.icon.set and apps.hosted.icon I'm going to leave this for a followup PR 🫡

PR #506, opened 2026-04-20
fix: avoid icon upload without experiment for non-hosted apps

srtaalej:
## Summary
- Without the `set-icon` experiment enabled, non-hosted (Bolt) apps were still attempting to upload an icon via `apps.hosted.icon`, which is only valid for hosted apps
- This fix routes icon uploads by app type: experiment on → `apps.icon.set` (works for all app types), experiment off + hosted → `apps.hosted.icon`, experiment off + non-hosted → skip
- Adds a `DEPRECATED` comment on the `apps.hosted.icon` path per reviewer suggestion

## Test plan
```sh
$ ./bin/slack run -e set-icon    # Bolt app: confirm icon uploads via apps.icon.set
$ ./bin/slack deploy             # Hosted app without experiment: confirm icon uploads via apps.hosted.icon
$ ./bin/slack run                # Bolt app without experiment: confirm icon upload is skipped
```
- `go test ./internal/pkg/apps/ -run Test_updateIcon` — 6 cases covering all routing paths and error handling

## Requirements


internal/pkg/apps/install.go
```diff
@@ -651,18 +651,20 @@ func appendLocalToDisplayName(manifest *types.AppManifest) {
...
 	var err error
-	if clients.Config.WithExperimentOn(experiment.SetIcon) {
+	if isHosted {
+		_, err = clients.API().Icon(ctx, clients.Fs, token, appID, iconPath)
+	} else if clients.Config.WithExperimentOn(experiment.SetIcon) {
 		_, err = clients.API().IconSet(ctx, clients.Fs, token, appID, iconPath)
 	} else {
-		_, err = clients.API().Icon(ctx, clients.Fs, token, appID, iconPath)
+		return nil
 	}
```
zimeg, line 663: 🪬 question: Are we hoping to have all uploads use `apps.icon.set`? I find this works as expected for hosted apps at the moment and am unsure that we need to change this.

zimeg (commented):
@srtaalej I left a comment if we want to continue supporting the `apps.hosted.icon` after the experiment concludes. It's unclear to me if we should continue to support paths that reference this method right now 🔭

internal/pkg/apps/install.go
srtaalej: hmm agreed - i guess i was unsure whether or not we were completely dropping `apps.hosted.icon` support 🤔 should the experiment only work with `apps.icon.set` then?

zimeg: @srtaalej I understand `apps.icon.set` to be the replacement and preferred method for updating an app icon now! Please correct me if this is wrong, but we should use that while the experiment is active if so 🧪

internal/pkg/apps/install.go
```diff
@@ -650,16 +650,18 @@ func appendLocalToDisplayName(manifest *types.AppManifest) {
...
 	var span opentracing.Span
 	span, ctx = opentracing.StartSpanFromContext(ctx, "updateIcon")
 	defer span.Finish()
 
 	var err error
 	if clients.Config.WithExperimentOn(experiment.SetIcon) {
 		_, err = clients.API().IconSet(ctx, clients.Fs, token, appID, iconPath)
-	} else {
+	} else if isHosted {
 		_, err = clients.API().Icon(ctx, clients.Fs, token, appID, iconPath)
```
zimeg, line 663:
```suggestion
		// DEPRECATED: Prefer IconSet once the SetIcon experiment concludes
		_, err = clients.API().Icon(ctx, clients.Fs, token, appID, iconPath)
```

🪬 suggestion(non-blocking): The order of conditionals is solid but we can leave breadcrumbs more to save confusion in upcoming changes. Personal preference though!

zimeg (approved):
@srtaalej Thanks for these updates! I understand the fix this PR is addressing now and changes LGTM 🚢

Before merging I'm wondering if some PR formalities can be polished? The confusion I had most was in the intent of this PR. The title explains what the code does but not why this change was needed. I might suggest:

> fix: avoid icon upload without experiment for non-hosted apps

Including a note in the description of when this might happen can be helpful in sharing about the issue as this is adjusting behavior in a current release. Perhaps with a `changelog` too?

The test plan also did not give me much to start with. Sharing the commands you used to test the development build with can make review faster for me if I can ask for that 🙏 

Something brief gives a lot of focus and we can save unit test steps for CI:

```sh
$ slack run -e set-icon  # Confirm icon uploads for Bolt app
```

We should also add a `bug` label and milestone to this before we merge so our release notes find it in the right spot.

PR #509, opened 2026-04-27
feat: support png, jpg, jpeg, gif for icon auto-detection

srtaalej:
## Changelog

```changelog
### Features
- Icon auto-detection now supports `.png`, `.jpg`, `.jpeg`, and `.gif` formats in `assets/` and project root (gated behind `set-icon` experiment)
- `assets/` directory is checked before project root; `.png` is preferred when multiple formats exist
```

## Summary
- Extracts icon resolution into a new `internal/icon` package with `ResolveIconPath`
- Searches for `icon.{png,jpg,jpeg,gif}` in `assets/` first, then project root — matching all formats the API accepts
- PNG has highest priority when multiple icon files exist
- Removes unused `hasValidIconPath` and `Verify` methods from `SlackYaml`

`semver:patch`

### Unit tests
```sh
go test ./internal/icon/ -v          # 12 cases: all formats, priority, manifest precedence, fallback paths
go test ./internal/pkg/apps/ -v      # install tests
go test ./...                        # full suite
```

### Manual testing

**1. Build the branch**
```sh
git fetch origin ale-icon-assets-fallback
git checkout ale-icon-assets-fallback
go build -o bin/slack .
```

**2. Set up a test app** (or use an existing bolt app)
```sh
./bin/slack create
cd <app-name>
```

**3. Test assets/ auto-detection with different formats**

Place an icon in `assets/` and run:
```sh
mkdir -p assets
cp /path/to/icon.png assets/icon.png
../bin/slack run -e set-icon
```
Verify: `Updated app icon: assets/icon.png`

Repeat with `assets/icon.jpg` (remove `icon.png` first) — should pick up the jpg.

**4. Test project root fallback**

Remove any icon from `assets/`. Place `icon.png` in the project root:
```sh
rm assets/icon.*
cp /path/to/icon.png icon.png
../bin/slack run -e set-icon
```
Verify: `Updated app icon: icon.png`

**5. Test priority (assets/ wins over root)**

Place icons in both locations:
```sh
cp /path/to/icon.png assets/icon.png
cp /path/to/icon.png icon.png
../bin/slack run -e set-icon
```
Verify: `Updated app icon: assets/icon.png` (assets/ takes precedence)

**6. Test manifest takes precedence in Deno apps**

Set `icon: "custom/my-icon.png"` in the manifest. Run again — should use the manifest path, not auto-detection.


internal/pkg/apps/install.go
```diff
@@ -40,6 +41,23 @@ const (
 
 const additionalManifestInfoNotice = "App manifest contains some components that may require additional information"
 
+var supportedIconExtensions = []string{".png", ".jpg", ".jpeg", ".gif"}
```
mwbrooks: question: Just curious, have you tested all of these images types? I just want to make sure that the API supports each.

internal/pkg/apps/install.go
```diff
@@ -40,6 +41,23 @@ const (
...
+	if manifestIcon != "" {
+		return manifestIcon
+	}
+	for _, dir := range []string{"assets", "."} {
+		for _, ext := range supportedIconExtensions {
+			candidate := filepath.Join(dir, "icon"+ext)
+			if _, err := fs.Stat(candidate); err == nil {
+				return candidate
+			}
+		}
```
mwbrooks: question: It looks like the implementation will attempt to find the first image that matches the file extensions and use it when there is no path defined?

internal/pkg/apps/install.go
```diff
@@ -40,6 +41,23 @@ const (
 
 const additionalManifestInfoNotice = "App manifest contains some components that may require additional information"
 
+var supportedIconExtensions = []string{".png", ".jpg", ".jpeg", ".gif"}
+
+func resolveIconPath(fs afero.Fs, manifestIcon string) string {
```
mwbrooks: suggestion: Thoughts on moving this to `internal/icon/icon.go` or maybe `internal/config/icon.go`. My preference is the first to avoid cluttering our already messy `package config`.

internal/shared/types/slack_yaml.go
```diff
@@ -27,15 +27,19 @@ type SlackYaml struct {
...
-			var defaultIconPath = "assets/icon.png"
-			if _, err := os.Stat(filepath.Join(wd, defaultIconPath)); !os.IsNotExist(err) {
-				sy.Icon = defaultIconPath
+		if sy.Icon == "" {
+			for _, ext := range supportedIconExtensions {
+				candidate := filepath.Join(wd, "assets", "icon"+ext)
+				if _, err := os.Stat(candidate); !os.IsNotExist(err) {
+					sy.Icon = filepath.Join("assets", "icon"+ext)
+					break
+				}
```
mwbrooks: suggestion: Since we're using this loop logic in 2 places, I think it would make sense to implement it as a function in `internal/icon/icon.go`.

mwbrooks (commented):
🙇🏻 Thanks for putting this PR together @srtaalej!

✏️ I've left a couple feedback questions and suggestions, but overall this is looking good.

❓ Curious, have you tested this on both Deno/ROSI and Bolt using multiple image formats for each?

❓ This implementation is using `manifest.json` to define the icon. Personally, I like how clean this is but it does put us in a pickle if someone tries to copy & paste the manifest into App Settings or use our manifest-schema package for validation. Are we planning to explore defining the icon in `config.json`?

internal/pkg/apps/install.go
srtaalej: yes thats the fallback. it doesnt have to be that way but im worried the `assets/icon.png` convention is a little strict since the image also has to be named icon

internal/pkg/apps/install.go
srtaalej: will do 🫡

srtaalej:
@mwbrooks 

> ❓ Curious, have you tested this on both Deno/ROSI and Bolt using multiple image formats for each?

i have tested with both Deno and Bolt - will update once all format types are tested!

srtaalej:
@mwbrooks Tested all supported image formats (png, jpg, jpeg, gif) against the API — all work as expected! 😄

internal/icon/icon.go
```diff
@@ -0,0 +1,38 @@
...
+
+package icon
+
+import (
+	"path/filepath"
+
+	"github.com/spf13/afero"
+)
+
+var SupportedExtensions = []string{".png", ".jpg", ".jpeg", ".gif"}
```
zimeg:
```suggestion
```

🪓 suggestion: If this isn't used in exports I'd initialize it within the function to avoid strange commands that might change it for some reason unexpected?

internal/icon/icon_test.go
```diff
@@ -0,0 +1,86 @@
...
+func Test_ResolveIconPath(t *testing.T) {
+	tests := map[string]struct {
+		manifestIcon string
+		files        []string
+		expected     string
+	}{
+		"manifest icon set returns it directly": {
+			manifestIcon: "custom/my-icon.png",
+			expected:     "custom/my-icon.png",
+		},
```
zimeg, line 34: 🧪 suggestion: Can we include a test that manifest icon is found before magic icons too?

internal/icon/icon_test.go
```diff
@@ -0,0 +1,86 @@
...
+		"assets/icon.jpeg found": {
+			files:    []string{"assets/icon.jpeg"},
+			expected: "assets/icon.jpeg",
+		},
+		"assets/icon.gif found": {
+			files:    []string{"assets/icon.gif"},
+			expected: "assets/icon.gif",
+		},
+		"png wins over gif in assets": {
+			files:    []string{"assets/icon.png", "assets/icon.gif"},
```
zimeg:
```suggestion
		"png wins over other extensions": {
			files:    []string{"assets/icon.jpg", "assets/icon.jpeg", "assets/icon.gif", "assets/icon.png"},
```

🧪 suggestion: Perhaps a separate test case but I'm wanting to guard against regression of the asset "icon.png" not being preferred?

internal/pkg/apps/install.go
```diff
@@ -524,12 +519,7 @@ func InstallLocalApp(ctx context.Context, clients *shared.ClientFactory, orgGran
...
 	if clients.Config.WithExperimentOn(experiment.SetIcon) {
-		var iconPath = slackManifest.Icon
-		if iconPath == "" {
-			if _, err := os.Stat("icon.png"); !os.IsNotExist(err) {
-				iconPath = "icon.png"
-			}
-		}
+		iconPath := icon.ResolveIconPath(clients.Fs, slackManifest.Icon)
 		if iconPath != "" {
 			_, iconErr := clients.API().IconSet(ctx, clients.Fs, token, app.AppID, iconPath)
```
zimeg, line 524: 🔭 thought: Unrelated to this PR but it's confusing to find the direct API call in this method but `updateIcon` called in the adjacent "install" method.

internal/shared/types/slack_yaml.go
```diff
@@ -28,28 +30,24 @@ type SlackYaml struct {
...
-		if sy.Icon == "" { // icon was not provided.  Let's check if the default one exists
-			var defaultIconPath = "assets/icon.png"
-			if _, err := os.Stat(filepath.Join(wd, defaultIconPath)); !os.IsNotExist(err) {
-				sy.Icon = defaultIconPath
-			}
-		} else {
-			if _, err := os.Stat(filepath.Join(wd, sy.Icon)); os.IsNotExist(err) {
-				return false
-			}
+func (sy *SlackYaml) hasValidIconPath(fs afero.Fs) bool {
```
zimeg:
```suggestion
```

🪓 suggestion: This function isn't used outside of tests AFAICT so we should delete it!

internal/shared/types/slack_yaml.go
```diff
@@ -28,28 +30,24 @@ type SlackYaml struct {
...
 		}
 	}
-
 	return true
 }
 
 // Verify checks that the app manifest meets some basic requirements
-func (sy *SlackYaml) Verify() error {
-	if !sy.hasValidIconPath() {
+func (sy *SlackYaml) Verify(fs afero.Fs) error {
```
zimeg:
```suggestion
```

🪓 suggestion: Similar this function isn't used outside of tests AFAICT and is more confusing than helpful I think? So I suggest removing it.

internal/shared/types/slack_yaml_test.go
zimeg: 🪓  suggestion: With the above removals we can perhaps save all of these tests for the new `icon` package?

zimeg (commented):
@srtaalej 📸 Nice! Before finding images to test with I want to leave a few comments of code changes that might improve ongoing maintenance!

zimeg:
📚 suggestion: A `changelog` for this new feature is important for our release notes so folks understand that images in these paths will be uploaded after updating to the latest version.

internal/icon/icon_test.go
srtaalej: yes good catch!

internal/icon/icon_test.go
srtaalej: true!

internal/shared/types/slack_yaml_test.go
srtaalej: 👍 yes

zimeg (approved):
@srtaalej LGTM! Thank you so much for cleaning code up alongside these changes!

I'm impressed to find `gif` uploads TBH! It's nice to find manifest updates also reupload the image! No changes more to request from me 🤩 

Image

PR #511, opened 2026-04-28
feat: include sandbox type when listing partner sandboxes

vegeris:
### Changelog

This does not need to be included in the Changelog

### Summary

This PR updates the Sandbox model to include the 'is_partner' bool which indicates whether this is a regular sandbox or a partner sandbox, and display it when listing one's partner sandboxes. We do not display the 'type' label for regular sandboxes to reduce visual noise

### Preview

Screenshot 2026-04-28 at 1 34 29 PM

### Testing

`% hermes sandbox list --experiment=sandboxes `

### Requirements

### DIFF
--- a/cmd/sandbox/list.go
+++ b/cmd/sandbox/list.go
@@ -125,6 +125,12 @@ func printSandboxes(cmd *cobra.Command, clients *shared.ClientFactory, token str
 			}
 		}
 
+		sandboxType := "Regular"
+		if s.IsPartner {
+			sandboxType = "Partner"
+		}
+		clients.IO.PrintInfo(ctx, false, "    %s", style.Secondary(fmt.Sprintf("Type: %s", sandboxType)))
+
 		if s.DateCreated > 0 {
 			clients.IO.PrintInfo(ctx, false, "    %s", style.Secondary(fmt.Sprintf("Created: %s", time.Unix(s.DateCreated, 0).Format(timeFormat))))
 		}
--- a/cmd/sandbox/list.go
+++ b/cmd/sandbox/list.go
@@ -16,6 +16,7 @@ package sandbox
 
 import (
 	"fmt"
+	"slices"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


cmd/sandbox/list.go
```diff
@@ -125,6 +125,12 @@ func printSandboxes(cmd *cobra.Command, clients *shared.ClientFactory, token str
 			}
 		}
 
+		sandboxType := "Regular"
+		if s.IsPartner {
+			sandboxType = "Partner"
+		}
+		clients.IO.PrintInfo(ctx, false, "    %s", style.Secondary(fmt.Sprintf("Type: %s", sandboxType)))
+
```
srtaalej: How would you feel about making the `Type` param optional? so that it only shows when there is a Partner sandbox 🤔 
since most sandboxes are regular the `Type` field adds visual noise to the output. so it might be helpful to only differentiate when there exists a partner sandbox

srtaalej (commented):
thank you for these updates ⭐ i left a comment about the output - i can make the change if you agree 💟

cmd/sandbox/list.go
vegeris: Sounds good to me 🙂 Particularly with Partners being a small percentage of developers

cmd/sandbox/list.go
```diff
@@ -125,6 +128,14 @@ func printSandboxes(cmd *cobra.Command, clients *shared.ClientFactory, token str
...
 		}
 
+		if hasPartner {
+			sandboxType := "Regular"
+			if s.IsPartner {
+				sandboxType = "Partner"
+			}
+			clients.IO.PrintInfo(ctx, false, "    %s", style.Secondary(fmt.Sprintf("Type: %s", sandboxType)))
+		}
+
```
vegeris: Thanks for making this update! 

From your earlier comment I was thinking we would want to only add this label if isPartner is true:
```
if s.IsPartner {
  clients.IO.PrintInfo(ctx, false, "    %s", style.Secondary("Type: Partner"))
}
```
(The API will return the is_partner param set to false for regular sandboxes)

srtaalej: hmm yes i guess my comment wasnt very clear - your version is simpler though so i'll go ahead and change it!

srtaalej (approved):

PR #512, opened 2026-04-28
feat: remove the experiment flag for the `sandbox` commands and add new partner template IDs

vegeris:
### Changelog

This is effectively the release of the `sandbox` commands and should be highlighted in the changelog:
```
slack sandbox create
slack sandbox delete
slack sandbox list
```

Details:

When creating a sandbox, users provide a name and domain for the sandbox, and a password for their user in the sandbox. Regular sandboxes can optionally use the 'default' or 'empty' templates. Partner developers can create partner sandboxes, which have some additional templates available to them (eg. 'finance' or 'hr' use cases). [The partner templates were added in this PR]

Users can view their sandboxes with the `list` command and delete sandboxes with the `delete` command.

### Summary

This PR removes the experiment flag currently required to run the sandbox commands

### Testing

Instead of:
```
lack sandbox list --experiment=sandboxes
```
We can run:
```
lack sandbox list
```

### Requirements


cmd/sandbox/sandbox.go
vegeris, line 1: At [line 80](https://github.com/slackapi/slack-cli/blob/main/cmd/sandbox/sandbox.go#L80) there was a recommendation to use charm for the 'Choose a Slack team where your email address matches...' info message; is charm stable yet, such that we can make that change?

cmd/sandbox/create.go
```diff
@@ -45,8 +45,16 @@ var createCmdFlags createFlags
...
 	"empty":   0, // The sandbox will be empty if the template param is not set
+	"default": 1, // The default template
+
+	// Additional partner sandbox template options
+	"finance":                2,
+	"hr":                     3,
+	"it-incident-management": 4,
+	"customer-support":       5,
+	"sales":                  6,
+	"marketing":              7,
```
vegeris, line 57: This PR also adds mappings for the new partner templates

mwbrooks: Thanks @vegeris. In the future, it would be better for this to be a separate PR so that it can stand out in the GitHub Release Notes and Changelog. 

suggestion: Can you please update the Changelog section of this PR Description so that the documentation team knows to document it on docs.slack.dev.

mwbrooks: suggestion: @vegeris Can you please add these to the Long Description so that developers can discover the partner template Ids that are available to use?

cmd/sandbox/sandbox.go
mwbrooks: Huh (Charm) is now the default and only prompt library. So, if you're using a prompt then you're now using Charm's Huh :)

internal/slackerror/errors.go
```diff
@@ -100,6 +100,7 @@ const (
 	ErrDotEnvFileRead                                = "dotenv_file_read_error"
 	ErrDotEnvFileWrite                               = "dotenv_file_write_error"
 	ErrDotEnvVarMarshal                              = "dotenv_var_marshal_error"
+	ErrDomainLong                                    = "domain_long"
```
mwbrooks: suggestion: Can this please be prefixed with `ErrSandboxDomainLong = "sandbox_domain_long"` so that we know it's part of the Sandbox error codes. You'll need to reorganize it alphabetically.

mwbrooks (changes requested):
🙏🏻 Thanks for removing the experiment @vegeris

✏️ I've left a few change request suggestions. In the future, I'd encourage the PR to only remove the experiment and add other additions in separate PRs. It allows both the documentation team to spot changes easier and our GitHub Release Notes (based on PR titles) to communicate the changes to our developers.

cmd/sandbox/sandbox.go
srtaalej: yea perhaps we get rid of that comment?

internal/slackerror/errors.go
vegeris: 💯 Updated this error code and alphabetized the sandbox error codes added previously!

cmd/sandbox/sandbox.go
vegeris: Ah, gotcha :) Updated to use the 'Help' config for the prompt instead of a separate log statement

cmd/sandbox/create.go
vegeris: I wasn't sure if we would want to highlight it in the Long Description as the grand majority of users would be regular non-partner developers; for the `list` command for example, we [only](https://github.com/slackapi/slack-cli/pull/511#discussion_r3154917269) add the 'sandbox type' label to partner sandboxes to reduce visual noise for regular sandboxes

> document it on docs.slack.dev

Speaking of, I'd like to double check the process for getting new commands documented. My current assumption was that we would have a brief announcement in the Changelog through this PR and that once these changes are included in a release, that I would post in #api-docs to get the commands documented on [docs.slack.dev](https://docs.slack.dev/tools/slack-cli/reference/commands/slack)?

vegeris: Also, if there's a strong preference I can apply the changes unrelated to removing the experiment flag first and then come back to this PR

internal/slackerror/errors.go
zimeg: 🚧 question(non-blocking): Are these error codes from the upstream APIs? I'm hesitant to prefix the error variables with "sandbox" but not the error code.

cmd/sandbox/create.go
zimeg: > process for getting new commands documented

@vegeris The `changelog` included in the PR description might be alright to find this with next release! But let's keep watch of the published release! Reference should be generated automatic FWIW.

Perhaps now we can also update the **experiments** page to note the conclusion?

🔗 https://docs.slack.dev/tools/slack-cli/reference/experiments

zimeg: 👾 ramble: I might also agree with separate PRs for future changes to make review more "obvious" for me with a single focused mind but I understand this now 🌚

zimeg (approved):
@vegeris So excited this made it to the finish line - superb 🏁 

I'm leaving a few comments on changes here because inline [comments are not working right](https://www.githubstatus.com/incidents/f5pb5d5mr9yh) now?

- The error codes for `sandbox_already_deleted` and `password_too_short` don't have messages included with the codes and I'm not sure if that's something for the backend or CLI?
- The sandbox `create` password prompt appears as plain text when I expected it to be hidden?

Otherwise nothing blocking! The above might be intentional or for another PR too. Whenever is best please feel free to merge for this upcoming release! 🔮 ✨ 

Edit: Oopies looks like comments are truncated too! I will continue this but LGTM 🚢

cmd/sandbox/create.go
mwbrooks: > I wasn't sure if we would want to highlight it in the Long Description as the grand majority of users would be regular non-partner developers

I feel we under utilitize the Long Description at the moment. We should feel comfortable documenting commands at length in the Long Description. After all, it's only shown when a developer wants to learn about that command, specifically. It's also generated into a webpage on docs.slack.dev.

All of that is to say, feel free to go full essay in a Long Description.

mwbrooks: > My current assumption was that we would have a brief announcement in the Changelog through this PR and that once these changes are included in a release, that I would post in #api-docs to get the commands documented on docs.slack.dev?

Yep, this is a good approach. Typically, the docs team is good at adding documentating while reviewing our releases changelog. However, a post in #api-docs will give them an earlier heads up. 

All help docs are generated into references pages, automatically. But the docs team may decide to create dedicated guides for sandbox management.

internal/slackerror/errors.go
```diff
@@ -55,6 +55,7 @@ const (
 	ErrAppNotHosted                                  = "app_not_hosted"
 	ErrAppRemove                                     = "app_remove_error"
 	ErrAppRenameApp                                  = "app_rename_app"
+	ErrAtActiveSandboxLimit                          = "at_active_sandbox_limit"
```
mwbrooks: suggestion: I'd like to see all of these new error codes prefixed with sandbox, similar to how our other error codes are prefixed with app, auth, etc. While our error codes aren't perfect, we're trying to keep the new ones more organized.

They should also be organized alphabetically after renamed.

For example:
```suggestion
	ErrSandboxAtActiveLimit                          = "sandbox_at_active_limit"
```

mwbrooks (commented):
📝 Thanks for the edits @vegeris! 🙇🏻 

I've left a few more comments. The largest one is renaming the error codes (go variables and error code strings) to be prefixed with sandboxes, unless they're meant to be used by other areas of the CLI codebase.

🪙 🪙 I think this has been mentioned, but in the future please try to keep PRs focused. This PR should be focused on removing the experiment flag, but it's doing quite a bit more. It makes reviewing the PR difficult and the CHANGELOG will fail to capture what true scope of the PR.

internal/slackerror/errors.go
mwbrooks: Just looping around on this PR. It looks like these error codes are returned by the Slack API, so we can't prefix them with `sandbox_`. Instead, I've added an inline comment stating that the codes are from the Slack API. Commit 3fb96eab4b3bff0333c87170eca3105e11747d82

mwbrooks (approved):
✅ I've added a commit to follow up on the last bit of feedback.

🙇🏻 Thanks for the wonderful `sandbox` command @vegeris! 🎉

PR #516, opened 2026-04-30
feat: add env init command that copies from template placeholders

zimeg:
### Changelog

> A new `env init` command copies a template placeholder ".env.sample" or ".env.example" file to the ".env" file of a project.

### Summary

This PR adds the `env init` command that copies from template placeholders 🌲 

### Preview

Included are expected cases and common edges to encounter! 🔬 ✨ 

#### ⚡ Bolt Frameworks

1. An existing placeholder file is used for the new ".env" file
2. An existing ".env" file is not overwritten
3. A missing placeholder file notes to use "env set" command

bolt

#### 🥀 ROSI Features

1. Deployed apps recommend the "env set" command
2. Local development requires selection and saves to project file

rosi

### Testing

Checkout the changes of this branch and experiment with either a Bolt or Deno app:

```sh
$ slack create asdf
$ cd asdf
$ slack install     # Required for Deno - Optional for Bolt
$ cat .env.sample   # Inspect values upcoming
$ slack env init
$ cat .env
$ slack env init    # Reattempt the initilization
```

### Notes

Some follow up remains proposed to improve this but is left out of scope for this PR 🏁 

- Prompting for values included in placeholders might be a meaningful improvement to avoid unexpected values. We might find the following format interesting toward this but without much exploration right now:

```sh
# COMMENTED_VARIABLE=       # Not prompted
UNCOMMENTED_VARIABLE=       # Prompted
SET_VARIABLE=debug_example  # Not prompted
```

- Including this command as part of `project init` ought make setup a breeze it's been said 😉 

### Requirements


cmd/env/env.go
```diff
@@ -48,6 +48,10 @@ func NewCommand(clients *shared.ClientFactory) *cobra.Command {
 			`Explore more: {{LinkText "https://docs.slack.dev/tools/slack-cli/guides/using-environment-variables-with-the-slack-cli"}}`,
 		}, "\n"),
 		Example: style.ExampleCommandsf([]style.ExampleCommand{
+			{
+				Meaning: "Initialize environment variables from template placeholders",
```
mwbrooks: suggestion: Minor suggestion to clarify that we're initializing from a file (plus it's shorter, haha).
```suggestion
				Meaning: "Initialize environment variables from a template file",
```

mwbrooks (commented):
🙇🏻 Thanks a bunch for opening this PR! 🎉 I'm excited to see this one landing.

🧠 Could we swap "template placeholders" for "template file" or ".env template file"? "Placeholder" tends to imply a variable or token that gets substituted, which is a bit ambiguous here. Since we're copying the file verbatim, ".env template file" is more precise and the common convention in the wild from a little Claude research.

cmd/env/env.go
zimeg: 🌚 @mwbrooks A kind suggestion to share! Changes toward this match https://github.com/slackapi/slack-cli/pull/516/commits/4eea43fdca7a552b8d0246c3c5a4d3e9066d6322!

zimeg:
@mwbrooks Apologies for the pause after review 🎁 ✨ 

The polish to how these concepts landed above and I think it puts this in a better place for ongoing iteration! Let's not think about that right now though. I requested another review for sureness of how this appears 🔮

internal/slackdotenv/slackdotenv.go
```diff
@@ -29,6 +29,42 @@ import (
 	"github.com/spf13/afero"
 )
 
+// Init copies an .env template file (.env.sample or .env.example) to .env. It
+// returns an error if .env already exists, or if no template file is found.
+func Init(fs afero.Fs) (string, error) {
+	sampleFiles := []string{".env.sample", ".env.example"}
```
mwbrooks, line 35: note: I had Claude do some research on github.com and these are the top 2 placeholder names for `.env` files and make up the largest percentage. So, I think this is a great start!

internal/slackerror/errors.go
```diff
@@ -96,9 +96,11 @@ const (
 	ErrDenoNotFound                                  = "deno_not_found"
 	ErrDeployedAppNotSupported                       = "deployed_app_not_supported"
 	ErrDocumentationGenerationFailed                 = "documentation_generation_failed"
+	ErrDotEnvFileAlreadyExists                       = "dotenv_file_already_exists"
```
mwbrooks, line 100: praise: error code prefixes for clean organization! 🎉

mwbrooks (approved):
✅ Thanks for the refactors @zimeg! This PR looks great and works like a charm! 👌🏻 

🧪 Manual tests are looking good and unit tests have good test coverage.

📝 A few follow-ups that come to mind:

1. suggestion: We should include `env init` as  part of `project init` so that the new projects come with a `.env` provisioned out-of-the-box.
2. question: When no `.env` exists and we run `env set`, the CLI creates `.env`. Should it create a blank `.env` or use `.env.example/sample` when it exists? My first impression is that we'd remain consistent if we tempalted from the `.env.example/sample`.

internal/slackdotenv/slackdotenv.go
zimeg: @mwbrooks Thanks for investigating too! IIRC our samples prefer `.env.sample` most and I'm willing to extend this as requested 🫡

internal/slackerror/errors.go
zimeg: @mwbrooks The prefix matches the package too! I hope we can continue this pattern as errors are wrapped across boundaries 🎁 ✨

zimeg:
@mwbrooks Super duper appreciate this being brought to the finish line! Thanks for sharing suggestions and feedback ahead 🏁 

I like both ideas if this `init` command continues to avoid prompting:

- During `project init`: We save the `cp .env.sample .env` command without setting values. IIRC projects might expect some values but emptiness should continue to error?
- Before `env set`: Similar great ideas. It makes "init" an invisible command and I like that because it can't be run after "set" so missing values would be a bummer.

Let's soon merge this for upcoming release I think! 🚢 💨

PR #519, opened 2026-05-01
feat: support SLACK_CLI_APP_ICON_PATH env var for icon override

srtaalej:
## Changelog

Add `SLACK_CLI_APP_ICON_PATH` environment variable to override the icon file path used during app install and local run. When set, the specified path takes priority over the manifest `icon` field and the default icon file search. A warning is displayed if the configured path does not exist.

## Summary
- Add `SLACK_CLI_APP_ICON_PATH` environment variable support to override the app icon path at install/deploy time
- Enables setting different icons per environment (dev/staging/prod) without changing project files
- Priority chain: `SLACK_CLI_APP_ICON_PATH` env var > manifest `icon` field > `assets/` and root fallback (via `icon.ResolveIconPath`)
- Extracts `resolveIconPath` wrapper in `install.go` that adds env var priority + path validation with warnings on top of `icon.ResolveIconPath`

### Unit tests
- [x] `go test ./internal/config/ -run Test_DotEnv` — env var loading (set, empty, whitespace trimming)
- [x] `go test ./internal/pkg/apps/ -run Test_resolveIconPath` — priority chain scenarios (env var, manifest, assets fallback, root fallback, missing paths)

### Manual testing

From a Slack app project directory:

1. **Override via env var**
   ```bash
   export SLACK_CLI_APP_ICON_PATH=icon.png
   ./bin/slack run -e set-icon
   ```
   Confirm output shows `Updated app icon: icon.png`

2. **Invalid env var path shows warning**
   ```bash
   export SLACK_CLI_APP_ICON_PATH=/nonexistent/icon.png
   ./bin/slack run -e set-icon
   ```
   Confirm warning: `icon path from SLACK_CLI_APP_ICON_PATH not found`

3. **Manifest icon field used when env var unset**
   ```bash
   unset SLACK_CLI_APP_ICON_PATH
   ./bin/slack run -e set-icon
   ```
   If manifest returns `"icon": "path/to/icon.png"`, confirm that path is used

4. **Fallback to `assets/icon.png` or `icon.png` in project root**
   With no env var and no manifest icon field, confirm icon auto-detection finds `assets/icon.png` (or `icon.png` in root)


internal/pkg/apps/install.go
```diff
@@ -649,6 +638,25 @@ func appendLocalToDisplayName(manifest *types.AppManifest) {
...
 
+// resolveIconPath determines the icon file path using the priority chain:
+// SLACK_CLI_APP_ICON_PATH env var > manifest icon field > icon.png in project root
+func resolveIconPath(ctx context.Context, clients *shared.ClientFactory, manifestIcon string) string {
+	if envIconPath := clients.Config.AppIconPathFlag; envIconPath != "" {
+		if _, err := os.Stat(envIconPath); !os.IsNotExist(err) {
+			return envIconPath
+		}
+		clients.IO.PrintDebug(ctx, "SLACK_CLI_APP_ICON_PATH file not found: %s", envIconPath)
+		_, _ = clients.IO.WriteOut().Write([]byte(style.SectionSecondaryf("Warning: icon path from SLACK_CLI_APP_ICON_PATH not found: %s", envIconPath)))
```
zimeg, line 649: ⚠️ issue: Falling through might be unexpected here if the `SLACK_CLI_APP_ICON_PATH` is set. IMHO we should print the warning and return no value instead?

internal/pkg/apps/install.go
```diff
@@ -649,6 +638,25 @@ func appendLocalToDisplayName(manifest *types.AppManifest) {
 	}
 }
 
+// resolveIconPath determines the icon file path using the priority chain:
+// SLACK_CLI_APP_ICON_PATH env var > manifest icon field > icon.png in project root
+func resolveIconPath(ctx context.Context, clients *shared.ClientFactory, manifestIcon string) string {
+	if envIconPath := clients.Config.AppIconPathFlag; envIconPath != "" {
+		if _, err := os.Stat(envIconPath); !os.IsNotExist(err) {
+			return envIconPath
+		}
```
zimeg: 🪬 question: Is this a check we should also include for `manifestIcon` values?

internal/pkg/apps/install.go
```diff
@@ -218,13 +218,7 @@ func Install(ctx context.Context, clients *shared.ClientFactory, auth types.Slac
 		}
 	}
 
-	// upload icon, default to icon.png
-	var iconPath = slackManifest.Icon
-	if iconPath == "" {
-		if _, err := os.Stat("icon.png"); !os.IsNotExist(err) {
-			iconPath = "icon.png"
-		}
-	}
```
zimeg: 📸 thought: Keeping this inline might be alright to avoided redirected logic?

⛈️ ramble: I might suggest we instead also extract the "upload" logic that follows this since it's duplicated too, but I'd much prefer the `Install` and `InstallLocalApp` functions be merged because the logic is almost identical if this is changing-

internal/config/config.go
```diff
@@ -22,6 +22,7 @@ import (
 )
 
 // Environment Variable constants
+const slackAppIconPathEnv = "SLACK_CLI_APP_ICON_PATH"
```
zimeg:
```suggestion
const slackCLIAppIconPathEnv = "SLACK_CLI_APP_ICON_PATH"
```

✏️ suggestion: To match the recasings!

internal/config/dotenv.go
```diff
@@ -45,6 +45,12 @@ func (c *Config) LoadEnvironmentVariables() error {
 		c.ConfigDirFlag = configDir
 	}
 
+	// Load app icon path from environment variables
+	var appIconPath = strings.TrimSpace(c.os.Getenv(slackAppIconPathEnv))
```
zimeg: 📠 issue: This might not capture values from the `.env` file also. I'm wondering if this extends to other values in this file so perhaps not a blocker?

internal/config/config.go
```diff
@@ -41,6 +42,7 @@ type Config struct {
 	APIHostFlag             string
 	APIHostResolved         string
 	AppFlag                 string
+	AppIconPathFlag         string
```
zimeg, line 47: 👾 thought: Interesting to find "app icon path" reads more clear than "CLI app icon path" but this ought not block decision on:

```diff
- SLACK_APP_ICON_PATH
+ SLACK_CLI_APP_ICON_PATH
```

🌲 ramble: I'm curious now again if "file" path needs to be specified in these values? Curious if path is clear enough on it's own...

zimeg (commented):
@srtaalej Sweeeet this is getting to a promising place 📸 

I'm leaving a few comments on fallback behavior and organization before testing more but I'd love to hear what you think about some of these notes?

internal/pkg/apps/install.go
```diff
@@ -649,6 +638,28 @@ func appendLocalToDisplayName(manifest *types.AppManifest) {
...
+			return envIconPath
+		}
+		clients.IO.PrintDebug(ctx, "SLACK_CLI_APP_ICON_PATH file not found: %s", envIconPath)
+		_, _ = clients.IO.WriteOut().Write([]byte(style.SectionSecondaryf("Warning: icon path from SLACK_CLI_APP_ICON_PATH not found: %s", envIconPath)))
+		return ""
+	}
+	if manifestIcon != "" {
+		if _, err := os.Stat(manifestIcon); !os.IsNotExist(err) {
+			return manifestIcon
+		}
```
zimeg, line 655:
```suggestion
		}
		clients.IO.PrintDebug(ctx, "manifest icon file not found: %s", manifestIcon)
		_, _ = clients.IO.WriteOut().Write([]byte(style.SectionSecondaryf("Warning: icon path from manifest not found: %s", manifestIcon)))
		return ""
```

🔬 note: To avoid falling through if errors occur?

zimeg (commented):
👋 @srtaalej Sharing another fallback case to return for but I'm curious still if we're wanting to read from ".env" files?

internal/config/dotenv.go
srtaalej: yup all other env vars are read from system !

internal/config/config.go
srtaalej: i can change it to CLIAppIconPath! i was matching it to the other vars here but it doesnt have to

zimeg (commented):
> 1. **Baseline — default icon via `slack run`**
>
> ```
> ./bin/slack run -e set-icon
> ```
> 
> Confirm output shows Updated app icon: assets/icon.png

⚠️ issue: When using the changes of https://github.com/slack-samples/bolt-js-starter-template/pull/145 I don't find the "assets/icon.png" is uploaded as expected?

```sh
$ slack create --branch add-app-icon -t slack-samples/bolt-js-starter-template
$ slack version
Using slack v4.0.1-ale-icon-env-var-15-g2778c01
```

Would this be a change for this PR or can I test this with another pattern? 💡

zimeg:
📚 suggestion: Would be good to note this as `changelog` as well for folks that've been uploading icons prior!

srtaalej:
> > 1. **Baseline — default icon via `slack run`**
> > 
> > ```
> > ./bin/slack run -e set-icon
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Confirm output shows Updated app icon: assets/icon.png
> 
> ⚠️ issue: When using the changes of [slack-samples/bolt-js-starter-template#145](https://github.com/slack-samples/bolt-js-starter-template/pull/145) I don't find the "assets/icon.png" is uploaded as expected?
> 
> ```shell
> $ slack create --branch add-app-icon -t slack-samples/bolt-js-starter-template
> $ slack version
> Using slack v4.0.1-ale-icon-env-var-15-g2778c01
> ```
> 
> Would this be a change for this PR or can I test this with another pattern? 💡

@zimeg thanks for raising this that was an issue with the pr description. this pr doesnt check for assets/, that gets added in 509 actually. i updated the pr description! but i think i should consolidate that check into this PR to since `resolveIconPath` function is in diff places in both PRs - 509 was updated with that before I opened 519 😿

internal/pkg/apps/install.go
```diff
@@ -639,6 +638,28 @@ func appendLocalToDisplayName(manifest *types.AppManifest) {
...
+	}
+	if manifestIcon != "" {
+		if _, err := clients.Fs.Stat(manifestIcon); err == nil {
+			return manifestIcon
+		}
+		clients.IO.PrintDebug(ctx, "manifest icon file not found: %s", manifestIcon)
+		_, _ = clients.IO.WriteOut().Write([]byte(style.SectionSecondaryf("Warning: icon path from manifest not found: %s", manifestIcon)))
+		return ""
+	}
+	return icon.ResolveIconPath(clients.Fs, "")
```
zimeg:
```suggestion
	return icon.ResolveIconPath(clients.Fs)
```

🪓 suggestion: Let's remove arguments that aren't used as part of these changes!

zimeg: 🪬 ramble: It's confusing to have multiple functions to resolve icon path functions IMHO and might consider the one of this file as a source of truth because we're writing output here. I'd be curious in refactoring the rest of the `icon` package inline if that's reasonable but no blocker for this PR!

zimeg (approved):
📸 @srtaalej LGTM! Thanks for separating these changes for more focused PRs!

Approving now with confidence and curiousness of potential code improvements to make, but nothing's blocking!

internal/pkg/apps/install.go
srtaalej: i agree! i figured i would leave that refacor for a separate pr though 🤔

PR #521, opened 2026-05-04
fix: preserve original app name as manifest display name

srtaalej:
### Changelog

Fix: App manifest display names now preserve the original app name instead of using the kebab-case directory name during `slack create`.

## Summary
- Separates the directory name (kebab-case) from the manifest display name (original user input) during app creation
- Manifest display names (`display_information.name`, `bot_user.display_name`, and Deno SDK `Manifest({ name })`) now preserve the user's original input instead of using the kebab-case directory name
- `package.json` and `pyproject.toml` names remain kebab-case per npm/Python conventions

### Behavior

```sh
$ slack create              # random name → path "vibrant-panther-540/", manifest "vibrant-panther-540"
$ slack create my-app       # path "my-app/", manifest "my-app"
$ slack create "My App"     # path "my-app/", manifest "My App"
```

## Test plan
- [x] `Test_App_UpdateDefaultProjectFiles` validates manifest files get the display name while package.json/pyproject.toml keep kebab-case
- [x] `Test_App_UpdateDefaultProjectFiles_WriteFileError` covers the WriteFile error path
- [x] `Test_RegexReplaceAppNameInManifest` updated to reflect display name input
- [ ] Manual: `slack create "My App"` → verify manifest display name is "My App", directory is `my-app/`, and package.json stays `my-app`


internal/app/app.go
```diff
@@ -46,18 +49,25 @@ func NewClient(
 	}
 }
 
+func kebabToTitleCase(s string) string {
+	return cases.Title(language.English).String(strings.ReplaceAll(s, "-", " "))
+}
+
 // UpdateDefaultProjectFiles should update any project specific files if any
 func UpdateDefaultProjectFiles(fs afero.Fs, dirPath string, appDirName string) error {
+	displayName := kebabToTitleCase(appDirName)
```
zimeg: 🔍 issue: We might want to avoid changing the app name format altogether if this is provided in flags? Forgive me if I'm not understanding requirements right, but this app manifest for example might break with replacements:

> Translator - Translate Languages

🔗 https://slack.com/marketplace/A8KHN4EDV-translator-translate-languages

zimeg: 🐍 ramble: Apps I keep might not fare well for additional example: "TIM", "TOM", or "snaek" have strange cases.

zimeg (changes requested):
@srtaalej Thanks for getting this started 👾 ✨ 

I'm hesitant on this approach and am wondering if we instead separate how arguments and flags are parsed at this point?

Some issues I ran into are in comments but I'd like to share how I think about the "create" command:

```sh
$ slack create           # Generates a random app name as current
$ slack create my-app    # Keeps path "my-app" and manifest "my-app"
$ slack create "my App"  # Uses path "my-app" and manifest "my App"
```

🧰 Additional cases might complicate this but it's a good time to untangle this perhaps? I'm curious also if adding the `--outdir` flag makes parts of this more clear, while still supporting the happiest path simple cases?

internal/pkg/create/create.go
```diff
@@ -150,7 +150,7 @@ func Create(ctx context.Context, clients *shared.ClientFactory, createArgs Creat
 	}()
 
 	// Update default project files' app name, bot name, etc
-	if err := app.UpdateDefaultProjectFiles(clients.Fs, projectDirPath, appDirName); err != nil {
+	if err := app.UpdateDefaultProjectFiles(clients.Fs, projectDirPath, appDirName, createArgs.AppName); err != nil {
```
zimeg, line 153: 🌟 praise: Nice! Keeping these separate makes it much easier to think about for me!

test/testdata/manifest-app-name.ts
```diff
@@ -12,7 +12,7 @@ export default {
     "major_version": 2
   },
   "display_information": {
-    "name": "vibrant-butterfly-1234"
+    "name": "Vibrant Butterfly 1234"
```
zimeg:
```suggestion
    "name": "Vibrant butterfly - 1234"
```

🧪 suggestion: Odd checks might improve our confidence in testing edge cases of expected behavior!

internal/app/app.go
```diff
@@ -47,17 +47,18 @@ func NewClient(
...
-		{"manifest.json", regexReplaceAppNameInManifest},
-		{"manifest.js", regexReplaceAppNameInManifest},
-		{"manifest.ts", regexReplaceAppNameInManifest},
-		{"package.json", regexReplaceAppNameInPackageJSON},
-		{"pyproject.toml", regexReplaceAppNameInPyprojectToml},
+		{"manifest.json", regexReplaceAppNameInManifest, displayName},
+		{"manifest.js", regexReplaceAppNameInManifest, displayName},
+		{"manifest.ts", regexReplaceAppNameInManifest, displayName},
+		{"package.json", regexReplaceAppNameInPackageJSON, appDirName},
+		{"pyproject.toml", regexReplaceAppNameInPyprojectToml, appDirName},
```
zimeg, line 61: 🧠 praise: Nice separation for project titles!

zimeg (commented):
@srtaalej This is running alright but I'm not certain the expected behavior matches the PR description and title since IIRC we're now avoiding title case defaults?

Leaving a few thoughts before approving and I also think a `changelog` toward this is best? Perhaps a standalone "bug" label too unless new featuresets are included too?

srtaalej:
@zimeg updated the pr description and tests!

internal/app/app_test.go
```diff
@@ -110,20 +118,11 @@ func Test_App_UpdateDefaultProjectFiles(t *testing.T) {
...
-		"WriteFile error": {
-			appDirName: "vibrant-butterfly-1234",
-			existingFiles: map[string]string{
-				"manifest.json": string(testdata.ManifestJSON),
-			},
-			expectedFiles: map[string]string{
-				"manifest.json": string(testdata.ManifestJSONAppName),
-			},
-			expectedErrorType: nil,
-		},
```
zimeg, line 126: 📠 question: Was this case patched? It's not clear to me why the test is removed?

zimeg (approved):
@srtaalej Thanks for polish toward the `create` command 🦋 

One question and hoping if this is toward next release let's include a changelog entry with the milestone 🚀 🔮

internal/app/app_test.go
srtaalej: that test case is identical to the "manifest.json file exists" test case - i removed it because it was redundant

srtaalej: i will add a test case that actually produces the write file error separately

PR #526, opened 2026-05-07
fix: preserve path separators in create command argument

srtaalej:
### Changelog

`semver:patch` — The `create` command preserves path separators when a nested path is provided as the project directory. The `--name` flag now only overrides the app's display name in the manifest without replacing the path argument.

```sh
$ slack create path/to/my-app --name "My App"
# Creates directory at ./path/to/my-app/ with manifest name "My App"
```

## Summary
- Fixes the `slack create path/to/my-app` regression where path separators were being replaced with dashes, creating a flat `path-to-my-app` directory instead of preserving the `path/to/my-app` structure
- Adds `parseAppPath` helper that splits user input into path prefix + basename, only kebab-casing the basename
- Display name for manifests is derived from the raw basename (not the full path)
- The `--name` flag now sets only the manifest display name, preserving any path from the positional argument

### Behavior

```sh
$ slack create my-app              # dir: ./my-app/        manifest: "my-app"
$ slack create "My App"            # dir: ./my-app/        manifest: "My App"
$ slack create path/to/my-app      # dir: ./path/to/my-app/ manifest: "my-app"
$ slack create path/to/My App      # dir: ./path/to/my-app/ manifest: "My App"
$ slack create path/to/my-app --name "APPPP"  # dir: ./path/to/my-app/ manifest: "APPPP"
```

Stacks on #521 which handles the display name vs dir name separation.

## Test plan
- [x] `TestParseAppPath` — 12 cases covering simple names, paths, absolute paths, dot-prefixed, trailing slashes, and error cases
- [x] Existing `TestGetProjectDirectoryName` still passes (single-component names)
- [x] Full test suite (`go test ./...`) passes
- [x] Updated `--name` flag tests to verify path is preserved and only display name is overridden
- [ ] Manual: `slack create path/to/my-app` with existing `path/to/` directory → verify nested structure created and manifest has `my-app`
- [ ] Manual: `slack create "My App"` → verify dir is `my-app`, manifest display is `My App`
- [ ] Manual: `slack create path/to/my-app --name "APPPP"` → verify dir is `path/to/my-app`, manifest display is `APPPP`


internal/iostreams/forms_test.go
```diff
@@ -485,6 +485,7 @@ func TestFormsAccessible(t *testing.T) {
 		assert.Equal(t, "", input)
 		assert.Contains(t, out.String(), "Name your app (default: cool-app-123):")
 	})
+
```
zimeg:
```suggestion
```

internal/pkg/create/create.go
zimeg, line 235: 🔬 thought: So curious! I thought creating missing directories was supported but I am wrong!

internal/pkg/create/create.go
```diff
@@ -67,16 +67,16 @@ func Create(ctx context.Context, clients *shared.ClientFactory, createArgs Creat
 		return "", slackerror.Wrap(err, slackerror.ErrAppDirectoryAccess)
 	}
 
-	// Get the app selection and accompanying app directory name (this may change when we find the unique directory name)
-	appDirName, err := getAppDirName(createArgs.AppName)
+	// Parse the app name input into a directory path and display name
+	appPath, displayName, err := parseAppPath(createArgs.AppName)
```
zimeg: 🐛 issue: I'm finding the app path can be changed with the "--name" flag which is unexpected as:

```
$ slack create path/to/my-app --name "APPPP"

📂 Created a new Slack project
   Cloning template slack-samples/bolt-js-starter-template
   To path ~/programming/tools/slack-cli/apppp

```

🔍 thought: I'd instead expect the argument path to be preferred here with the name being applied to the app itself? I'm unsure if separating these functions makes this more clear but think this case is good to test too!

zimeg (changes requested):
@srtaalej Thank you for bringing a fix to PR! I'm leaving a comment of another case to consider with this but I'm eager to get this merged 🚀

internal/pkg/create/create.go
srtaalej: yes very curious! i thought so too but i think it would be nice to have 🤔 perhaps in followups 🚀

internal/pkg/create/create.go
srtaalej: hmmm i thought the --name flag was always preferred? this was introduced with the name flag in https://github.com/slackapi/slack-cli/pull/327
but thanks for catching this because i would also expect `--name` to override the display name and leave the path alone

zimeg: @srtaalej Ahh we might be finding nuance of this command 😓 I recall these discussion too and didn't account for these edges...

internal/pkg/create/create.go
zimeg:
```suggestion
	// Normalize to a variation of kebab-case: replace non-alphanumeric with dashes, collapse, and trim
```

🌠 issue(non-blocking): The lowercasing might be confusing to CI that creates apps with a random name. We might consider keeping uppercase to match inputs that make a valid directory?

```
$ pwd
~/Documents
$ slack create MyApp
...
   To path ~/Documents/myapp
```

zimeg, line 185:
```suggestion
```

🪓 suggestion(non-blocking): With the above comment I share this.

cmd/project/create.go
```diff
@@ -116,10 +116,16 @@ func runCreateCommand(clients *shared.ClientFactory, cmd *cobra.Command, args []
...
+	// from the positional argument. When no positional arg is given (e.g.
+	// "slack create --name APPPP"), the name flag also becomes the directory
+	// path since there's nothing else to derive it from.
+	displayNameOverride := ""
 	if nameFlagProvided {
-		appNameArg = createAppNameFlag
+		displayNameOverride = createAppNameFlag
+		if appNameArg == "" {
+			appNameArg = createAppNameFlag
+		}
```
zimeg: 🪴 suggestion: We might want to move this alongside logic below with some shuffling:

https://github.com/slackapi/slack-cli/blob/7e4d486a101953b9e27f099789d867193ee625c9/cmd/project/create.go#L148-L166

cmd/project/create.go
```diff
@@ -164,10 +170,11 @@ func runCreateCommand(clients *shared.ClientFactory, cmd *cobra.Command, args []
 		subdir = template.GetSubdir()
 	}
 	createArgs := create.CreateArgs{
-		AppName:   appNameArg,
-		Template:  template,
-		GitBranch: createGitBranchFlag,
-		Subdir:    subdir,
+		AppName:     appNameArg,
```
zimeg: 🪬 thought: The meaning of these arguments has inverted for me and I now think about this variable as the app path that might be used for the app name defaults. The path separators of this PR must have inspired such idea!

🔮 suggestion(follow-up): We might find it useful to update this but I'd like to know if I'm thinking about this right?


```suggestion
		AppPath:     appPathArg,
```

cmd/project/create.go
```diff
@@ -116,10 +116,16 @@ func runCreateCommand(clients *shared.ClientFactory, cmd *cobra.Command, args []
 		}
 	}
 
-	// --name flag overrides any positional app name argument
-	// This allows users to name their app "agent" without triggering the AI Agent shortcut
+	// --name flag overrides the manifest display name but preserves any path
+	// from the positional argument. When no positional arg is given (e.g.
+	// "slack create --name APPPP"), the name flag also becomes the directory
+	// path since there's nothing else to derive it from.
+	displayNameOverride := ""
```
zimeg:
```suggestion
	displayName := ""
```

🔬 suggestion: I understand within `pkg` this overrides a parsed path but that term might be confusing from this calling code.

zimeg (approved):
🔮 @srtaalej This gets to a great place! I'm leaving a handful more comments that I hope untangle some arguments but find this is a nice fix-

- **Uppercase directories**: Earlier change lowercased path to directories too in normalization but downstream issues are noticed from this, albeit with workarounds. I note that it might be a nice change for a separate PR!
- **App path placement**: The "app name" argument feels outdated now that we're supporting "display name" too. A more meaningful change to perhaps make in this PR centralizes some of this logic but I think we can save other refactors?

I'm approving now but feel free to rerequest review if refactors feel awkward!

cmd/project/create.go
srtaalej: i think AppPath makes more sense

internal/pkg/create/create.go
srtaalej: since this is a pre-exisiting behavior i think maybe we leave this for a separate PR?

PR #532, opened 2026-05-08
fix: replace hashicorp/go-version with golang.org/x/mod/semver

mwbrooks:
### Changelog

- N/A

### Summary

This pull request fixes a Synk license policy issue by removing the direct dependency on `github.com/hashicorp/go-version` (MPL-2.0 license).

We now use `golang.org/x/mod/semver`, which is already a direct dependency. The `go-version` remains as an indirect dependency through the linter tooling, but is no longer flagged by Snyk.

### Preview

- N/A

### Testing

```bash
$ lack --version
# → Confirm the version is shown as "v4.0.1-XX-gXXXX" (e.g. v4.0.1-43-gbd5075f)

$ lack upgrade
# → Confirm the upgrade prompt is displayed for the dev build
# → Note: I think I may have a follow-up that disables showing the upgrade prompt on dev builds :thinking:
```

### Requirements

### DIFF
--- a/internal/update/semver.go
+++ b/internal/update/semver.go
@@ -25,10 +25,12 @@ func SemVerGreaterThan(release string, current string) (bool, error) {
 	r := ensureVPrefix(release)
 	c := ensureVPrefix(current)
 	if !semver.IsValid(r) {
-		return false, slackerror.New(slackerror.ErrInvalidSemVer)
+		return false, slackerror.New(slackerror.ErrInvalidSemVer).
+			WithMessage("Value %s is not a semantic version", release)
 	}
 	if !semver.IsValid(c) {
-		return false, slackerror.New(slackerror.ErrInvalidSemVer)
+		return false, slackerror.New(slackerror.ErrInvalidSemVer).
+			WithMessage("Value %s is not a semantic version", current)
 	}
 	return semver.Compare(r, c) > 0, nil
 }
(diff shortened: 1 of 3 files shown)


internal/update/semver.go
```diff
@@ -15,32 +15,40 @@
...
 
 // SemVerGreaterThan returns true if release is greater than current
 func SemVerGreaterThan(release string, current string) (bool, error) {
-	releaseVersion, err := version.NewVersion(release)
-	if err != nil {
-		return false, slackerror.New(slackerror.ErrInvalidSemVer).WithRootCause(err)
+	r := ensureVPrefix(release)
+	c := ensureVPrefix(current)
+	if !semver.IsValid(r) {
+		return false, slackerror.New(slackerror.ErrInvalidSemVer)
```
zimeg: ⚠️ thought: This might be surfaced in verbose logs too but I'm concerned that not wrapping this error can cause a less meaningful message for debugging?

internal/update/semver.go
```diff
@@ -15,32 +15,40 @@
...
 	}
-	currentVersion, err := version.NewVersion(current)
-	if err != nil {
-		return false, slackerror.New(slackerror.ErrInvalidSemVer).WithRootCause(err)
+	if !semver.IsValid(c) {
+		return false, slackerror.New(slackerror.ErrInvalidSemVer)
 	}
-	return releaseVersion.GreaterThan(currentVersion), nil
+	return semver.Compare(r, c) > 0, nil
```
zimeg, line 35: 🚀 praise: RC? Release candidate?

zimeg (approved):
@mwbrooks LGTM and will be fun to have for next release-

internal/update/semver.go
mwbrooks: 🧠 I could update this with a verbose log.

I had the same concern about not wrapping the error, but there is no error returned anymore. We could manually create a descriptive error, I suppose. But the `ErrInvalidSemVer` already captures the core essence.

zimeg: @mwbrooks Not blocking at all! I meant to suggest a "message" change with this but was also stumped about what'd be meaningful to include:

```suggestion
		return false, slackerror.New(slackerror.ErrInvalidSemVer).
			WithMessage("Value %s is not a semantic versioning", r)
```

mwbrooks: I like it! Commit f152325 adds the `.WithMessage` to each clause.

mwbrooks:
Thanks for the review @zimeg! 🎉 It'll be nice to have one less dependency and to get Snyk off our back about a license policy.

PR #538, opened 2026-05-10
feat: add `slack api` command

mwbrooks:
### Changelog

> Added `slack api <method> [key=value ...] [flags]` command for calling any Slack API method directly from the Slack CLI.
>
> Examples:
> - `slack api chat.postMessage channel=C123 text="hi"`
> - `slack api chat.postMessage '{"channel":"C123","text":"hi"}'`

### Summary

Adds a new top-level `slack api` command that lets users call any Slack API method without leaving the terminal.

**Key features:**

- **Token resolution**
  1. `--token` flag
  2. `--app` flag
  3. `SLACK_BOT_TOKEN` env var
  4. `SLACK_USER_TOKEN` env var
  5. App Prompt (only in project)
- **Body Format Auto-detection**:
  - Form-Encoded: `slack api chat.postMessage channel=C123 text="hi"`
  - JSON: `slack api chat.postMessage '{"channel":"C123","text":"hi"}'`
- **Body Format Flags: `--json <string>` and `--data <string>`**
  - Form-Encoded: `slack api chat.postMessage --data 'channel=C123 text="hi"'`
  - JSON: `slack api chat.postMessage --json '{"channel":"C123","text":"hi"}'`
- **HTTP Options Flags**
  - Method override (`-X`)
  - Custom headers (`-H`)
  - Show Response headers (`--include`)
- **Output formatting**
  - Pretty-printed JSON in TTY
  - Compact for piped output / scripting
- **Retry support**
  - Automatic retry on 429/503 with Retry-After

### Open Questions

- I originally had support for a team prompt, but removed it because it uses our Slack CLI tooling token, which has limited scope access. Sounds good?

### Preview

```
$ slack api auth.test --token xoxb-...
{
    "ok": true,
    "url": "https://myworkspace.slack.com/",
    "team": "My Workspace",
    "user": "bot",
    "team_id": "T12345",
    "user_id": "U12345"
}
```

#### Post Message as Form Data (with app selection prompt):

https://github.com/user-attachments/assets/64f297ce-2826-4289-bec3-e274f5d47ce7

#### Post Message as JSON (with app selection prompt):

https://github.com/user-attachments/assets/ccf2a770-7bb0-4f19-993d-751d67e9822d

#### Post Message with Custom Token `--token`:

https://github.com/user-attachments/assets/c185c397-66a8-48d7-91a7-c6e21ba5c81c

#### JSON Responses are Pretty-Printed in TTY:

https://github.com/user-attachments/assets/e35a7d16-0d3a-4841-b5e7-50541a344b97

#### JSON Responses are Condense in Scripting/Non-TTY:

https://github.com/user-attachments/assets/57d31dea-cea4-4138-a1cf-c60154a556b3

#### `--include` All Response Headers:

https://github.com/user-attachments/assets/ea1f39d0-7c16-41d0-9cd6-24724fd426ca

### Testing

- `slack api --help` - shows correct token priority and examples
- `slack api auth.test --token xoxb-...` - returns auth info
- `slack api auth.test --app local` - installs app and uses bot token (in project)
- `SLACK_BOT_TOKEN=xoxb-... slack api auth.test` - uses env var (outside project)
- `slack api auth.test` - prompts interactively when no flags/env set (in project)
- `slack api chat.postMessage channel=C123 text="hi"` - sends form-encoded body
- `slack api chat.postMessage '{"channel":"C123","text":"hi"}'` - auto-detects JSON
- `slack api auth.test --include` - shows HTTP status and response headers
- `slack api auth.test --token xoxb... | cat` - outputs compact JSON (non-TTY)
- `go test ./cmd/api/... ./internal/api/...` - all unit tests pass

### Requirements


srtaalej (commented):
This is looking great! ⭐ thanks for adding the test commands in the description 🫂 
the command is working great except
> `lack api auth.test | cat`

the test implies there should be JSON output but `slack api auth.test` is an interactive prompt so it errors out instead. perhaps we update the pr description? the command should be 
`lack api auth.test --token xoxb-... | cat` 
`lack api auth.test --app A... | cat` 
right?

internal/api/raw_request_test.go
```diff
@@ -0,0 +1,139 @@
...
+	assert.Equal(t, "other", receivedHeaders.Get("X-Another"))
+}
+
+func Test_RawRequest_RetryOnTooManyRequests(t *testing.T) {
+	attempts := 0
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		attempts++
+		if attempts == 1 {
+			w.Header().Set("Retry-After", "1")
+			w.WriteHeader(http.StatusTooManyRequests)
```
srtaalej, line 122: qq: does this cover retries on both 429 and 503 returns 🤔

srtaalej (approved):
left 2 testing-related comments/ questions but this is working great ! ⭐ ⭐ ⭐ such a handy command

cmd/api/api.go
```diff
@@ -0,0 +1,280 @@
...
+			return runAPICommand(cmd, clients, args)
+		},
+	}
+
+	cmd.Flags().StringVarP(&flags.method, "method", "X", "POST", "HTTP method for the request")
+	cmd.Flags().StringVar(&flags.json, "json", "", "JSON request body (uses Bearer token in Authorization header)")
+	cmd.Flags().StringVar(&flags.data, "data", "", "form-encoded request body string (e.g. \"key1=val1&key2=val2\")")
+	cmd.Flags().StringSliceVarP(&flags.headers, "header", "H", nil, "additional HTTP headers (format: \"Key: Value\")")
+	cmd.Flags().BoolVarP(&flags.include, "include", "i", false, "include HTTP status code and response headers in output")
+	cmd.MarkFlagsMutuallyExclusive("json", "data")
```
zimeg, line 111: 🏁 praise: Amazing find! I didn't know about this option!

cmd/api/api.go
```diff
@@ -0,0 +1,280 @@
...
+
+	if sdkConfigExists, _ := clients.SDKConfig.Exists(); sdkConfigExists {
+		selected, err := prompts.AppSelectPrompt(ctx, clients, prompts.ShowAllEnvironments, prompts.ShowInstalledAppsOnly)
+		if err == nil && selected.App.AppID != "" {
+			token, err := installAndGetBotToken(ctx, clients, selected)
+			if err == nil && token != "" {
+				return token, nil
+			}
+		}
+	}
```
zimeg: 🦠 issue: We might want to check environment variables before prompting? I find this can be blocking from a project:

```
$ SLACK_BOT_TOKEN="xoxb" slack api chat.postMessage
┃ Select an app
┃ ❱ A0B3AEPJUAH devrelsandbox T038J6TH5PF

{
    "ok": false,
    "error": "invalid_auth"
}
```

cmd/api/api.go
```diff
@@ -0,0 +1,280 @@
...
+
+	if sdkConfigExists, _ := clients.SDKConfig.Exists(); sdkConfigExists {
+		selected, err := prompts.AppSelectPrompt(ctx, clients, prompts.ShowAllEnvironments, prompts.ShowInstalledAppsOnly)
+		if err == nil && selected.App.AppID != "" {
+			token, err := installAndGetBotToken(ctx, clients, selected)
+			if err == nil && token != "" {
+				return token, nil
+			}
+		}
```
zimeg, line 245:
```suggestion
		}
		if err != nil {
			return "", err
		}
```

🗣️ issue: To catch interrupts if the suggestion adjacent seems right?

cmd/api/api.go
```diff
@@ -0,0 +1,280 @@
...
+		customHeaders[strings.TrimSpace(key)] = strings.TrimSpace(value)
+	}
+
+	resp, err := apiClient.RawRequest(ctx, flags.method, method, token, bodyReader, contentType, customHeaders)
+	if err != nil {
+		return err
+	}
+
+	if flags.include {
+		fmt.Fprintf(cmd.OutOrStdout(), "HTTP %d\n", resp.StatusCode)
```
zimeg, line 195: 🎨 quibble: I forgot how newlines are formatted but we might prefer outputs through `IO` with:

```
clients.IO.PrintInfo
```

Although I forget if that appends to debug logs or not...

cmd/api/api.go
zimeg: 👾 ramble: This might also guard against unknown `--app` flag?

cmd/api/api.go
```diff
@@ -0,0 +1,280 @@
...
+		for key, values := range resp.Header {
```
zimeg: 🧮 suggestion: Let's **sort** the request and response headers of this more verbose output! I'm finding these change each time and I think that's a `go` internal...

cmd/api/api_test.go
zimeg, line 1: 🧪 quibble: I understand these cases cover different expectations but I'm curious if we can use table tests to share setup between most?

zimeg (approved):
@mwbrooks Top tier PR! Incredible! 🏆 ✨

I'm eager to approve this with good findings in tests. But a few comments follow around token selection and outputs. I understand we have planned follow up but also want to suggest a workaround without tokens for the `api.test` method if that's not needed?

So excited to find these changes! 🎁

internal/api/raw_request_test.go
mwbrooks: Good catch! This test only covered 429. Commit 12630c8 adds a new test to explicitly cover the 503 case as well. 🙇🏻

cmd/api/api.go
mwbrooks: Yea, I think we could use this in a few other commands to clean up some resolution logic. 👌🏻

cmd/api/api.go
mwbrooks: Fantastic catch @zimeg! 🙇🏻 Commit 6891a56 fixes the resolution order and adds a test to check the order.

```bash
# Inside a project
$ SLACK_BOT_TOKEN="xoxb..." slack api auth.test
```

cmd/api/api.go
mwbrooks: Very good find @zimeg! I'm glad we caught this before shipping it 🙇🏻 Commit ab3a387 adds a variation of your suggestion and test coverage.

cmd/api/api.go
mwbrooks: Thanks for thinking of this. I looked into it and `clients.IO.PrintInfo` unfortunately has a lot of overhead - it adds a newline which makes non-TTY formatting a little less parsable and it adds tracing overhead but we don't want these responses in our logs.

I think we'll keep it as-is. But ultimately, we should really offer a cleaner, simpler `client.IO` print that doesn't do so much.

cmd/api/api.go
mwbrooks: praise: Great catch @zimeg! 🙇🏻 The headers are now sorted and we have a test 🔠 🧪 Commit ad581a7.

cmd/api/api_test.go
mwbrooks: Very fair point and thanks for noticing. I've consolidated the body format tests in commit e69776c. However, I don't think we want to use table tests for the other areas - there isn't enough overlap and it becomes quite messy to read. This is a big improvement though! 🙇🏻

mwbrooks:
Thanks for the reviews @srtaalej and @zimeg! Solid feedback from both of you 🙇🏻 I've followed up on everything and will merge the PR now. 🚀

zimeg:
OIP-2109305485

PR #543, opened 2026-05-13
feat: add two-way manifest sync between project and app settings

mwbrooks:
### Changelog

Added `slack manifest sync` command (and `slack sync` alias) that detects per-field differences between the local manifest.json and remote app settings, prompts users to resolve conflicts, and writes the merged result to both sides. Gated behind `--experiment=manifest-sync`.

### Summary

This pull request adds two-way manifest sync so developers can resolve differences between their local manifest.json and the remote app settings on api.slack.com.

Today, when `manifest.source=local`, the local manifest silently overwrites the API on next install — discarding changes made on the web. This feature replaces the binary "Overwrite?" prompt with a per-field resolution flow when divergence is detected.

Key additions:
- `internal/pkg/manifest/` — flatten, diff, merge, display, writeback, and sync orchestrator
- `cmd/manifest/sync.go` — standalone `slack manifest sync` command
- `slack sync` alias in `cmd/root.go`
- `manifest-sync` experiment flag for gated rollout
- Integration into `shouldUpdateManifest()` in the install flow

### Preview

```
📚 App Manifest
   Found 2 difference(s) between project and app settings

  display_information.description
    Project:      "A cool app for teams"
    App settings: "A cool app for collaboration"

  functions.new_func (only in project)
    Value: {"title":"New Function","description":"Greets"}

? How would you like to resolve these differences?
  > Use all project values
    Use all app settings values
    Choose for each difference

  Syncing manifest...
  ✓ Updated app settings
  ✓ Updated manifest.json

📚 App Manifest
   Finished manifest sync for "My App"
```

### Testing

1. Build the CLI: `make build`
2. In a project with `manifest.json`, install an app: `./bin/slack install -e manifest-sync`
3. **No differences**: Run `./bin/slack manifest sync -e manifest-sync` — should report "in sync"
4. **Remote change**: Edit the manifest on https://app.slack.com/app-settings, then run `./bin/slack manifest sync -e manifest-sync` — should show the difference, choose "Use all app settings values", verify manifest.json updated
5. **Local change**: Edit manifest.json, run `./bin/slack manifest sync -e manifest-sync` — choose "Use all project values", verify app settings updated
6. **Both sides changed**: Edit manifest.json and app settings (different fields), run sync, choose "Choose for each difference" — verify per-field resolution works and both sides reflect the merged result
7. **Key order preserved**: After sync, `git diff manifest.json` should only show value changes, not key reordering
8. **Triggered during install**: Make a remote change, run `./bin/slack install -e manifest-sync` — sync flow triggers instead of old "Overwrite?" prompt
9. **Force flag**: Make a remote change, run `./bin/slack install -e manifest-sync --force` — old behavior, local overwrites remote
10. **Without experiment**: Run `./bin/slack install` with a remote change — old "Overwrite?" prompt (unchanged behavior)

### Notes

- The `set-manifest` hook for code-generated manifests (e.g., manifest.ts) is not included in this initial implementation. For those projects, the merged manifest is pushed to the API only, with a warning that the local project was not updated.
- Uses a two-way (local vs remote) comparison model rather than three-way merge. This avoids stale cached base issues in multi-developer scenarios.
- The `--force-remote` flag (remote overwrites local) is a follow-up item.

### Requirements


srtaalej:
⌛**Update**: Added test coverage for `sync.go` and `display.go` (commit [0b98837](https://github.com/slackapi/slack-cli/pull/543/commits/0b98837d07b6acb9e653ef1fed018193aa3ee01f))
Patch coverage: 46.7% → 81.76%

`internal/manifest/display_test.go` (17 cases) — covers `DisplayDiffs` output rendering (all diff types, sort order, empty case), `PromptResolutionStrategy` (all 3 strategy mappings + error), `PromptFieldResolutions` (correct resolution for modified/local-only/remote-only fields in both directions + error), and `formatValue` (nil, string, map, truncation).

`internal/manifest/sync_test.go` (15 cases) — covers every error path through the `Sync` orchestrator (remote source rejection, fetch failures, API failure, cache failures, non-TTY without `--force`) and all happy paths (no differences, force flag, missing `manifest.json`, TTY with all-local/all-remote/per-field strategies, prompt cancellation).

srtaalej (approved):
⭐

internal/slackerror/errors.go
```diff
@@ -105,6 +105,7 @@ const (
 	ErrDotEnvPlaceholderNotFound                     = "dotenv_placeholder_not_found"
 	ErrDotEnvVarMarshal                              = "dotenv_var_marshal_error"
 	ErrEnterpriseNotFound                            = "enterprise_not_found"
+	ErrExperimentRequired                            = "experiment_required"
```
zimeg, line 108: 👁️‍🗨️ suggestion(non-blocking): A similar error is duplicated but unused. Without preference to either I do think we can find common patterns toward these:

https://github.com/slackapi/slack-cli/blob/d329dd95902caab4290523f09bb22d7a6728cf38/internal/slackerror/errors.go#L1198-L1201

internal/manifest/writeback.go
```diff
@@ -0,0 +1,192 @@
...
+	tmp := dest + ".tmp"
+	if err := afero.WriteFile(fs, tmp, data, mode); err != nil {
+		return err
+	}
+	if err := fs.Rename(tmp, dest); err != nil {
+		_ = fs.Remove(tmp)
+		return err
+	}
+	return nil
+}
```
zimeg, line 86: 🚂 thought: This was discussed in an adjacent PR but IMHO we should put these workarounds behind shared utilities after #616

internal/manifest/writeback.go
```diff
@@ -0,0 +1,192 @@
...
+
+func marshalFresh(manifest types.AppManifest) ([]byte, error) {
+	data, err := json.MarshalIndent(manifest, "", "  ")
+	if err != nil {
+		return nil, err
+	}
+	return append(data, '\n'), nil
+}
```
zimeg, line 192: 🔭 note: I think these should be moved into a shared package of either:

- internal/goutils/json.go
- internal/slackjson/json.go

internal/manifest/writeback.go
```diff
@@ -0,0 +1,192 @@
...
+
+// WriteManifestLocal writes the merged manifest back to the project's
+// manifest.json file, preserving the original file's key ordering by
+// using the same JSON structure.
+func WriteManifestLocal(fs afero.Fs, workingDir string, manifest types.AppManifest) (WriteBackResult, error) {
+	manifestPath := filepath.Join(workingDir, manifestFileName)
+
+	exists, err := afero.Exists(fs, manifestPath)
+	if err != nil {
+		return WriteBackResult{}, fmt.Errorf("failed to check manifest file: %w", err)
```
zimeg, line 45: ⚠️ note: We ought use wrapped error codes instead of formatted strings to trace debugs. Can be a fast follow with hope of improving this overall!

internal/manifest/writeback.go
```diff
@@ -0,0 +1,192 @@
...
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/slackapi/slack-cli/internal/shared/types"
+	"github.com/spf13/afero"
+)
+
+const manifestFileName = "manifest.json"
```
zimeg, line 28: 🌟 note: Let's document this as a standard ongoing! I think a CLI managed manifest makes these handoffs more clear and fits a personal model I have of app development better.

internal/manifest/writeback.go
```diff
@@ -0,0 +1,192 @@
...
+
+	exists, err := afero.Exists(fs, manifestPath)
+	if err != nil {
+		return WriteBackResult{}, fmt.Errorf("failed to check manifest file: %w", err)
+	}
+	if !exists {
+		return WriteBackResult{
+			Warning: fmt.Sprintf("No %s found in project root — merged manifest was not written locally", manifestFileName),
+		}, nil
+	}
```
zimeg, line 51: 🔮 thought: Longterm we might want to deprecate the `get-manifest` hook and check for the hook existence here instead. I can imagine oddities where a base "manifest.json" has extra values added after `sync` because of magic in the hook adding values at call time. No blocker for now but before this becomes stable I'd like us to have a smooth path for this!

internal/manifest/writeback.go
```diff
@@ -0,0 +1,192 @@
...
+type WriteBackResult struct {
+	Written  bool
+	FilePath string
+	Warning  string
+}
+
+// WriteManifestLocal writes the merged manifest back to the project's
+// manifest.json file, preserving the original file's key ordering by
+// using the same JSON structure.
+func WriteManifestLocal(fs afero.Fs, workingDir string, manifest types.AppManifest) (WriteBackResult, error) {
```
zimeg, line 40: 🪬 thought: I have hope that this function can pair well with adjacent manifest operations but for now no blocker in "internal" packages!

```diff
  clients.AppClient().Manifest.GetManifestLocal
+ clients.AppClient().Manifest.SetManifestLocal
```

internal/manifest/merge.go
```diff
@@ -0,0 +1,204 @@
...
+				delete(merged, path)
+			}
+		case ResolveRemote:
+			if val, exists := remoteFlat[path]; exists {
+				merged[path] = val
+			} else {
+				delete(merged, path)
+			}
+		}
+	}
```
zimeg, line 87: 🧠 praise: So interesting! I'm a fan of these pattern.

internal/manifest/sync.go
```diff
@@ -0,0 +1,154 @@
...
+	switch {
+	case clients.Config.ForceFlag:
+		merged, err = MergeAllFrom(localManifest.AppManifest, remoteManifest.AppManifest, diffs, MergeAllLocal)
+		if err != nil {
+			return nil, err
+		}
+	case !clients.IO.IsTTY():
+		return nil, slackerror.New(slackerror.ErrAppManifestUpdate).
+			WithRemediation("Run %s interactively to resolve manifest differences, or pass %s to push the project manifest to app settings",
+				style.CommandText("slack manifest sync"),
```
zimeg:
```suggestion
				style.Commandf("manifest sync", false),
```

🔬 quibble: To handle aliases as expected

internal/manifest/display.go
```diff
@@ -0,0 +1,169 @@
...
+		return MergeAllLocal, nil
+	case 1:
+		return MergeAllRemote, nil
+	default:
+		return MergePerField, nil
+	}
+}
+
+// PromptFieldResolutions asks the user to resolve each difference individually.
+func PromptFieldResolutions(ctx context.Context, io iostreams.IOStreamer, diffs *DiffResult) ([]FieldResolution, error) {
```
zimeg, line 89: 📣 thought: IMHO prompts and outputs are better to include with `cmd` to chain these together instead of calling into `internal` across various commands. This might be a significant refactor at this time but seems specific to the `sync` command.

internal/manifest/sync.go
```diff
@@ -0,0 +1,154 @@
...
+
+	diffs, err := Diff(localManifest.AppManifest, remoteManifest.AppManifest, app.IsDev)
+	if err != nil {
+		return nil, fmt.Errorf("failed to compute manifest differences: %w", err)
+	}
+
+	if !diffs.HasDifferences() {
+		clients.IO.PrintInfo(ctx, false, "\n%s", style.Sectionf(style.TextSection{
+			Emoji:     "books",
+			Text:      "App Manifest",
```
zimeg, line 70: 🦠 note(non-blocking): The "App Manifest" heading is consistent but I'd be curious to match command names across all of the manifest commands:

- Manifest Diff
- Manifest Info
- Manifest Sync
- Manifest Validate

https://github.com/slackapi/slack-cli/blob/b6ea8e7908c9fc936b35d10e99315392a323461b/.github/STYLE_GUIDE.md?plain=1#L65-L79

internal/manifest/display.go
```diff
@@ -0,0 +1,169 @@
...
+	"encoding/json"
+	"fmt"
+	"sort"
+
+	"github.com/slackapi/slack-cli/internal/iostreams"
+	"github.com/slackapi/slack-cli/internal/style"
+)
+
+// DisplayDiffs prints the differences to the terminal.
+func DisplayDiffs(ctx context.Context, io iostreams.IOStreamer, diffs *DiffResult) {
```
zimeg, line 28: 📣 suggestion: We should reuse the formatting of the `manifest diff` command here!

internal/manifest/sync.go
```diff
@@ -0,0 +1,154 @@
...
+	manifestSource, err := clients.Config.ProjectConfig.GetManifestSource(ctx)
+	if err != nil {
+		return nil, err
+	}
+	if manifestSource.Equals(config.ManifestSourceRemote) {
+		return nil, slackerror.New(slackerror.ErrAppManifestUpdate).
+			WithMessage("Manifest sync is unavailable for projects with remote source of truth").
+			WithRemediation("This project's manifest source is %s. Edit the manifest at %s.",
+				config.ManifestSourceRemote.Human(),
+				style.CommandText("slack app settings"),
```
zimeg:
```suggestion
				style.Commandf("app settings", false),
```

🎨 todo: For aliased commands

internal/manifest/sync.go
```diff
@@ -0,0 +1,154 @@
...
+	default:
+		merged, err = resolveInteractively(ctx, clients, localManifest.AppManifest, remoteManifest.AppManifest, diffs)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// Push merged manifest to API
+	clients.IO.PrintInfo(ctx, false, "\n  Syncing manifest...")
+	_, err = clients.API().UpdateApp(ctx, auth.Token, app.AppID, merged, true, true)
```
zimeg, line 100: 🪬 thought: Related to other manifest comments it'd be nice to have a solid interface for manifest operations being written too:

```diff
  clients.AppClient().Manifest.GetManifestRemote
+ clients.AppClient().Manifest.SetManifestRemote
```

zimeg (approved):
@mwbrooks Love to see these changes 😍 Albeit quite a bit here... 😅 

I'm alright merging this behind the `manifest-sync` experiment and will add this to the docs with this PR. I left a few other comments that we should revisit in ongoing iterations but nothing's blocking AFAICT.

- **Standardizing CLI-managed manifest:** We've had various discussion of extending how manifest values are gathered and set and this PR has a huge move on this! I think we should formalize the "manifest.json" file and deprecate `get-manifest` hooks as a default. This would give us more confidence writing a manifest for complex cases I think.
- **Separating command logic from internal:** A lot of prompting and output happens in internal packages at the moment but we should keep this with each command I think. This might help with duplicated formatting code?
- **Wrapped error codes**: Related to separated packagings we should wrap these errors that cross package boundaries with a formal `slackerror` value.

Other quibbles are around outputs that need changes more than this PR adds. Also I'm curious about a formalized internal manifest interface 🎭

Big thanks @srtaalej also for help with testing this all! 🧪 ✨

PR #548, opened 2026-05-14
fix: disable upgrade notification for the manifest info command

mwbrooks:
### Changelog

> Disable upgrade notification for the `slack manifest` and `slack manifest info` command. The command returns JSON that may be used for scripting and the upgrade notifications interfere with parsing the response output.

### Summary

This pull request adds `manifest info` to the list of commands that suppress update notifications. The `manifest info` command outputs structured data consumed by other tools, so injecting an upgrade banner would corrupt the output.

Also extends `isIgnoredCommand()` to support multi-word subcommands (e.g., `manifest info` is ignored but `manifest validate` is not).

### Preview

N/A

### Testing

1. Build with an artificially old version to trigger the upgrade notification:
   ```bash
   go build -ldflags="-X 'github.com/slackapi/slack-cli/internal/version.Version=v1.0.0'" -o bin/slack
   ```
2. Delete the `lastUpdateCheckedAt` timestamp from `~/.slack/config.json` so the check isn't skipped due to recency.
3. Create a project:
   ```bash
   ./bin/slack create my-app/
   cd my-app/
   ```
4. Verify `slack manifest info` does NOT show an upgrade notification:
   ```bash
   ./bin/slack manifest info --source local
   ```
6. Verify another command DOES show the upgrade notification.
   ```bash
   ./bin/slack manifest validate
   ```

### Requirements

### DIFF
--- a/internal/update/update.go
+++ b/internal/update/update.go
@@ -19,11 +19,11 @@ import (
 	"net/http"
 	"os"
 	"reflect"
+	"strings"
 	"sync"
 	"time"
 
 	"github.com/slackapi/slack-cli/internal/api"
-	"github.com/slackapi/slack-cli/internal/goutils"
 	"github.com/slackapi/slack-cli/internal/shared"
 	"github.com/slackapi/slack-cli/internal/slackerror"
 	"github.com/spf13/cobra"
@@ -223,13 +223,17 @@ func (u *UpdateNotification) isCI() bool {
 
 // isIgnoredCommand returns true when the process is in the list of commands.
 func (u *UpdateNotification) isIgnoredCommand() bool {
-	ignoredCommands := []string{"_fingerprint", "api", "version"}
-	osStr := os.Args[0:]
-	if len(osStr) < 2 {
+	ignoredCommands := []string{"_fingerprint", "api", "manifest info", "version"}
+	if len(os.Args) < 2 {
 		return false
 	}
-	commandName := osStr[1]
-	return goutils.Contains(ignoredCommands, commandName, true)
+	commandStr := strings.Join(os.Args[1:], " ")
+	for _, ignored := range ignoredCommands {
+		if commandStr == ignored || strings.HasPrefix(commandStr, ignored+" ") {
+			return true
+		}
+	}
+	return false
 }
 
 // isLastUpdateCheckedAtGreaterThan returns true when the time since the last update check is greater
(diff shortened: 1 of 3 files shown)


internal/update/update.go
```diff
@@ -223,13 +223,17 @@ func (u *UpdateNotification) isCI() bool {
 
 // isIgnoredCommand returns true when the process is in the list of commands.
 func (u *UpdateNotification) isIgnoredCommand() bool {
-	ignoredCommands := []string{"_fingerprint", "api", "version"}
-	osStr := os.Args[0:]
-	if len(osStr) < 2 {
+	ignoredCommands := []string{"_fingerprint", "api", "manifest info", "version"}
```
zimeg:
```suggestion
	ignoredCommands := []string{"_fingerprint", "api", "manifest", "manifest info", "version"}
```

🪬 question: Is the shortened command automatic? I prefer extended commands often in scripts but am unsure if this should be covered too.

internal/update/update_test.go
```diff
@@ -165,14 +166,22 @@ func Test_UpdateNotification_isIgnoredCommand(t *testing.T) {
 			command:  "version",
 			expected: true,
 		},
+		"manifest info command": {
+			command:  "manifest info",
+			expected: true,
+		},
+		"manifest validate command": {
+			command:  "manifest validate",
+			expected: false,
```
zimeg, line 187: 👁️‍🗨️ praise: Thanks! This makes me think much about tests as spec more than unit.

zimeg (approved):
@mwbrooks LGTM! And thanks for saving confusion or error 🙏 

I'm leaving a question that's alright to ignore and a praise for keeping code health improving constant! Please let's get this merged when time is right.

internal/update/update.go
mwbrooks: brutal, I forgot we offer `slack manifest` as a short-form version of `slack manifest info`. We should probably deprecate that now that we have more manifest commands - normally running the top-level parent command would show the help info about the sub-commands.

answer: No, this matches against `os.Args` so it's the raw commands typed by the user. So aliases are not captured, either.

Let's add `manifest` here. It's safer because it captures `manifest info` but will also capture all other manifest commands. We'd require a re-write after Cobra parsing to capture aliases properly, but today the update goroutine is started before parallel to the Cobra parsing (not inside Cobra).

mwbrooks: @zimeg Commit 33d7e0ce8baf25e4d25cd43185b45eacbeefa477 adds your suggested, some tests, and comments. It also adds a `DEPRECATE(semver:major)` to the `slack manifest` command to no longer print the manifest info. You should be credited as a co-author ✏️ 😄

Issue #562, opened 2026-05-27
Huh confirm prompt: selected vs unselected button is hard to distinguish


srtaalej:
## Problem

When using a Yes/No confirm prompt (huh), the currently selected button is hard to distinguish from the unselected one. The `ThemeSurvey` theme doesn't override the default huh button styles, so it falls through to `ThemeBase` defaults:

- **FocusedButton** (selected): black text on white background (ANSI color 0 on 7)
- **BlurredButton** (not selected): white text on black background (ANSI color 7 on 0)

Depending on terminal theme and background color, these can look very similar or the contrast between them isn't obvious enough to tell which one you're currently on.

<img width="457" height="185" alt="Image" src="https://github.com/user-attachments/assets/4a4b2ff9-ff49-4597-b676-f83232c57514" />

## Expected

It should be immediately clear which button (Yes/No) is currently selected when navigating.

## Suggestion

Add explicit `FocusedButton` and `BlurredButton` styles to `ThemeSurvey` in `internal/style/theme.go` with better visual differentiation — e.g. using the existing blue for the focused button, or adding a chevron/indicator.

zimeg:
Fixed in #564 🏁

PR #564, opened 2026-05-27
feat: show confirm prompts as a vertical selection

srtaalej:
### Changelog

Y/N select prompts now display as vertical selection similar to 'slack create'. 

### Summary

This PR changes the Y/N select prompt from left/right selector to up/down selector. 

When using a Yes/No confirm prompt (huh), the currently selected button is hard to distinguish from the unselected one. The ThemeSurvey theme doesn't override the default huh button styles, so it falls through to ThemeBase defaults. 
Depending on terminal theme and background color, these can look very similar or the contrast between them isn't obvious enough to tell which one you're currently on.

### Preview

**Before** 
Screenshot 2026-05-27 at 12 07 11 PM

**After** 
Screenshot 2026-05-27 at 3 51 26 PM

### Testing

- [ ] Run slack app delete or any command with a confirm prompt and verify the Yes/No options appear as a vertical list with the chevron selector
- [ ] Confirm arrow keys (up/down) navigate between Yes and No
- [ ] Confirm Enter submits the highlighted option
- [ ] Verify accessible mode still works (--accessible flag shows numbered options)

### Notes

Adresses issue https://github.com/slackapi/slack-cli/issues/562

### Requirements


internal/iostreams/forms.go
zimeg, line 97: 🌵 question: Is setting the default value to "true" - or the top option - for all confirmation prompts too wild an idea?

🦔 ramble: I found it jarring when testing and think it might be most predictable to prompt from start.

https://github.com/slackapi/slack-cli/blob/85f5c961466fde4a42f7f836f3c748ec6403c149/.github/STYLE_GUIDE.md?plain=1#L16-L22

🏁 ramble: We ought keep flag defaults as is but also that might bring follow up discussion on flag pairings for various prompts. The force flag is overused and unclear IMHO. For the `delete` example I'd expected instead:

```diff
- --force
+ --confirm-delete
```

zimeg (commented):
@srtaalej Super appreciate this change! It makes selections feel more expected 🎨 ✨ 

I'm a fan of the approach and am leaving a comment before approving after testing this a bit 🔬 Let me know what you think! This is exciting! 🚢

internal/iostreams/forms.go
srtaalej: ahh yes this makes more sense! thank you for sharing that doc 👁️ ❤️

zimeg: @srtaalej Ah I meant to suggest keeping the ordering as was:

- Yes
- No

But removing the default value from forms. The latest commit 9a8cbee8cb9edf7086c13defa9916a3ac4b37595 reversed the ordered but now default values that are "true" will appear on bottom:

```
$ slack init
```

Image

zimeg (changes requested):
@srtaalej Was late writing but think we ought still remove default values that aren't starting at the top for forms - I comment an example more 👁️‍🗨️

internal/iostreams/forms.go
```diff
@@ -82,15 +82,19 @@ func inputForm(io *IOStreams, _ context.Context, message string, cfg InputPrompt
...
+			huh.NewOption("No", false),
+		).
 		Value(choice)
 	return newForm(io, field)
 }
 
 // confirmForm interactively prompts for a yes/no confirmation.
-func confirmForm(io *IOStreams, _ context.Context, message string, defaultValue bool) (bool, error) {
-	var choice = defaultValue
+func confirmForm(io *IOStreams, _ context.Context, message string, _ bool) (bool, error) {
```
srtaalej: > note: defualtValue is unused but I renamed it to _ because it's required by the IOStreams interface signature

zimeg: @srtaalej IMO we should update the call signature to avoid confusion when passing a value here?

```suggestion
func confirmForm(io *IOStreams, _ context.Context, message string) (bool, error) {
```

👾 ramble: I'm unsure how much change that might cause but I think now'd be the right time to keep updates together!

zimeg (changes requested):
@srtaalej Much appreciated to the changes of 199785bcde316e24ed7fcd1bd05ade430f0e8469 but I want to request changes to keep prompts and forms consistent across the interface here. Thanks for calling out this change earlier 👾 ✨

internal/iostreams/forms.go
srtaalej: i actually think that the caller can be updated in a different PR because it touches files outside the scope of this pr 🤔 what do you think

zimeg: 🔭 @srtaalej I'm understanding removing defaults from confirm prompt for vertical selections is a change to those prompts also?

Fair to follow up with removing this because it's internal refactor but current branch has code in odd state without removing these defaults:

https://github.com/slackapi/slack-cli/blob/4e52cbdbe5a1a7bfed8aca2e57a60bf48fb6963a/cmd/app/delete.go#L123

https://github.com/slackapi/slack-cli/blob/4e52cbdbe5a1a7bfed8aca2e57a60bf48fb6963a/cmd/app/link.go#L139

mwbrooks: suggestion: @srtaalej I don't think we can ignore the default value (`_ bool`). Many commands set the default value to `false` (No) and this is extremely important to avoid destructive behaviour. For example, deleting an app will default to no because we don't want a developer to accidentally, permanently deleting their App ID.

A quick search shows the following commands deliberately passed `false` to the confirmation prompt:

- `cmd/app/delete.go:123` - "Are you sure you want to delete the app?"
- `cmd/app/uninstall.go:117` - "Are you sure you want to uninstall?"
- `cmd/sandbox/delete.go:87` - "Are you sure you want to delete the sandbox?"
- `cmd/app/unlink.go:104` - "Are you sure you want to unlink this app?"
- `cmd/externalauth/remove.go:111+`- Remove all tokens
- `internal/api/app.go:799` - "Cancel the current request…?"

So we'll want a implementation something like:

```golang
func confirmForm(io *IOStreams, _ context.Context, message string, defaultValue bool) (bool, error) {
    var choice = defaultValue
    err := buildConfirmForm(io, message, &choice).Run()
    // ...
}
```

Visually, we want to keep the "Yes / No" order unchanged with "Yes" on top. When the default value is `false` the second option ("No") will automatically be selected.

```bash
┃ Do you want to do this?
┃   Yes
┃ ❱ No
```

mwbrooks (changes requested):
🙇🏻 Thanks for putting together this PR @srtaalej!

🧪 I've left some important feedback around default values. I think we need to address it before we can merge this PR.

- We must support default values for the confirmation prompt to avoid regression
- The visual order should never change, just the default selected choice

internal/iostreams/forms.go
srtaalej: thank you for this clarification @mwbrooks! my latest commit reverts the logic back to the initial commit 🫡

internal/iostreams/forms.go
```diff
@@ -82,8 +82,12 @@ func inputForm(io *IOStreams, _ context.Context, message string, cfg InputPrompt
 
 // buildConfirmForm constructs an interactive form for yes/no confirmation prompts.
 func buildConfirmForm(io *IOStreams, message string, choice *bool) *huh.Form {
-	field := huh.NewConfirm().
+	field := huh.NewSelect[bool]().
```
srtaalej, line 85: order is unchanged and the `defaultValue` the cursor lands on depends on the command 🫡 

`No`
Image

`Yes`
Image

srtaalej (commented):
thank you @zimeg and @mwbrooks for kind reviews!
this latest commit keeps the changes that were made in the initial commit. to clarify on the behavior:
the cursor lands on whatever defaultValue is passed by the command in use 🧚 no changes to the logic only the UI 🫡

zimeg (approved):
@srtaalej Thank you for working through feedback I shared - let's get this merged with default options to match current patterns. If this feels unexpected in actual use I'm open to revisiting this without fear of breaking form options 🚢 💨

PR #565, opened 2026-05-28
feat: add --app flag support to slack create for linking existing apps

srtaalej:
### Changelog

`slack create` now accepts the `--app` flag alongside `--template` to scaffold a project and automatically link it to an existing app via `app link`.

### Summary

This PR adds `--app [ID]` and `--environment` flag support to `slack create`. When used with `--template`, the CLI will:

1. Scaffold the project from the template (existing behavior)
2. Run the existing `app link` flow to save the app ID to the project

The `--environment` flag (local/deployed) is passed through to `app link` to determine which file the app is saved to. If omitted, the user is prompted.

Flag validation:
- `--app` without `--template` returns an error
- `--environment` without `--app` returns an error

### Example

```bash
slack create my-project -t slack-samples/bolt-js-starter-template --app A0123456789 --environment local
```

### Testing

```bash
make test testdir=cmd/project testname=TestCreateCommand_AppFlag
make test testdir=cmd/app testname=TestLinkCommand
```

Manual verification:
1. `./bin/slack create my-project -t slack-samples/bolt-js-starter-template --app <real-app-id> --environment local`
2. Confirmed `.slack/apps.dev.json` contains linked app with correct team/app IDs
3. Confirmed `--app` without `--template` returns helpful error
4. Confirmed `--environment` without `--app` returns helpful error
5. Confirmed normal `slack create` (no `--app`) works unchanged

### Notes

- Follow-up PR: fetch and write the remote manifest to the project after linking

### Requirements


cmd/project/create_app.go
```diff
@@ -0,0 +1,116 @@
...
+
+// fetchRemoteManifest retrieves the app manifest from the platform via apps.manifest.export.
+func fetchRemoteManifest(ctx context.Context, clients *shared.ClientFactory, token string, appID string) (types.SlackYaml, error) {
+	manifest, err := clients.AppClient().Manifest.GetManifestRemote(ctx, token, appID)
+	if err != nil {
+		return types.SlackYaml{}, slackerror.New(slackerror.ErrInvalidManifest).
+			WithMessage("Failed to fetch manifest for app %s", appID)
+	}
+	return manifest, nil
+}
```
zimeg:
```suggestion
```

🪓 suggestion: Let's inline this! I think the error returned might sometimes be different from invalid manifest that we might want to surface

cmd/project/create_app.go
```diff
@@ -0,0 +1,116 @@
...
+			WithMessage("Failed to serialize app manifest")
+	}
+
+	manifestPath := filepath.Join(projectPath, "manifest.json")
+	if err := afero.WriteFile(fs, manifestPath, append(manifestData, '\n'), 0644); err != nil {
+		return slackerror.Wrap(err, slackerror.ErrProjectFileUpdate).
+			WithMessage("Failed to write manifest to project")
+	}
+	return nil
+}
```
zimeg: 🛻 suggestion: Let's move this logic to `internal/app/manifest` for adjacent changes of #543

cmd/project/create.go
```diff
@@ -183,6 +210,30 @@ func runCreateCommand(clients *shared.ClientFactory, cmd *cobra.Command, args []
...
+		originalDir, _ := clients.Os.Getwd()
+		if err := os.Chdir(absProjectPath); err != nil {
+			return slackerror.Wrap(err, slackerror.ErrAppDirectoryAccess)
+		}
+		linkErr := linkAppToProject(ctx, clients, appAuth, clients.Config.AppFlag, remoteManifest)
+		_ = os.Chdir(originalDir)
+		if linkErr != nil {
+			return linkErr
+		}
+	}
```
zimeg, line 235: 🧊 suggestion: We might want to instead share existing logic of `internal` package here and add an `AppID` to the "CreateArgs" above? I'd keep the `cmd` portion limited to input and output perhaps-

https://github.com/slackapi/slack-cli/blob/9a8cbee8cb9edf7086c13defa9916a3ac4b37595/internal/pkg/create/create.go#L149-L161

cmd/project/create_app.go
```diff
@@ -0,0 +1,116 @@
...
+		EnterpriseID: auth.EnterpriseID,
+	}
+
+	if manifest.IsFunctionRuntimeSlackHosted() {
+		return clients.AppClient().SaveDeployed(ctx, app)
+	}
+	app.IsDev = true
+	app.UserID = auth.UserID
+	return clients.AppClient().SaveLocal(ctx, app)
+}
```
zimeg: 🔭 question: Can we reuse logic of the `app link` command? We perhaps might change outputs but I'm hoping we move toward focused and atomic commands that perhaps compose!

cmd/project/create.go
```diff
@@ -127,6 +130,30 @@ func runCreateCommand(clients *shared.ClientFactory, cmd *cobra.Command, args []
...
 	}
 
+	// --app requires --template (Mode 2 deferred)
+	appFlagProvided := clients.Config.AppFlag != "" && types.IsAppID(clients.Config.AppFlag)
+	if appFlagProvided && !templateFlagProvided {
+		return slackerror.New(slackerror.ErrMismatchedFlags).
+			WithMessage("The --app flag requires the --template flag when used with create")
+	}
+
+	// Fail fast: resolve auth and fetch manifest before creating the project
```
zimeg: 🪬 thought: Related to comments of logic moved to `internal` I'm curious if we can move this check too? I understand a mismatched app ID will cause error but I don't think we should prompt for name when the `--app` flag is used...

🐮 ramble: For example a minimal example seems excessive for CI use case:

```
$ slack create --app A0582JYKGB1 --template zimeg/slacks --branch snaek --name snaek --force
```

cmd/project/create_app.go
zimeg: 👾 issue: I'm concerned of the forced default local app here. The same CI example I share earlier is for a production app and I don't have immediate option to "deploy" the right app after using this:

```
$ slack create --app A0582JYKGB1 --template zimeg/slacks --branch snaek --name snaek --force
```

🌠 suggestion: We might want to use the `--environment` flag to decide this? I still think a default "local" makes sense - CI should be explicit!

zimeg (commented):
@srtaalej Awesome changes going on here! 🎁 

I'm leaving a handful of comments around refactoring logic into adjacent places. Hoping that we can compose commands overall and avoid adding too much to `create` for ongoing iteration.

Two notable changes to the experience that I'll call out here include:
- Accepting the environment flag alongside a default
- Skipping the name prompt when an existing app ID is provided

Quite excited for what this hopes to unlock 🔏

cmd/project/create.go
srtaalej: you're right, this should def go in internal!

cmd/project/create.go
srtaalej: true! 🫡

cmd/project/create.go
```diff
@@ -127,6 +133,36 @@ func runCreateCommand(clients *shared.ClientFactory, cmd *cobra.Command, args []
 			WithMessage("The --subdir flag requires the --template flag")
 	}
 
+	// --app requires --template (Mode 2 deferred)
```
zimeg: 👁️‍🗨️question: Which mode 2 in this case?

cmd/project/create.go
zimeg: 🐌 question: Was this still something to move into `internal`? I'm hoping we might preferred the deferred `chdir` existing!

internal/pkg/apps/link.go
```diff
@@ -0,0 +1,157 @@
...
+
+// FetchRemoteManifest retrieves the app manifest from the platform via apps.manifest.export.
+func FetchRemoteManifest(ctx context.Context, clients *shared.ClientFactory, token string, appID string) (types.SlackYaml, error) {
+	manifest, err := clients.AppClient().Manifest.GetManifestRemote(ctx, token, appID)
+	if err != nil {
+		return types.SlackYaml{}, slackerror.Wrap(err, slackerror.ErrInvalidManifest).
+			WithMessage("Failed to fetch manifest for app %s", appID)
+	}
+	return manifest, nil
+}
```
zimeg: 🪓 suggestion: Let's inline this! I think the error returned might sometimes be different from invalid manifest that we might want to surface
```suggestion
```

internal/pkg/apps/link.go
```diff
@@ -0,0 +1,157 @@
...
+		}
+	case false:
+		if clients.Config.ForceFlag || (deploy.IsNew() && local.AppID != app.AppID) {
+			return clients.AppClient().SaveDeployed(ctx, app)
+		}
+	}
+	return slackerror.New(slackerror.ErrAppFound).
+		WithMessage("A saved app was found and cannot be overwritten").
+		WithRemediation("Remove the app from this project or try again with %s", style.Bold("--force"))
+}
```
zimeg: 🦠 suggestion: This is duplicate to `app link` implementation we might not want to duplicate this?

https://github.com/slackapi/slack-cli/blob/dc30dc38282e3152715a9c2b14f34f177ced75e3/cmd/app/link.go#L283-L307

internal/pkg/apps/link.go
```diff
@@ -0,0 +1,157 @@
...
+			WithMessage("Failed to serialize app manifest")
+	}
+
+	manifestPath := filepath.Join(projectPath, "manifest.json")
+	if err := afero.WriteFile(fs, manifestPath, append(manifestData, '\n'), 0644); err != nil {
+		return slackerror.Wrap(err, slackerror.ErrProjectFileUpdate).
+			WithMessage("Failed to write manifest to project")
+	}
+	return nil
+}
```
zimeg: 🗳️ suggestion: We might want to make this `internal/app/manifest.go` I think we want to move away from `internal/pkg/.......` ongoing

internal/pkg/apps/link.go
```diff
@@ -0,0 +1,157 @@
...
+	for i := range allAuths {
+		if _, err := clients.API().GetAppStatus(ctx, allAuths[i].Token, []string{appID}, allAuths[i].TeamID); err == nil {
+			return allAuths[i], nil
+		}
+	}
+
+	return types.SlackAuth{}, slackerror.New(slackerror.ErrAppNotFound).
+		WithMessage("No authenticated workspace has access to app %s", appID).
+		WithRemediation("Run %s to sign in to the workspace that owns this app", style.Commandf("login", false))
+}
```
zimeg: 💾 thought: This might be something we use to replace the following:

https://github.com/slackapi/slack-cli/blob/aceb7a45ed171d8238cdcaabea41dccd3b532f64/cmd/app/link.go#L211-L240

zimeg: 👾 ramble: Am leaving comments in hopes that these additions can be reused more as:

- Run the `create` command
  - If "--app" flag then run the `link` command
  - No "--app" flag continues without change

cmd/project/create.go
zimeg: 🧂 ramble: Similar comment as below I wonder if this is logic can be passed as `createArgs` or perhaps afterwards to the `link` command?

zimeg (changes requested):
@srtaalej Leaving a few more comments around logic that we might want to separate between `project create` and `app link` command 🔭 

To me it's seeming more that these commands should be sequenced so the `--app` and `--environment` flags are passed through the "link" command after the "create" command clones the template. I'm unsure of right scope for this but notice a few improvements happening:

- Finding saved authentication for a provided `link` app 🎁 
- Copying the existing app manifest from upstream app settings ⚙️ 
- Saving the provided app ID from `create` flags 🏁 

Am requesting approval to hope we can reuse more command logic and think these enhancements might be alright to break into multiple changesets if the notes above seem right?

srtaalej:
@zimeg thanks for re-review! i agree these should be separate changesets 🤔 your comments definitely clarified the command flow!

cmd/project/create.go
```diff
@@ -37,6 +40,7 @@ var createGitBranchFlag string
 var createAppNameFlag string
 var createListFlag bool
 var createSubdirFlag string
+var createEnvironmentFlag string
```
zimeg: 🧮 quibble: For alphabetical order?

cmd/project/create.go
```diff
@@ -67,6 +71,7 @@ name your app 'agent' (not create an AI Agent), use the --name flag instead.`,
 			{Command: "create my-project -t slack-samples/deno-hello-world", Meaning: "Start a new project from a specific template"},
 			{Command: "create --name my-project", Meaning: "Create a project named 'my-project'"},
 			{Command: "create my-project -t org/monorepo --subdir apps/my-app", Meaning: "Create from a subdirectory of a template"},
+			{Command: "create my-project -t slack-samples/bolt-js-starter-template --app A0123456789", Meaning: "Create from template and link to an existing app"},
```
zimeg: 🪬 suggestion: Including the `--environment` flag might be meaningful with this example to avoid retries in CI setup or experiments avoiding prompts?

cmd/project/create.go
```diff
@@ -127,6 +133,19 @@ func runCreateCommand(clients *shared.ClientFactory, cmd *cobra.Command, args []
...
+	if appFlagProvided && !templateFlagProvided {
+		return slackerror.New(slackerror.ErrMismatchedFlags).
+			WithMessage("The --app flag requires the --template flag when used with create")
+	}
+
+	// --environment requires --app
+	if cmd.Flags().Changed("environment") && !appFlagProvided {
+		return slackerror.New(slackerror.ErrMismatchedFlags).
+			WithMessage("The --environment flag requires the --app flag when used with create")
+	}
```
zimeg: 🔬 thought: We might guard against unexpected `--environment` values with this check or perhaps reset all changes if `link` command errors? I find template remains in an incomplete state after this command:

```
$ slack create --template slack-samples/bolt-js-starter-template --app A0B93LJK24A --name asdf --environment deploy --team slackbox-ez
```

🦠 ramble: The issue here is "deploy" instead of "deployed" I believe!

cmd/project/create_test.go
```diff
@@ -853,3 +853,34 @@ func TestCreateCommand_confirmExternalTemplateSelection(t *testing.T) {
...
+				CreateFunc = createClientMock.Create
+			},
+			ExpectedErrorStrings: []string{"The --environment flag requires the --app flag when used with create"},
+			ExpectedAsserts: func(t *testing.T, ctx context.Context, cm *shared.ClientsMock) {
+				createClientMock.AssertNotCalled(t, "Create", mock.Anything, mock.Anything, mock.Anything)
+			},
+		},
+	}, func(cf *shared.ClientFactory) *cobra.Command {
+		return NewCreateCommand(cf)
+	})
```
zimeg, line 983: 🧪 suggestion: Adding cases to cover the expected happy paths might be meaningful to guarantee the `create` command keeps support for these paths as underlined commands change

zimeg (commented):
@srtaalej Thanks for breaking this into multiple changes 🪬 ✨ 

The comments I'm leaving are much more focused and I think we can land this in separated changes to build confidence throughout. Right now I comment on testing and handling edge cases more before building on this more 🔭 

Overall an exciting experience to start with an existing app! 🎁

cmd/project/create.go
mwbrooks: suggestion: Instead, how about we add a CI/Scripting example and allow the above example to stay focused on creating a project with an existing app ID.

srtaalej: agreed!

cmd/project/create.go
srtaalej: true! 🫡

cmd/project/create.go
```diff
@@ -127,6 +134,25 @@ func runCreateCommand(clients *shared.ClientFactory, cmd *cobra.Command, args []
 			WithMessage("The --subdir flag requires the --template flag")
 	}
 
+	// --app requires --template
+	appFlagProvided := clients.Config.AppFlag != "" && types.IsAppID(clients.Config.AppFlag)
```
mwbrooks: minor: `--app local` will silently skip linking because it's not an App ID. While it's correct that we don't want to support `--app local` and `--app deployed` when creating an app, we should display an error immediately when it's not an App ID.

cmd/project/create.go
```diff
@@ -183,6 +209,22 @@ func runCreateCommand(clients *shared.ClientFactory, cmd *cobra.Command, args []
...
+		absProjectPath, err := filepath.Abs(appDirPath)
+		if err != nil {
+			return slackerror.Wrap(err, slackerror.ErrAppDirectoryAccess)
+		}
+		originalDir, _ := clients.Os.Getwd()
+		if err := os.Chdir(absProjectPath); err != nil {
+			return slackerror.Wrap(err, slackerror.ErrAppDirectoryAccess)
+		}
+		linkErr := LinkFunc(ctx, clients, &types.App{}, false)
+		_ = os.Chdir(originalDir)
```
mwbrooks: minor: We should handle the error returned by `clients.Os.Getwd()` and handle the error / add a deferr to `os.Chdir(origianlDir).

cmd/project/create_test.go
```diff
@@ -853,3 +854,92 @@ func TestCreateCommand_confirmExternalTemplateSelection(t *testing.T) {
...
+				createClientMock = new(CreateClientMock)
+				createClientMock.On("Create", mock.Anything, mock.Anything, mock.Anything).Return(t.TempDir(), nil)
+				CreateFunc = createClientMock.Create
+
+				linkCalled := false
+				LinkFunc = func(ctx context.Context, clients *shared.ClientFactory, a *types.App, shouldConfirm bool) error {
+					linkCalled = true
+					assert.False(t, shouldConfirm)
+					return nil
+				}
```
mwbrooks: note: We should be backing up the original `LinkFunc` and resetting it on each test. Right now the new `func(...)` will carry-over to future tests.

cmd/project/create_test.go
mwbrooks: If I understand correctly, Eden would prefer that we don't mock the `LinkFunc` (old pattern that's becoming an anti-pattern for us) and instead have our tests call into the real link code to verify that the app is correctly linked.

mwbrooks (commented):
Looking great @srtaalej! This PR has came a long way and test coverage is way up!

I'm leaving some feedback. Since it's late on your side, I may push a few commits up to address my own comments. 🙏🏻

cmd/project/create.go
mwbrooks: Commit c9f5bbf now displays an error during the creation if a non-App-ID is provided. For example, `--app local` or `--app deployed`.

cmd/project/create.go
mwbrooks: Commit 7564232 now handles errors when getting or changing the working directory. It also defers restoring the original working directory, so that it's restored safely during errors.

cmd/project/create_test.go
mwbrooks: Commit 6533815 removes this code and now uses the real link function.

cmd/project/create_test.go
mwbrooks: Commit 6533815 removes the `LinkFunc` stub and mock. It now uses `app.LinkExistingApp(...)` in the implementation and tests.

mwbrooks (commented):
✅ Thanks for putting together this PR @srtaalej. I'm throwing an approval on because I think all of the feedback is handled.

💻 I've added 3 new commits based on the remaining feedback:   
1. c9f5bbf - fix: validate --app value as an app ID in create
2. 7564232 - fix: restore working directory safely after linking in create
3. 6533815 - refactor: remove LinkFunc indirection from create tests

🧪 Local testing looks good. However, I had to add a `--team <team_id>` flag in order to skip the team selection prompt. This is something that we'll want to improve in a future PR @srtaalej.

🧠 I think a future PR should also strive to clean up the output of the create experience when `--app` is provided. The linking output looks noisy and confusing with multiple App Link and App Manifest headers.

Image

📝 @srtaalej A few follow-up tasks that you should note are:

1. App Settings Creation should include the `--team T0123` and `--environment local` flags in order to skip the prompts.
2. Improve the output experience for `create` when `--app` is provided. 
3. Discuss updating the CLI to not require the `--team` flag unless there are multiple auth'd teams with access to the App ID. This would require discovering the Team ID that has access to the App ID.
4. Discuss updating the CLI to not require the `--environment local` flag for the create command. A default could be `local` and advanced developers could specify it if they want deployed.

srtaalej:
@mwbrooks @zimeg thank you thank you for reviewing! excited to see this come together in app settings!

PR #566, opened 2026-05-29
feat: add --no-auth flag to api command for unauthenticated requests

WilliamBergamin:
### Changelog

Allow the `api` command to send unauthenticated requests when no token is available or when `--no-auth` is specified.

### Summary

Some Slack API methods (e.g. `api.test`, `blocks.validate`) don't require authentication. Previously, the `api` command always resolved a token and failed with an error if none was available, even when calling methods that don't need one.

This PR makes two changes:

1. **Graceful fallback**: if no token is found through any resolution path (flags, env vars, app prompt), the request is sent without authentication instead of erroring.
2. **`--no-auth` flag**: skips token resolution entirely. Useful in scripts or CI where you want to avoid interactive prompts and explicitly signal that no auth is needed.

The app selection prompt (shown in project directories) now includes a "No app" option, allowing users to opt out of authentication interactively.

`--no-auth` cannot be combined with `--token` or `--app` flags.

### Preview

Screenshot 2026-06-01 at 4 01 26 PM

---

Screenshot 2026-06-01 at 4 00 50 PM

### Testing

```sh
# Explicit no-auth
slack api blocks.validate --no-auth blocks='[{"type":"section","text":{"type":"mrkdwn","text":"Hello"}}]'
slack api api.test --no-auth

# Graceful fallback (no token configured, no project)
slack api api.test
```

### Notes

None

### Requirements


cmd/api/api.go
zimeg:
```suggestion
	return "", nil
```

🔮 thought: We might want to accept no token as valid input without requiring to know which flag is needed. I realize multiple cases exist for gathering tokens and we could use `--help` outputs toward this. The API response I think is meaningful too:

```
{
    "ok": false,
    "error": "not_authed"
}
```

🌛 suggestion: Might adding a "no app" option to app selection perhaps [support](https://github.com/slackapi/slack-cli/blob/85f5c961466fde4a42f7f836f3c748ec6403c149/.github/STYLE_GUIDE.md?plain=1#L22) having a `--no-auth` flag?

```diff
  Select an app
    A0B78JUQG4U devrelsandbox T02A074M3U3
    A0B78KLUSFN tinyspek E06LPMFSSTU
+ ❱ No app
```

🐮 ramble: In scripts a missing `--token` flag might have similar results and I hope the earlier error and documentation has details toward remediation!

zimeg (commented):
@WilliamBergamin Super nice unlocks this is bringing 🔓 ✨ 

I'm leaving a comment on an adjacent approach we can consider? Not against having `--no-auth` also but think it could pair with a new app select prompt option "no app" within projects 🎲 so details on this flag or methods needing it isn't needed in scripts otherwise 👾

WilliamBergamin:
@zimeg thanks for this awesome feedback 💯 

I opted to add the "No app" option to the app selection prompt to allow users to submit a request without an app token 🚀 But I think we need to keep some sort of `--no-auth` flag to allow scripts and CI to bypass authentication if they need to 🤔

I did think of renaming it to `--no-app`, but this feels weird because the logic truly just bypasses the token resolution.

internal/prompts/app_select.go
```diff
@@ -91,6 +91,23 @@ var appTransferDisclaimer = style.TextSection{
 	},
 }
 
+// ErrNoAppSelected is returned when the user selects "No app" in the prompt
+var ErrNoAppSelected = fmt.Errorf("no app selected")
```
zimeg: 🚧 suggestion: Can we import this from [`errors.go`](https://github.com/slackapi/slack-cli/blob/main/internal/slackerror/errors.go) instead?

internal/prompts/app_select.go
```diff
@@ -91,6 +91,23 @@ var appTransferDisclaimer = style.TextSection{
...
+
+// AppSelectOption configures optional behavior of AppSelectPrompt
+type AppSelectOption func(*appSelectConfig)
+
+type appSelectConfig struct {
+	includeNoApp bool
+}
```
zimeg, line 99: 🌟 praise: Excited for what this pattern unlocks in customizations!

zimeg (approved):
@WilliamBergamin Clean implementation! This puts things in great direction for building configurations of a selection! Thanks for improving this alongside feedback suggestions 🧰 ✨

PR #569, opened 2026-06-02
chore: order alphabetical error codes and constants in reference

zimeg:
### Changelog

N/A — no change to behavior but docs of #553 might keep new order 👁️‍🗨️ 📚 

### Summary

This pull request sorts the `internal/slackerror/errors.go` file alphabetically and ensures `docgen` produces deterministic output.

- Constants block sorted by `ErrXxx` identifier name
- `ErrorCodeMap` entries sorted by map key
- `docgen` now explicitly sorts errors before rendering the template

### Preview

```
$ slack docgen docs/reference
```

### Testing

- [x] Confirm the removed `diff` of this PR has matching additions 🔬 

### Notes

The Go spec states ([ref](https://go.dev/ref/spec#For_range)):

> "The iteration order over maps is not specified and is not guaranteed to be the same from one iteration to the next."

Without explicit sorting, `docgen` output depends on undefined map iteration order, which can vary across Go versions and builds.

### Requirements


internal/slackerror/errors.go
```diff
@@ -360,27 +355,24 @@ var ErrorCodeMap = map[string]Error{
 			ErrorDetail{Message: "Deployment to Slack is currently supported for apps written with the Deno Slack SDK."},
 		},
 		Remediation: fmt.Sprintf(`Learn about building apps with the Deno Slack SDK:
-
 https://docs.slack.dev/tools/deno-slack-sdk
-
 If you are using a Bolt framework, add a deploy hook then run: %s
-
 Otherwise start your app for local development with: %s`,
```
zimeg, line 363: 🦠 todo: Will return spaces!

WilliamBergamin (approved):

cmd/docgen/docgen.go
```diff
@@ -118,7 +118,14 @@ func runDocGenCommandFunc(clients *shared.ClientFactory, cmd *cobra.Command, arg
...
 	}
-	err = tmpl.Execute(file, slackerror.ErrorCodeMap)
+	errors := make([]slackerror.Error, 0, len(slackerror.ErrorCodeMap))
+	for _, e := range slackerror.ErrorCodeMap {
+		errors = append(errors, e)
+	}
+	slices.SortFunc(errors, func(a, b slackerror.Error) int {
+		return strings.Compare(a.Code, b.Code)
+	})
+	err = tmpl.Execute(file, errors)
```
WilliamBergamin, line 128: If the errors are already sorted do we need this? Or are go Maps unordered?

zimeg: @WilliamBergamin I understand maps don't have a specified order so want to build confidence in these outputs 📠 But I do realize this doesn't change our actual pages...

💡 thought: Will sit on this a bit before merging because it's additional code to maintain-

zimeg:
@WilliamBergamin I slept on this a bit and will go ahead with merging now to make it more explicit what we expect in outputs! Thank you for reviewing 🎁 ✨

PR #575, opened 2026-06-04
ci: skip release candidate if generated reference match current release

zimeg:
### Changelog

> N/A - But with hopes our workflow runs appear with more green checks.

### Summary

This PR fixes the release workflow attempting to create an RC PR if generated changes match what's current on `main` branch. This happens during the release process between merge and tag or if no semver labels exist on PRs since a latest tag.

### Preview

Error: https://github.com/slackapi/slack-cli/actions/runs/26916634889/job/79407765870

Note: We might not test this until next RC candidate merges unless we avoid adding a label before merge perhaps? I do not know!

### Notes

- The `make rc` command still errors if a commit isn't created (no changes needed for docs) to show the RC failed and matches stables instead. I comment that our check for this is interesting...

### Requirements


.github/workflows/release.yml
```diff
@@ -84,6 +84,13 @@ jobs:
               ;;
           esac
 
+          # Skip if docs already reference this version
+          if grep -q "Using slack v${NEXT_VERSION}" docs/guides/installing-the-slack-cli-for-mac-and-linux.md; then
+            echo "Docs already at v$NEXT_VERSION, nothing to release"
+            echo "has_changes=false" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
```
zimeg, line 92: 🪬 thought: Curious of perhaps recommended place to search for the latest version after discussion with @lukegalbraithrussell and @srtaalej earlier! This current approach isn't certain IMHO.

👾 ramble: Overall I prefer checking for an expected version here to skip the "commit" and "PR" steps following if no changes are found without adding to those steps!

zimeg (commented):
🔮 A comment of the current approach - I don't think this is rushed to merge until next release and it shouldn't be blocking a release either!

mwbrooks (approved):
✅ Thanks for fixing the failure on merge ❌

zimeg:
@mwbrooks Thanks so much! I'll remove the `patch` label in hopes of testing this once more 🚀 ✨

PR #585, opened 2026-06-09
feat(create): fetch remote manifest after linking app during create

srtaalej:
### Changelog

`slack create --app` now fetches the app's manifest from remote settings and writes it to the project's `manifest.json`, replacing the template's generic manifest with the actual app configuration.

### Summary

Follow-up to [#565](https://github.com/slackapi/slack-cli/pull/565). After linking an existing app during `slack create --app --template`, the CLI now fetches the remote manifest from App Settings and overwrites the template's `manifest.json` with it.

This also updates `LinkExistingApp` to return (`*types.SlackAuth`, error) so callers can use the auth token for subsequent API calls (like manifest fetching).

Design decisions:
- Manifest source remains local in `.slack/config.json` — the remote content is written locally
- Only `manifest.json` is overwritten (not `manifest.ts/.js` which are code files in Deno projects)
- On fetch failure, a warning is printed and the command continues — the project and link still succeed

### Testing

```
make test testdir=cmd/project testname=TestCreateCommand_AppFlag_FetchesRemoteManifest
make test testdir=cmd/project testname=TestCreateCommand_AppFlag
make test testdir=cmd/project testname=Test_Project_InitCommand
make test testdir=cmd/app testname=TestLinkCommand
```
 Manual:
1. `make build`
2. `./bin/slack create my-test -t slack-samples/bolt-js-starter-template --app <real-app-id> --environment local`
3. Verify `my-test/manifest.json` contains the remote app's manifest
4. Test with an invalid app ID to confirm the warning is shown and the project is still created

### Notes 🔴 
- Open question: should we set manifest.source to remote in `.slack/config.json` after writing the fetched manifest? Currently left as local so the user "owns" the file going forward. 
> A future PR could add merging logic between template and remote manifests.
- `manifest.ts/.js` (Deno SDK) projects are not supported

⌛ not merging until #598 is merged because of conflicts

### Requirements

- [x] I've read and understood the [Contributing Guidelines](https://github.com/slackapi/slackcli/blob/main/.github/CONTRIBUTING.md) and have done my best effort to follow them.


cmd/project/create.go
```diff
@@ -289,6 +307,21 @@ func printCreateSuccess(ctx context.Context, clients *shared.ClientFactory, appP
...
+	}
+	data, err := json.MarshalIndent(slackYaml.AppManifest, "", "  ")
+	if err != nil {
+		return err
+	}
+	data = append(data, '\n')
+	manifestPath := filepath.Join(projectPath, "manifest.json")
+	return afero.WriteFile(clients.Fs, manifestPath, data, 0644)
+}
+
```
zimeg: 📠 suggestion: We might find a new file better for this? It might be shared in other commands or PRs #591!

```
internal/manifest/export.go
```

cmd/project/create.go
```diff
@@ -229,8 +231,24 @@ func runCreateCommand(clients *shared.ClientFactory, cmd *cobra.Command, args []
...
+		if auth != nil && linkedApp.AppID != "" {
+			fetchErr := fetchAndWriteRemoteManifest(ctx, clients, auth.Token, linkedApp.AppID, absProjectPath)
+			if fetchErr != nil {
+				clients.IO.PrintWarning(ctx, "%s", style.Sectionf(style.TextSection{
+					Text: "Could not fetch the remote app manifest",
+					Secondary: []string{
+						fetchErr.Error(),
+						"The template manifest was kept unchanged",
+					},
+				}))
```
zimeg: ⚠️ question: Could returning the error here bring more confidence for scripts? I'm unsure when this might fail but would hope that invalid app IDs error overall to start

cmd/app/link.go
```diff
@@ -170,26 +170,26 @@ func LinkExistingApp(ctx context.Context, clients *shared.ClientFactory, app *ty
...
 		clients.IO.PrintDebug(ctx, "Error saving app to file when linking existing app: %s", err)
-		return err
+		return nil, err
 	}
 
 	// Footer section
 	LinkAppFooterSection(ctx, clients, app)
 
-	return nil
+	return auth, nil
```
zimeg: 🔗 question: Instead of returning an `auth` to calling functions that's often ignored, would setting up the manifest as part of the `link` command if no apps are saved make sense?

👾 ramble: I'm wanting to avoid mixing concepts with the `create` command which should be from a project template I think while `link` is seeming more app specific.

cmd/project/create.go
```diff
@@ -289,6 +307,21 @@ func printCreateSuccess(ctx context.Context, clients *shared.ClientFactory, appP
 	clients.IO.PrintTrace(ctx, slacktrace.CreateSuccess)
 }
 
+// fetchAndWriteRemoteManifest fetches the app manifest from remote settings and writes it to the project.
+func fetchAndWriteRemoteManifest(ctx context.Context, clients *shared.ClientFactory, token, appID, projectPath string) error {
+	slackYaml, err := clients.AppClient().Manifest.GetManifestRemote(ctx, token, appID)
+	if err != nil {
+		return err
+	}
+	data, err := json.MarshalIndent(slackYaml.AppManifest, "", "  ")
```
zimeg: 🔬 note: I'm finding some values are escaped but this hasn't caused an error when testing the custom step template:

```diff
-           "type": "slack#/types/user_id",
+           "type": "slack#\/types\/user_id",
```

cmd/project/create.go
```diff
@@ -289,6 +307,21 @@ func printCreateSuccess(ctx context.Context, clients *shared.ClientFactory, appP
...
+	if err != nil {
+		return err
+	}
+	data = append(data, '\n')
+	manifestPath := filepath.Join(projectPath, "manifest.json")
+	return afero.WriteFile(clients.Fs, manifestPath, data, 0644)
```
zimeg: 🏁 suggestion: We should save the hash alongside this to avoid a confusing prompt of a changed manifest on the first "run" command:

https://github.com/slackapi/slack-cli/blob/f984c3305510b42b4c5d1e431d5fc4bd1036234e/internal/pkg/apps/install.go#L167-L176


```
$ cd my-test
$ slack run

📚 App Manifest
   Manifest values for this app are overwritten on reinstall

┃ Overwrite manifest on app settings with the project's manifest file?
┃   Yes
┃ ❱ No
```

zimeg (commented):
@srtaalej Thanks for sharing these changes 👾 ✨ 

I'm leaving a few comments on a first `run` experience around prompts and earlier errors. Other comments are on code organization with thoughts around the difference between `create` and `link` logic that might be interesting to clear up ✂️ 📠 

Testing is solid for me too but I leave a note around escaped values that are unsettling to me. I'm unsure what caused that 👻

mwbrooks (commented):
**Answers to Open Questions:**

> Open question: should we set manifest.source to remote in .slack/config.json after writing the fetched manifest? Currently left as local so the user "owns" the file going forward.

It's a good question - I think we should not change the manifest source. The main reason is that this keeps the existing functionality and it's unclear whether the developer would prefer App Settings or manifest.json when creating an app locally from App Settings.

So, since there's no clear answer, I'd rather keep the functionality unchanged. Instead, we can continue to move forward on the 2-way sync solution to remove this issue entirely.

> manifest.ts/.js (Deno SDK) projects are not supported

Totally fine to not support manifest.ts/.js. This is a Deno SDK only feature and App Settings isn't offering Deno SDK creation journeys afaik.

cmd/app/link.go
```diff
@@ -130,7 +130,7 @@ func LinkAppHeaderSection(ctx context.Context, clients *shared.ClientFactory, sh
 // When shouldConfirm is true, a confirmation prompt will ask the user if they want to
 // link an existing app and additional information is included in the header.
 // The shouldConfirm option is encouraged for third-party callers.
-func LinkExistingApp(ctx context.Context, clients *shared.ClientFactory, app *types.App, shouldConfirm bool) (err error) {
+func LinkExistingApp(ctx context.Context, clients *shared.ClientFactory, app *types.App, shouldConfirm bool) (_ *types.SlackAuth, err error) {
```
mwbrooks: suggestion: We should consider returning a single `prompts.SelectedApp` (contains an `app` and `auth`) instead of mixing a pointer out-parameter (`app *types.App`) with the new returned `types.SlackAuth`.

The `app` parameter is currently an "out-param" dressed up as an "in-param" - it's always overwritten at L171 (`*app, auth, err = promptExistingApp(...)`). Every caller passes a throwaway `&types.App{}`.

After this PR, `app` comes out via the pointer while `auth` comes out via the return value - two different output mechanisms for two values produced at the same call site. This makes the function confusing and brittle.

`prompts.SelectedApp` already pairs `App` + `Auth` and is the exact shape this function produces. Also, `cmd/app` already imports `internal/prompts`, so no new dependency.

Suggested signature:

```go
func LinkExistingApp(ctx context.Context, clients *shared.ClientFactory, shouldConfirm bool) (prompts.SelectedApp, error)
```

A zero `SelectedApp{}` (i.e. `selected.App.AppID == ""`) signals "user declined at the confirm prompt" - which is what `create.go` already checks on L243.

Caller changes:

- `LinkCommandRunE` (this file): drop the `app *types.App` param; call `_, err := LinkExistingApp(ctx, clients, false)`. Also remove the throwaway `app := &types.App{}` in `NewLinkCommand`.
- `cmd/project/init.go:113`: `_, err = app.LinkExistingApp(ctx, clients, true)`
- `cmd/project/create.go:232`: `selected, linkErr := app.LinkExistingApp(ctx, clients, false)` then `if selected.App.AppID != "" { fetchAndWriteRemoteManifest(ctx, clients, selected.Auth.Token, selected.App.AppID, absProjectPath) }` — the existing `auth != nil && linkedApp.AppID != ""` guard collapses to one condition.

Diff size is comparable to the current PR, no callers need to manufacture an empty `App{}`, and the API answers the obvious question: yes, this function returns a selected app.

mwbrooks (commented):
💬 @srtaalej This is shaping up really nice! I've left a suggestion on how we can tighten up the `LinkExistingApp` function to return the auth while also improving how it currently sets `app`. While it feels like a larger change, it turns out to be a similar line count and should help align this PR around @zimeg's feedback.

cmd/project/create.go
srtaalej: addressed in [040a3c8](https://github.com/slackapi/slack-cli/pull/585/commits/040a3c8f36154ea4802e0c4bb0df81fddef5090a) 💟

cmd/project/create.go
srtaalej: hmm I'm hesitant because the `create`+`--app` command doesn't necessarily fail if fetching remote manifest fails. the app is still created and linked successfully. only one part of the command fails in this case so maybe it would be better to provide a way to try again in follow up PRs? 
or simply telling users to run `slack manifest info --source remote` and copy that into `manifest.json` 🤔

cmd/project/create.go
srtaalej: addressed in [1092da0](https://github.com/slackapi/slack-cli/pull/585/commits/1092da0b8adb91e7e87182024b17c54e7e80c848) 🚀

cmd/project/create.go
srtaalej: addressed in [aecb626](https://github.com/slackapi/slack-cli/pull/585/commits/aecb6269241cd64dc577944d676252f4270cfb19) 🌟

cmd/app/link.go
srtaalej: thank you for this clarification @mwbrooks! I agree the `LinkExistingApp` is a bit messy 🤔 since i have [another pr that also makes changes](https://github.com/slackapi/slack-cli/pull/598) to `LinkExistingApp`, i feel like i should leave that refactor to a follow-up PR to avoid more conflicts. for now i have taken out the `auth` param i introduced and instead get it from:
```
@create.go
if linkedApp.AppID != "" {
      auth, err := clients.Auth().AuthWithTeamID(ctx, linkedApp.TeamID)
      if err != nil { return err }
      fetchErr := manifest.FetchAndWriteRemoteManifest(ctx, clients, auth.Token, ...)
      ...
```
so that `link.go` remains unchanged. what do you think of this approach?

mwbrooks: @srtaalej I think your approach to manually get the `auth` param works and keeps the impact low. So feel free to run with that!

srtaalej:
## Update

The diff is much smaller now:

- `FetchAndWriteRemoteManifest` was moved to its own file (`internal/manifest/export.go`) so it can be reused by other commands (e.g. #591)
- The `LinkExistingApp` signature is unchanged from main — no auth return was added. Instead, `create.go` resolves the token via `clients.Auth().AuthWithTeamID()` after linking
- Fixed forward slash escaping in the exported manifest JSON
- Saved the manifest hash after writing so the first `slack run` doesn't prompt about a "changed manifest"

## Follow-ups

- **Easier recovery for failed manifest sync**: Currently there's no CLI command to retry fetching the remote manifest after `create` finishes (no `slack manifest pull`). For now users can run `slack manifest info --source remote > manifest.json` as a workaround. A dedicated command would be better.
- **Refactor `LinkExistingApp`**: The `app *types.App` out-param pattern is awkward (per Michael's feedback). PR #598 is already reworking this — will align after that lands.

cmd/project/create.go
```diff
@@ -229,8 +230,27 @@ func runCreateCommand(clients *shared.ClientFactory, cmd *cobra.Command, args []
...
+		}
+
+		if linkedApp.AppID != "" {
+			auth, err := clients.Auth().AuthWithTeamID(ctx, linkedApp.TeamID)
+			if err != nil {
+				return err
+			}
+			fetchErr := manifest.FetchAndWriteRemoteManifest(ctx, clients, auth.Token, linkedApp.AppID, absProjectPath)
+			if fetchErr != nil {
+				clients.IO.PrintWarning(ctx, "%s", style.Sectionf(style.TextSection{
```
srtaalej: left a comment about this silent fail 🪂  https://github.com/slackapi/slack-cli/pull/585#discussion_r3499915667 @zimeg

internal/manifest/export.go
```diff
@@ -0,0 +1,49 @@
...
+	if err != nil {
+		return err
+	}
+	var buf bytes.Buffer
+	encoder := json.NewEncoder(&buf)
+	encoder.SetEscapeHTML(false)
+	encoder.SetIndent("", "  ")
+	if err := encoder.Encode(slackYaml.AppManifest); err != nil {
+		return err
+	}
```
zimeg: 🔭 question: Can we reuse the stringifier in adjacent package?

https://github.com/slackapi/slack-cli/blob/19cfb5913abb7573fd763ae6a10b67d8245f9a8e/internal/goutils/json.go#L62-L74

internal/manifest/export.go
zimeg, line 1: 🧪 todo: Let's add unit tests alongside internal packages for more strict checks!

cmd/project/create.go
zimeg: 👁️‍🗨️ thought: I'm worried I'd miss warnings in CI and overwrite an existing app manifest for "A001" with these commands if the manifest step of "create" fails:

```sh
$ slack create my-app -t slack-samples/bolt-js-blank-template -a A001 -E deployed
$ slack deploy
```

Recommending to save the outputs of `manifest info` to a file is a good remediation!

internal/manifest/export.go
```diff
@@ -0,0 +1,49 @@
...
+	"context"
+	"encoding/json"
+	"path/filepath"
+
+	"github.com/slackapi/slack-cli/internal/shared"
+	"github.com/spf13/afero"
+)
+
+// FetchAndWriteRemoteManifest fetches the app manifest from remote settings and writes it to the project.
+func FetchAndWriteRemoteManifest(ctx context.Context, clients *shared.ClientFactory, token, appID, projectPath string) error {
```
zimeg:
```suggestion
func SetManifestLocal(ctx context.Context, clients *shared.ClientFactory, slackYaml types.SlackYaml) error {
```

🧮 thought: This might be nice to pair with other app manifest methods although a best file structure isn't clear to me right now:

- GetManifestRemote
- GetManifestLocal
- SetManifestLocal: New!

zimeg: 📝 note: This suggestion might go well with the idea to extend `manifest info` too?

internal/manifest/export.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+	manifestPath := filepath.Join(projectPath, "manifest.json")
+	if err := afero.WriteFile(clients.Fs, manifestPath, buf.Bytes(), 0644); err != nil {
+		return err
+	}
+	hash, err := clients.Config.ProjectConfig.Cache().NewManifestHash(ctx, slackYaml.AppManifest)
+	if err != nil {
+		return err
+	}
+	return clients.Config.ProjectConfig.Cache().SetManifestHash(ctx, appID, hash)
```
zimeg, line 44: 🏆 praise: Thanks for including this!

🦔 thought: I'm now curious if this would be better as part of the `app link` command? We might pull the manifest to write a hash when linking a specific app which can improve the `link` experience outside of `create` command?

zimeg (commented):
@srtaalej Super nice to find earlier PRs building up to this feature! 💡 ✨

I'm leaving a few comments after reviewing this that I'm curious to discuss perhaps with the last point:

🌩️ **Missing outputs**: It's not clear that the template manifest is overwritten based on the output. I think a line or so toward this might be helpful.
🍵 **Hidden errors**: Erroring when writing the remote manifest can lead to unexpected cases in CI and I share when this might happen and that earlier returns are favorable to debug IMO.
🌈 **Saving to file**: The suggestion you share to remediate these errors is great and I'm curious now if we extend the `manifest info` command to support such, that we might call here?

```
$ slack manifest info --app A001 --to-file manifest.json
```

📸 Here's the output referenced!

Image

```
$ slack create my-test -t slack-samples/bolt-js-starter-template --app A07V06USK0X --team T038J6TH5PF --environment local
```

🗣️ If the `manifest info` approach seems alright I think it'd be a good separate PR that'd make this one even more focused! These are all promising changes.

internal/manifest/export.go
srtaalej: ahh thank you for sharing this!! didnt know we had one 🧐

srtaalej: adressed in [22a96a9](https://github.com/slackapi/slack-cli/pull/585/commits/22a96a932665383ae05b0b1e89221e81ca069c40)

srtaalej:
e2e failures are unrelated to this pr 😢

srtaalej:
**Update**
- Renamed `FetchAndWriteRemoteManifest` → `SetManifestLocal` to align with `GetManifestLocal/GetManifestRemote` naming
- Reused `goutils.JSONMarshalUnescapedIndent` (done in prior commit)
- Added success output when manifest is written: ✏️  Manifest — Written to `manifest.json` from remote app settings
- Unit tests added in `internal/manifest/export_test.go`

The `manifest info --to-file` and `app link` integration are tracked as follow-up work.

cmd/project/create.go
zimeg: @srtaalej I'm still hesitant to passthrough with an unexpected manifest. I'll leave a review comment but am curious if earlier thinking still stands?

zimeg (commented):
📣 Ignored errors often leaves projects in an unexpected state which I'm wanting to avoid introducing here. If these errors are common I think we can revisit it though!

But I'm now wondering if we can workaround exporting a manifest altogether?

Since we're cloning from a template that has a manifest with defined scopes and events already, would it be suitable to keep the template manifest and update the existing name:

https://github.com/slackapi/slack-cli/blob/43e5d3dd746d3c3f2ff3733c8656ef8136d6be58/internal/app/app.go#L49-L78

srtaalej:
ah so sorry @zimeg! i realize i hadn't pushed the commit that introduces the error messaging for when fetching fails! pushing incoming 😬 

> update: 
addressed in [9ad0d79](https://github.com/slackapi/slack-cli/pull/585/commits/9ad0d79a1ea76082f4bcdc6fda4e85ef92e76dc6) 🚀

srtaalej:
closing this PR over preference for a `manifest sync` command 🚀

PR #591, opened 2026-06-13
feat: add slack manifest diff command

mwbrooks:
### Changelog

> Added the `slack manifest diff` command, which prints differences between the project manifest and the app settings.

### Summary

This pull request adds a read-only `slack manifest diff` command that compares the project manifest against the app settings on Slack and prints any differences.

- Useful for CI/CD guardrails, pre-check before deploys, and debugging manifest conflicts.
- Read-only - no API mutations, no file writes.
- 91.5% test coverage of `internal/manifest`.
- Carved out from PR #543 to make that PR easier to review. The two-way `slack manifest sync` command will build on these primitives.
- Ships without an experiment gate because it's a stand-alone read-only command. Happy to gate it, if reviewers want.
- Exits 0 even when differences are found, to match `slack manifest validate` (which exits 0 even when the manifest is invalid). Standardizing exit-code semantics for scripting use-cases is a separate, project-wide conversation that should change `validate` and `diff` together.

### Preview

https://github.com/user-attachments/assets/bf05fbe2-608f-456f-837b-fc7f045d97ce

### Testing

**Testing outside of a project will error:**

```bash
$ ./bin/slack manifest diff

🚫 This is an invalid Slack app project directory (invalid_app_directory)
```

**Setup a new a project:**

```bash
$ ./bin/slack create my-app --template slack-samples/bolt-js-starter-template
$ cd my-app/
```

**Testing inside of a project with no App IDs will error:**

```bash
$ ../bin/slack manifest diff

🚫 A valid installation of this app is required to take this action (installation_required)
```

**Setup a new App ID for testing:**

```bash
$ ../bin/slack install -E local
```

**Testing a new install should return no differences:**

```bash
$ ../bin/slack manifest diff

📚 App Manifest
   Project manifest and app settings match
```

**Testing different values for project and app settings:**

```bash
$ vim manifest.json
# Set "features.bot_user.always_online" to false

$ ../bin/slack manifest diff

📚 App Manifest
   Found 1 difference between project and app settings

  features.bot_user.always_online
    Project:      false
    App settings: true
```

**Testing project missing a value:**

```bash
$ vim manifest.json
# Delete the line `"home_tab_enabled": true,`

$ ../bin/slack manifest diff

📚 App Manifest
   Found 2 differences between project and app settings

  features.app_home.home_tab_enabled
    Project:      (not set)
    App settings: true

  features.bot_user.always_online
    Project:      false
    App settings: true
```

**Testing app settings missing a value:**

```bash
$ vim manifest.json
# Add the following line to the top-level of the manifest object:
# 
# "outgoing_domains": ["example.com"],

$ ../bin/slack manifest diff

📚 App Manifest
   Found 3 differences between project and app settings

  features.app_home.home_tab_enabled
    Project:      (not set)
    App settings: true

  features.bot_user.always_online
    Project:      false
    App settings: true

  outgoing_domains
    Project:      ["example.com"]
    App settings: (not set)
```

**Clean up:**

```bash
$ lack delete -f
$ cd ../
$ rm -rf my-app/
```

### Requirements


cmd/manifest/diff.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/slackapi/slack-cli/internal/prompts"
+	"github.com/slackapi/slack-cli/internal/shared"
+	"github.com/slackapi/slack-cli/internal/style"
+	"github.com/spf13/cobra"
+)
+
+func NewDiffCommand(clients *shared.ClientFactory) *cobra.Command {
+	return &cobra.Command{
+		Use:   "diff",
+		Short: "Show differences between the project manifest and app settings",
```
mwbrooks, line 33: note: I feel we should say ~App Settings~ Slack API, but the CLI currently references App Settings, so I thought I'd stay consistent.

cmd/manifest/diff.go
```diff
@@ -0,0 +1,81 @@
...
+				return err
+			}
+
+			if !diffs.HasDifferences() {
+				clients.IO.PrintInfo(ctx, false, "\n%s", style.Sectionf(style.TextSection{
+					Emoji:     "books",
+					Text:      "App Manifest",
+					Secondary: []string{"Project manifest and app settings are in sync"},
+				}))
+				return nil
```
mwbrooks, line 76: note: We don't return a non-zero exit code when there is a manifest diff. Ideally, we should. However, `manifest validate` does not return a non-zero exist code on when the manifest is invalid and the CLI plumbing would surface a visual error to interactive users. So, we should have a follow-up task to standardize non-zero exit code handling.

internal/manifest/diff.go
```diff
@@ -0,0 +1,154 @@
...
+
+	"github.com/slackapi/slack-cli/internal/shared/types"
+)
+
+// ignoredDiffPaths are top-level manifest fields that the project may declare
+// but Slack's apps.manifest.export does not echo back. Diffs at or under these
+// paths are dropped to avoid spurious "only in project" entries on every run.
+var ignoredDiffPaths = []string{
+	"_metadata", // SDK-side schema annotations; not stored in app settings
+}
```
mwbrooks, line 30: note: The `manifest.export` API method doesn't return the `_metadata` version properties. This creates an infinite conflict between the project and app settings, so I've chosen to ignore the `_metadata` properties during the diff.

internal/manifest/diff.go
```diff
@@ -0,0 +1,198 @@
...
+// devLocalSuffixPaths are flattened paths where Slack's apps.manifest.export
+// appends " (local)" for dev-installed apps. Diffs at these paths are
+// dropped only when removing the suffix would make the values equal; real
+// renames still surface.
+var devLocalSuffixPaths = []string{
+	"display_information.name",
+	"features.bot_user.display_name",
+}
+
+const devLocalSuffix = " (local)"
```
mwbrooks, line 41: note: For local dev apps, the App Settings manifest has ` (local)` in the app name fields, but the `manifest.json` does not. This creates an infinite conflict. So, we include a list of affected fields and expected suffix ` (local)`. 

Follow-up PR should update the `GetManifestLocal` function to include ` (local)`, but this is larger change and could have unexpected side-effects. It's out-of-scope of this PR.

On the next semver major, we would like to drop ` (local)` entirely. We can discuss whether can do this sooner as well. However, the change would need to be coordinated with docs.

This implementation will continue to work after ` (local)` is removed.

internal/manifest/diff.go
```diff
@@ -0,0 +1,234 @@
...
+
+// remoteFalseDefaultPaths are flattened paths where Slack's
+// apps.manifest.export emits a default `false` for every app, even when the
+// project has not declared the field. Remote-only diffs at these paths are
+// dropped when the value is false so users do not see a phantom entry on
+// every run; a real disagreement (e.g. local sets the field to true) still
+// surfaces as a Modified diff.
+var remoteFalseDefaultPaths = []string{
+	"settings.is_mcp_enabled",
+}
```
mwbrooks: note: I don't love this, but it seems to be the safest approach that is forward-change compatible.

**Current problem:** The manifest schema defines `settings.is_mcp_enabled` as optional. However, all `manifest.exports` return `settings.is_mcp_enabled: false` when it is unset. Essentially, making it not optional. 😅 The reason is that there is a database write that transforms unset to `false`.

We could change the CLI manifest schema to default to `false`. However, this means all apps - including ROSI apps - will suddenly set this property to `false` on the next run or deploy. This could lead to new, unexpected issues.

So, I believe the safer approach is to have the `manifest diff` command aware of this BE quirk. After all, it may be fixed later.

mwbrooks (commented):
Comments for the kind reviewers 💬

internal/manifest/flatten.go
```diff
@@ -0,0 +1,80 @@
...
+
+// Flatten converts a manifest (as JSON-serializable struct) into a flat map
+// where keys are dot-notation paths and values are the leaf values.
+// Arrays are treated as leaf values (not recursed into individually) because
+// array element identity is ambiguous without a key field.
+//
+// Keys that contain literal dots (e.g. function IDs like "slack.users.lookup")
+// have those dots backslash-escaped in the path so flatten/unflatten round-
+// trip cleanly. splitPath honors the same escape sequence.
+func Flatten(manifest any) (map[string]any, error) {
```
srtaalej, line 33: super nice ⭐

internal/manifest/diff.go
srtaalej: hmm weird little be quirk 🤔 ill take a look into this on webapp 👁️

srtaalej (approved):
this is working great ⭐ love how clean the output looks 😸

internal/manifest/diff.go
zimeg: @mwbrooks IIRC finding similar but consistent to major version "1" from this method! 🤓

cmd/manifest/diff.go
```diff
@@ -0,0 +1,83 @@
...
+
+			diffs, err := manifest.Diff(localManifest.AppManifest, remoteManifest.AppManifest)
+			if err != nil {
+				return err
+			}
+
+			if !diffs.HasDifferences() {
+				clients.IO.PrintInfo(ctx, false, "\n%s", style.Sectionf(style.TextSection{
+					Emoji:     "books",
+					Text:      "App Manifest",
```
zimeg: 🔭 question: Would "Manifest Diff" or "Differences" or something else be odd to use here? I realize we call out "App Manifest" as a concept often but want to avoid using this for all commands related to the manifest.

zimeg: 🪬 ramble: A comment related finds that command errors itself:

```
$ lack app manifest
...
🚫 unknown command "manifest" for "lack app" 
```

cmd/manifest/diff_test.go
```diff
@@ -0,0 +1,90 @@
...
+				}, nil)
+				cf.AppClient().Manifest = manifestMock
+				cf.SDKConfig = hooks.NewSDKConfigMock()
+			},
+			ExpectedOutputs: []string{
+				"display_information.name",
+				"Project Name",
+				"Remote Name",
+			},
+		},
```
zimeg, line 86: 🌟 praise: Superb testcase I think!

internal/manifest/diff.go
zimeg: 🔍 suggestion(non-blocking): While we're still supporting this, could we perhaps keep strict checks against this?

🪬 ramble: We might avoid changing how local manfiests are gathered. I think an app selection `IsDev` value and perhaps manifest source as "local" would give us confidence to make this change before comparing:

https://github.com/slackapi/slack-cli/blob/2093f5321a28aaa87bbe22833f3fd154a313d166/internal/pkg/apps/install.go#L633-L639

internal/manifest/diff.go
zimeg: @mwbrooks I'm hesitant to guard against this and might consider this a new value to support in default manifests. Even if it's false? Although I most support the explorations of @srtaalej toward this.

internal/manifest/diff.go
```diff
@@ -0,0 +1,234 @@
...
+
+// DiffType describes how a field differs between local and remote.
+type DiffType int
+
+const (
+	DiffModified   DiffType = iota // Both sides have the field but with different values
+	DiffLocalOnly                  // Field exists only in local (added locally or deleted remotely)
+	DiffRemoteOnly                 // Field exists only in remote (added remotely or deleted locally)
+)
```
zimeg, line 50: 💡 praise: Fascinating approach to comparisons I'm a fan!

internal/manifest/diff.go
```diff
@@ -0,0 +1,234 @@
...
+			result.Diffs = append(result.Diffs, FieldDiff{
+				Path:       path,
+				Type:       DiffLocalOnly,
+				LocalValue: localVal,
+			})
+			continue
+		}
+		equal, err := valuesEqual(localVal, remoteVal)
+		if err != nil {
+			return nil, fmt.Errorf("failed to compare manifest values at %q: %w", path, err)
```
zimeg: ⚠️ todo: We should use a formal `slackerror` for later error code detections!

internal/manifest/diff.go
```diff
@@ -0,0 +1,234 @@
...
+				LocalValue:  localVal,
+				RemoteValue: remoteVal,
+			})
+		}
+	}
+
+	for path, remoteVal := range remote {
+		if seen[path] {
+			continue
+		}
```
zimeg, line 174: 🧠 praise: Amazing comparison to make @mwbrooks!

internal/manifest/diff_display.go
```diff
@@ -0,0 +1,107 @@
...
+func truncateRunes(s string, max int) string {
+	if max <= 3 {
+		return s
+	}
+	runes := []rune(s)
+	if len(runes) <= max {
+		return s
+	}
+	return string(runes[:max-3]) + "..."
+}
```
zimeg: 🎨 suggestion: I'd be a fan of moving this to perhaps:

```
internal/style/style.go
```

internal/manifest/diff_display.go
```diff
@@ -0,0 +1,107 @@
...
+		default:
+			local = formatValue(d.LocalValue)
+			remote = formatValue(d.RemoteValue)
+		}
+		io.PrintInfo(ctx, false, "  %s", style.Bold(d.Path))
+		io.PrintInfo(ctx, false, "    Project:      %s", local)
+		io.PrintInfo(ctx, false, "    App settings: %s", remote)
+	}
+	io.PrintInfo(ctx, false, "")
+}
```
zimeg: 🐙 suggestion: The "display" parts of `diff` might be alright to bring to the command output file unless it's planned to share ongoing? I'm hoping to avoid adding to "internal" packages when possible-

```
cmd/manifest/diff.go
```

internal/manifest/flatten.go
```diff
@@ -0,0 +1,80 @@
...
+
+// SortedKeys returns the keys of a flat map in sorted order for deterministic output.
+func SortedKeys(m map[string]any) []string {
+	keys := make([]string, 0, len(m))
+	for k := range m {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+	return keys
+}
```
zimeg, line 81: 🔏 praise: Huge fan of this flattening and sorting approach!

internal/manifest/flatten.go
```diff
@@ -0,0 +1,80 @@
...
+// Arrays are treated as leaf values (not recursed into individually) because
+// array element identity is ambiguous without a key field.
+//
+// Keys that contain literal dots (e.g. function IDs like "slack.users.lookup")
+// have those dots backslash-escaped in the path so flatten/unflatten round-
+// trip cleanly. splitPath honors the same escape sequence.
+func Flatten(manifest any) (map[string]any, error) {
+	data, err := json.Marshal(manifest)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal manifest: %w", err)
```
zimeg: ⚠️ todo: Let's bring `slackerror` to this package?

internal/manifest/flatten.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+	var raw map[string]any
+	if err := json.Unmarshal(data, &raw); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal manifest: %w", err)
```
zimeg: ⚠️ todo: Let's bring `slackerror` to this package!

internal/manifest/diff_display.go
```diff
@@ -0,0 +1,107 @@
...
+			local = formatValue(d.LocalValue)
+			remote = absentValue
+		case DiffRemoteOnly:
+			local = absentValue
+			remote = formatValue(d.RemoteValue)
+		default:
+			local = formatValue(d.LocalValue)
+			remote = formatValue(d.RemoteValue)
+		}
+		io.PrintInfo(ctx, false, "  %s", style.Bold(d.Path))
```
zimeg:
```suggestion
		io.PrintInfo(ctx, false, "  %s", d.Path)
```

👁️‍🗨️ quibble(non-blocking): The bold is somewhat strong to me and I think the different indentation might be enough to callout sections from values?

internal/manifest/diff_test.go
```diff
@@ -0,0 +1,291 @@
...
+	}
+	result, err := Diff(local, remote)
+	require.NoError(t, err)
+	require.True(t, result.HasDifferences())
+	require.Len(t, result.Diffs, 1)
+	assert.Equal(t, "settings.is_mcp_enabled", result.Diffs[0].Path)
+	assert.Equal(t, DiffModified, result.Diffs[0].Type)
+}
+
+func ptrBool(b bool) *bool { return &b }
```
zimeg, line 275:
```suggestion
```

🧪 suggestion: Let's use "new" syntax here!

```diff
-			IsMCPEnabled: ptrBool(false),
+			IsMCPEnabled: new(false),
```

internal/manifest/diff_display_test.go
```diff
@@ -0,0 +1,201 @@
...
+			// max <= 3 leaves no room for the "..." sentinel, so the
+			// helper short-circuits and returns the input as-is.
+			input:    "abcdef",
+			max:      3,
+			expected: "abcdef",
+		},
+		"multi-byte runes are not cut mid-character": {
+			// Each emoji is 4 bytes in UTF-8 but one rune. Byte-based
+			// slicing would split the middle emoji.
+			input:    "🐶🐱🐭🐹🐰🦊🐻🐼🐨🐯🦁🐮",
```
zimeg: 🌞 praise: Good runes!

zimeg (approved):
@mwbrooks Amazing changes to be landing! I'm excited to see how we include these outputs elsewhere 😉 

Leaving a few handful of comments that focus on a few things:

- Command output: Both wording and placement of code but nothing problematic
- Unexpected fields appearing: I think we should do less magic and accept that some fields might appear if the project manifest doesn't have these specified. To me that's an abrupt sign the project should be updated.

Once things are seeming right to you I think it'll be exciting to release but please let me know if I can test more 🔍  ✨

srtaalej:
**Update: Addressed review feedback from @zimeg:**
1. Renamed section header from `App Manifest` → `Manifest Diff`
2. Removed bold styling from diff paths (indentation alone is sufficient)
3. Replaced `fmt.Errorf` with `slackerror.Wrap` using a new `ErrAppManifestCompare` error code in `diff.go` and `flatten.go`
4. Moved `truncateRunes` to `internal/style/` as an exported `TruncateRunes` utility
5. Moved display logic (`DisplayDiffs`, `formatValue`) from `internal/manifest/` to `cmd/manifest/` since it's presentation-only and not shared
6. Gated the `(local)` suffix filter behind `isDev` so it only suppresses for dev-installed apps
7. Removed `is_mcp_enabled` suppression — the field now surfaces in diffs, helping users discover MCP as a feature

In commits: [77be121](https://github.com/slackapi/slack-cli/pull/591/commits/77be121677b6824430bfd00b3f259f53e115554a), [efa30b4](https://github.com/slackapi/slack-cli/pull/591/commits/efa30b42b48f3cee5720e2a47c035387dea5af4b), [690740c](https://github.com/slackapi/slack-cli/pull/591/commits/690740c2480d73d87147421e0a4f228170780ae9)

internal/manifest/diff.go
mwbrooks: This is outdated, so it's no longer the latest code.

mwbrooks:
Wow @srtaalej, thanks for applying all of the feedback! 🙇🏻 

Before merging, please test this thoroughly. I had some hesitation around adding all of this feedback, but if it works let's go with it!

PR #598, opened 2026-06-22
feat(create): cleanup app creation outputs when `slack create -app` passed

srtaalej:
### Changelog

Reduced verbose output when using `create --app` to link an existing app during project creation.

### Summary
When `--app` is provided to `slack create`, the linking step previously printed multiple noisy sections (App Link header, App Manifest details, App footer, and "Added existing app" message). Now it only prints the single `🏠 App` summary section showing team, App ID, Team ID, User ID, and Status.

#### Before

before

#### After

after

### Testing

- `make test`
- Manual test - `./bin/slack create my-test -t slack-samples/bolt-js-starter-template --app A123 --team T123 --environment local`

### Notes

Adresses comments on PR #565 - [Improve the output experience for create when --app is provided.](https://github.com/slackapi/slack-cli/pull/565#:~:text=Improve%20the%20output%20experience%20for%20create%20when%20%2D%2Dapp%20is%20provided.)

### Requirements

### DIFF
--- a/cmd/app/link.go
+++ b/cmd/app/link.go
@@ -32,9 +32,6 @@ import (
 	"github.com/spf13/cobra"
 )
 
-// LinkAppConfirmPromptText is displayed when prompting to add an existing app
-const LinkAppConfirmPromptText = "Do you want to add an existing app?"
-
 // appLinkFlagSet contains flag values to reference
 type appLinkFlagSet struct {
 	environmentFlag string
--- a/cmd/app/link.go
+++ b/cmd/app/link.go
@@ -192,6 +192,37 @@ func LinkExistingApp(ctx context.Context, clients *shared.ClientFactory, app *ty
 	return nil
 }
 
+// LinkExistingAppQuiet links an existing app without verbose output. It resolves
+// the team and environment, validates the app, saves it, and prints only the app
+// summary. Used by `create --app` where the surrounding create output is enough.
+func LinkExistingAppQuiet(ctx context.Context, clients *shared.ClientFactory, app *types.App) error {
+	linkedApp, auth, err := promptExistingApp(ctx, clients)
+	if err != nil {
+		return err
+	}
+	*app = linkedApp
+
+	appIDs := []string{app.AppID}
+	_, err = clients.API().GetAppStatus(ctx, auth.Token, appIDs, app.TeamID)
+	if err != nil {
+		return err
+	}
+
+	err = saveAppToJSON(ctx, clients, *app)
+	if err != nil {
+		clients.IO.PrintDebug(ctx, "Error saving app to file when linking existing app: %s", err)
+		return err
+	}
+
+	clients.IO.PrintInfo(ctx, false, "\n%s", style.Sectionf(style.TextSection{
+		Emoji:     "house",
+		Text:      "App",
+		Secondary: formatListSuccess([]types.App{*app}),
+	}))
+
+	return nil
+}
+
 // LinkAppFooterSection displays the details of app that was added to the project.
 func LinkAppFooterSection(ctx context.Context, clients *shared.ClientFactory, app *types.App) {
 	clients.IO.PrintInfo(ctx, false, "\n%s", style.Sectionf(style.TextSection{
--- a/cmd/app/link.go
+++ b/cmd/app/link.go
@@ -98,7 +98,7 @@ func LinkCommandRunE(ctx context.Context, clients *shared.ClientFactory, app *ty
 	clients.IO.PrintInfo(ctx, false, "")
 
 	// Header section
-	LinkAppHeaderSection(ctx, clients, false)
+	LinkAppHeaderSection(ctx, clients)
 
 	// App Manifest section
 	manifestSource, err := clients.Config.ProjectConfig.GetManifestSource(ctx)
@@ -141,23 +141,14 @@ func LinkCommandRunE(ctx context.Context, clients *shared.ClientFactory, app *ty
 }
 
 // LinkAppHeaderSection displays a section explaining how to find existing apps.
-// External callers can use extraSecondaryText to show additional information.
-// When shouldConfirm is true, additional information is included in the header
-// explaining how to link apps, in case the user declines.
-func LinkAppHeaderSection(ctx context.Context, clients *shared.ClientFactory, shouldConfirm bool) {
-	var secondaryText = []string{
-		"Add an existing app from app settings",
-		"Find your existing apps at: " + style.Underline("https://api.slack.com/apps"),
-	}
-
-	if shouldConfirm {
-		secondaryText = append(secondaryText, "Manually add apps later with "+style.Commandf("app link", true))
-	}
-
+func LinkAppHeaderSection(ctx context.Context, clients *shared.ClientFactory) {
 	clients.IO.PrintInfo(ctx, false, "%s", style.Sectionf(style.TextSection{
-		Emoji:     "house",
-		Text:      "App Link",
-		Secondary: secondaryText,
+		Emoji: "house",
+		Text:  "App Link",
+		Secondary: []string{
+			"Add an existing app from app settings",
+			"Find your existing apps at: " + style.Underline("https://api.slack.com/apps"),
+		},
 	}))
 }
 
--- a/cmd/app/link.go
+++ b/cmd/app/link.go
@@ -80,7 +80,7 @@ func NewLinkCommand(clients *shared.ClientFactory) *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			ctx := cmd.Context()
 			clients.IO.PrintTrace(ctx, slacktrace.AppLinkStart)
-			return LinkCommandRunE(ctx, clients, app)
+			return LinkCommandRunE(ctx, clients, app, false, false)
 		},
 		PostRunE: func(cmd *cobra.Command, args []string) error {
 			ctx := cmd.Context()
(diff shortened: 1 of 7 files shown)


cmd/app/link.go
```diff
@@ -192,6 +192,37 @@ func LinkExistingApp(ctx context.Context, clients *shared.ClientFactory, app *ty
...
+	}
+
+	clients.IO.PrintInfo(ctx, false, "\n%s", style.Sectionf(style.TextSection{
+		Emoji:     "house",
+		Text:      "App",
+		Secondary: formatListSuccess([]types.App{*app}),
+	}))
+
+	return nil
+}
```
zimeg: 🔭 thought: This is near identical to the lines above and I'm wondering if we can somehow keep a single source of truth for this logic?

https://github.com/slackapi/slack-cli/blob/a0f86d78f3c640faf9762f68a90c801510900941/cmd/app/link.go#L170-L190

zimeg (commented):
@srtaalej Super thanks for taking a pass at better `create` outputs. I'm leaving some comments to help us maintain these commands as ongoing changes are introduced with more "DRY" functions 👁️‍🗨️

cmd/app/link.go
srtaalej: sure good call! @zimeg to clarify though, do you mean the entire `link` command should return the way `LinkExistingAppQuiet` is returning? or should we have something like 
```
func LinkExistingApp(ctx context.Context, clients *shared.ClientFactory, app *types.App, shouldConfirm bool, quiet bool) (err error) {
	if !quiet {
		// Header section
		LinkAppHeaderSection(ctx, clients, shouldConfirm)
	}

```

zimeg: @srtaalej I missed the edited mention! question But separating output from logic might make this possible? I think outputs and even the "should confirm" logic should be moved to adjacent functions:

- app.LinkExistingApp: What I permalink above
- app.LinkCommandRunE: Perhaps most the "link" outputs not meant to share

srtaalej: ah got it 🙌 !

srtaalej:
@zimeg thank you for your earlier review! i believe your feedback has been addressed in [7b615b7](https://github.com/slackapi/slack-cli/pull/598/commits/7b615b7ae8abf5ca69a5b66e1e9658124216e88c) 🚀 good call on separating the outputs and linking logic! 💟

cmd/app/link.go
```diff
@@ -93,15 +93,71 @@ func NewLinkCommand(clients *shared.ClientFactory) *cobra.Command {
...
 	if err != nil {
 		return err
 	}
 
+	// App summary section
+	clients.IO.PrintInfo(ctx, false, "%s", style.Sectionf(style.TextSection{
+		Emoji:     "house",
+		Text:      "App",
+		Secondary: formatListSuccess([]types.App{*app}),
+	}))
```
zimeg, line 126: 🏗️ question: I'm a fan of separating outputs and logic overall but am curious about including this within `LinkExistingApp` to keep the function signature minimal? I share another comment about ongoing refactor but am curious if this seems right or not:

```go
func LinkCommandRunE(ctx context.Context, clients *shared.ClientFactory, app *types.App)
```

📣 ramble: So that we avoid conditional statements in these functions!

cmd/project/init.go
```diff
@@ -110,7 +110,7 @@ func projectInitCommandRunE(clients *shared.ClientFactory, cmd *cobra.Command, a
 	_ = create.InstallProjectDependencies(ctx, clients, projectDirPath)
 
 	// Add an existing app to the project
-	err = app.LinkExistingApp(ctx, clients, &types.App{}, true)
+	err = app.LinkCommandRunE(ctx, clients, &types.App{}, true, false)
```
zimeg: 💡 thought: No blocker but we might want to refactor the confirmation prompt to be here instead of within the `link` command?

⚡ ramble: I find boolean arguments unclear and also think a command should run if it's called - it'd be up to the caller to decide if that command should run instead!

zimeg (commented):
@srtaalej Super appreciate changes toward this! I'm feeling a lot better about how the code structures are and I'm leaving a few more thoughts ongoing if these might move things in a good direction?

cmd/app/link_test.go
```diff
@@ -572,53 +572,16 @@ func Test_Apps_Link(t *testing.T) {
 }
 
 func Test_Apps_LinkAppHeaderSection(t *testing.T) {
-	tests := map[string]struct {
```
zimeg, line 575: 🏗️ quibble: Keeping table tests structured - even if one case - instead of a flattened single case might encourage additional cases

cmd/app/link_test.go
```diff
@@ -572,53 +572,16 @@ func Test_Apps_Link(t *testing.T) {
...
-			},
-		},
-		"When shouldConfirm is true": {
-			shouldConfirm: true,
-			expectedOutputs: []string{
-				"Add an existing app from app settings",
-				"Find your existing apps at: https://api.slack.com/apps",
-				"Manually add apps later with",
-			},
-		},
```
zimeg, line 597: 🔬 note: Existing tests continue to cover this removed case in the `init` command.

https://github.com/slackapi/slack-cli/blob/4565149496dec13280322490325d99c1badda29c/cmd/project/init_test.go#L56-L207

cmd/project/create.go
```diff
@@ -229,9 +229,15 @@ func runCreateCommand(clients *shared.ClientFactory, cmd *cobra.Command, args []
...
-		if err := app.LinkExistingApp(ctx, clients, &types.App{}, false); err != nil {
+		linkedApp := &types.App{}
+		if err := app.LinkExistingApp(ctx, clients, linkedApp); err != nil {
 			return err
 		}
+		clients.IO.PrintInfo(ctx, false, "%s", style.Sectionf(style.TextSection{
+			Emoji:     "house",
+			Text:      "App",
+			Secondary: app.FormatListSuccess([]types.App{*linkedApp}),
+		}))
```
zimeg, line 240: 🌟 praise: Inlining this output within `create` seems to've fixed an extra linebreak!

cmd/project/create.go
```diff
@@ -229,9 +229,15 @@ func runCreateCommand(clients *shared.ClientFactory, cmd *cobra.Command, args []
 		defer func() {
 			_ = os.Chdir(originalDir)
 		}()
-		if err := app.LinkExistingApp(ctx, clients, &types.App{}, false); err != nil {
+		linkedApp := &types.App{}
```
zimeg, line 232: 🔭 note: IIRC adjacent PR has discussion about returning instead of providing an `app` argument. No blocker but it has support in eventual follow up!

cmd/app/link.go
zimeg: 🗣️ note: Resolved in earlier commits

cmd/app/link.go
zimeg: 👾 suggestion: This is now used for `init` and not within `link` so I suggest we move it to the "project" package and make it private?

cmd/project/init.go
```diff
@@ -109,11 +109,26 @@ func projectInitCommandRunE(clients *shared.ClientFactory, cmd *cobra.Command, a
...
-	// Add an existing app to the project
-	err = app.LinkExistingApp(ctx, clients, &types.App{}, true)
+	// Prompt to add an existing app to the project
+	app.LinkAppHeaderSection(ctx, clients)
+	clients.IO.PrintInfo(ctx, false, "   %s\n", "Manually add apps later with "+style.Commandf("app link", true))
+	proceed, err := clients.IO.ConfirmPrompt(ctx, app.LinkAppConfirmPromptText, true)
 	if err != nil {
-		// Display the error but continue to init
-		clients.IO.PrintError(ctx, "%s", err.Error())
+		clients.IO.PrintDebug(ctx, "Error prompting to add an existing app: %s", err)
```
zimeg:
```suggestion
		return err
```

🗣️ suggestion: The earlier debug might've been too generic? Since we've moved this prompt callsite to `init` we might want to return this prompt error to stop the command after interrupts?

cmd/project/init.go
```diff
@@ -109,11 +109,26 @@ func projectInitCommandRunE(clients *shared.ClientFactory, cmd *cobra.Command, a
 	// Existing projects initialized always default to config.ManifestSourceLocal.
 	_ = create.InstallProjectDependencies(ctx, clients, projectDirPath)
 
-	// Add an existing app to the project
-	err = app.LinkExistingApp(ctx, clients, &types.App{}, true)
+	// Prompt to add an existing app to the project
+	app.LinkAppHeaderSection(ctx, clients)
+	clients.IO.PrintInfo(ctx, false, "   %s\n", "Manually add apps later with "+style.Commandf("app link", true))
```
zimeg:
```suggestion
```

🪓 suggestion: Let's remove this to avoid duplication in later output

Image

cmd/project/init.go
```diff
@@ -109,11 +109,26 @@ func projectInitCommandRunE(clients *shared.ClientFactory, cmd *cobra.Command, a
...
+	proceed, err := clients.IO.ConfirmPrompt(ctx, app.LinkAppConfirmPromptText, true)
```
zimeg: 🌚 praise: Bringing this confirmation prompt to the command that uses it is great! It unlocks more meaningful checks IMHO like checking if an app is saved before prompting.

zimeg (approved):
@srtaalej Thanks for the huge cleanups to output 🎨 ✨ 

I'm leaving a handful more comments with some suggestions on tests and error handling and one more line of output. I'll update the preview with a current difference now too 📸

PR #615, opened 2026-07-14
feat: recommend the official slack plugin to claude code users

zimeg:
### Changelog

The CLI now recommends the official Slack plugin to Claude Code users. No change for anyone running the CLI outside Claude Code.

### Summary

When the CLI runs inside Claude Code (detected via `CLAUDECODE=1`), it emits a one-line `<claude-code-hint />` marker to stderr recommending the official `slack@claude-plugins-official` plugin. Claude Code strips the marker before the output reaches the model (so it never counts toward token usage), dedupes it per plugin and per session, and shows the user a one-time install prompt. Claude Code never installs anything automatically — the user always confirms.

Protocol reference: https://code.claude.com/docs/en/plugin-hints

Design decisions worth calling out:

- **Placement.** The emitter lives beside the existing `CLAUDECODE` detection (`GetAIAgent`) in `internal/useragent`, since that package already owns "behavior conditioned on which AI agent is running us." A dedicated `internal/plugins` (or `internal/agents`) package felt premature for a ~10-line, single-vendor feature — Codex/Gemini/etc. have no equivalent protocol today. If a second agent-conditioned behavior or a second vendor protocol lands, that's the trigger to extract.
- **Two call sites.** `root.go`'s `PersistentPreRunE` covers every normal subcommand. Cobra serves `--help` *before* `PersistentPreRunE` runs, so the custom help func (`cmd/help/help.go`) needs its own emit to cover `slack --help` and `slack <cmd> --help` (two of the touchpoints the docs recommend).
- **stderr, on its own line.** The marker is written to stderr via `IO.WriteErr()` so it stays out of stdout pipelines and out of the help text, and the protocol's one hard requirement (own line) is satisfied by `Fprintln`.

### Preview

Emitted on stderr only when `CLAUDECODE=1`:

```
$ CLAUDECODE=1 slack version 2>&1 1>/dev/null
<claude-code-hint v="1" type="plugin" value="slack@claude-plugins-official" />
```

Nothing is emitted otherwise.

### Testing

Unit tests:
- `internal/useragent/hint_test.go` — emits on its own line when `CLAUDECODE=1`; no-op when unset or set to another value.
- `cmd/help/help_test.go` — help func emits the hint on stderr (not stdout) inside Claude Code; nothing outside.

Manual end-to-end against a built binary:
- `CLAUDECODE=1 slack version` → 1 hint on stderr, 0 on stdout ✅
- `slack version` (no `CLAUDECODE`) → 0 bytes on stderr ✅
- `CLAUDECODE=1 slack --help` and `slack auth --help` → 1 hint on stderr each, help text unaffected on stdout ✅
- `slack --help` without `CLAUDECODE` → 0 ✅

`go build ./...`, `go test ./cmd/... ./internal/useragent/...`, and `golangci-lint` on the touched packages all pass.

### Notes

- **Depends on an official-marketplace listing.** The hint only renders for plugins in Anthropic's curated `claude-plugins-official` marketplace; hints pointing elsewhere are silently dropped. Until a `slack` plugin is listed there, this marker is emitted but not shown. This needs to be coordinated with an Anthropic contact — flagging so it isn't a surprise.
- Emitting on every invocation is intentional and has no downside: Claude Code dedupes by plugin and shows at most one hint prompt per session.

### Requirements

### DIFF
--- a/cmd/help/help.go
+++ b/cmd/help/help.go
@@ -77,7 +77,7 @@ func PrintHelpTemplate(cmd *cobra.Command, data style.TemplateData) {
 		cmd.PrintErrln(err)
 	}
 
-	// Recommend the official Slack plugin when running inside Claude Code.
+	// Recommend the official Slack plugin when running inside Claude Code
 	useragent.EmitClaudeCodePluginHint(cmd.ErrOrStderr())
 }
 
--- a/cmd/help/help.go
+++ b/cmd/help/help.go
@@ -51,11 +51,6 @@ func HelpFunc(
 			"Experiments": experiments,
 		}
 		PrintHelpTemplate(cmd, data)
-
-		// Cobra serves help before PersistentPreRunE runs, so the root emit is
-		// skipped on `--help` paths. Recommend the official Slack plugin here too
-		// when running inside Claude Code. No-op in every other environment.
-		useragent.EmitClaudeCodePluginHint(clients.IO.WriteErr())
 	}
 }
 
@@ -81,6 +76,9 @@ func PrintHelpTemplate(cmd *cobra.Command, data style.TemplateData) {
 	if err != nil {
 		cmd.PrintErrln(err)
 	}
+
+	// Recommend the official Slack plugin when running inside Claude Code
+	useragent.EmitClaudeCodePluginHint(cmd.ErrOrStderr())
 }
 
 // ════════════════════════════════════════════════════════════════════════════════
--- a/cmd/help/help.go
+++ b/cmd/help/help.go
@@ -77,7 +77,7 @@ func PrintHelpTemplate(cmd *cobra.Command, data style.TemplateData) {
 		cmd.PrintErrln(err)
 	}
 
-	// Recommend the official Slack plugin when running inside Claude Code
+	// Recommend the official Slack plugin when running inside Claude Code.
 	useragent.EmitClaudeCodePluginHint(cmd.ErrOrStderr())
 }
 
--- a/cmd/help/help.go
+++ b/cmd/help/help.go
@@ -21,6 +21,7 @@ import (
 	"github.com/slackapi/slack-cli/internal/experiment"
 	"github.com/slackapi/slack-cli/internal/shared"
 	"github.com/slackapi/slack-cli/internal/style"
+	"github.com/slackapi/slack-cli/internal/useragent"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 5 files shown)


internal/useragent/hint.go
```diff
@@ -0,0 +1,37 @@
...
+	"fmt"
+	"io"
+	"os"
+)
+
+// claudeCodePluginHint is the marker to recommend the official Slack plugin to
+// users of Claude Code.
+//
+// https://code.claude.com/docs/en/plugin-hints
+const claudeCodePluginHint = `<claude-code-hint v="1" type="plugin" value="slack@claude-plugins-official" />`
```
zimeg, line 27: ⚠️ note: I can't seem to have `claude` prompt but this matches spec!

zimeg (commented):
📝 A note for reviewers before I fall into too much a rabbit hole.

internal/useragent/hint.go
WilliamBergamin: Did you double check that the plugin was not already installed on your system?

mwbrooks, line 27: praise: I agree that `internal/useragent` reads well. If we get a second agent-conditioned behaviour, then we can consider a new pacakge.

cmd/help/help.go
```diff
@@ -75,6 +76,9 @@ func PrintHelpTemplate(cmd *cobra.Command, data style.TemplateData) {
 	if err != nil {
 		cmd.PrintErrln(err)
 	}
+
+	// Recommend the official Slack plugin when running inside Claude Code
```
mwbrooks, line 80: praise: Nice catch on the two call sites. This is likely a high-hit-rate call site and is subtle enough for most people to miss.

internal/useragent/hint.go
mwbrooks: I'm not seeing the marker prompt either. I spent some time digging into this and I have a lead:

`CLAUDE_CODE_USE_BEDROCK=1` disables plugin hints entirely. So, our Slack work laptops will always disable this suggested prompt.

I'll test on my personal computer and share the results.

mwbrooks (commented):
Thanks for putting this together @zimeg! 🎉 🙇🏻 

The reason the plugin hints are showing is because our work computer have  `CLAUDE_CODE_USE_BEDROCK=1`, which disables all hints. I'll follow-up after testing on my personal computer.

While debugging, I learned that we can use either stdout or stderr, but Claude recommends stderr. I agree that this is a better place for it.

internal/useragent/hint.go
zimeg: @mwbrooks Fantastic intuition! 🧠 ✨  It makes me eager to test with a different build as well!

mwbrooks (approved):
✅ Good news @zimeg, the hint appears on a regular Claude Code installation!

🖼️ Screenshot below:

Image

zimeg:
@mwbrooks Thanks immense for testing this! I confirmed similar on a personal machine too so am feeling alright with landing this as correct 🔮 🚀 

@WilliamBergamin Thanks for both finding this feature so useful and reviews also 🎁

cmd/help/help.go
zimeg: @mwbrooks I'm not thrilled with the spread of this feature but it's focused enough that I think we can iterate to improvement as needed. I do agree that an ask for "help" is ideal to output!

PR #616, opened 2026-07-15
fix: treat empty on-disk JSON state files as empty state

mwbrooks:
## Changelog

> Fixed a recurring `unable_to_parse_json` error caused by empty JSON files (credentials, config, or app state) left behind by an interrupted write - the CLI now treats them as empty state and recovers automatically.

## Summary

This pull request helps to address `unable_to_parse_json` internal errors that some users are experiencing.

## Problem

`afero.WriteFile` opens with `O_TRUNC` before writing. If the process is interrupted (SIGKILL, host crash, oom, user cancel) after the truncate but before the write completes, the JSON state file is left at **zero bytes**. Every subsequent CLI invocation then reads the file, calls `json.Unmarshal(nil, …)` on the empty contents, and raises `ErrUnableToParseJSON`. The file never self-heals: the reader errors out before the caller reaches the code path that would rewrite it.

Affected state files:

- `~/.slack/credentials.json`
- `~/.slack/config.json`
- Project `.slack/config.json`
- Project `.slack/apps.json`
- Project `.slack/apps.dev.json`

## What this PR changes

Adds a single helper `goutils.IsEmptyJSON([]byte) bool` and updates the five readers so that an empty (or whitespace-only) file returns the same zero-value state as the "file does not exist" branch, rather than surfacing a parse error:

- `internal/auth/auth.go` — `auths()` returns an empty `AuthByTeamID`
- `internal/config/system.go` — `UserConfig` returns a config with `Surveys` initialised
- `internal/config/project.go` — `ReadProjectConfigFile` returns a config with `Surveys` initialised
- `internal/app/app_client.go` — `readDeployedApps` and `readLocalApps` return empty maps **and re-serialise a valid `{}`** so the file self-heals on the next write cycle

**Genuinely malformed JSON still returns `ErrUnableToParseJSON`.** Existing broken-JSON tests are untouched.

## Test plan

- [x] `Test_IsEmptyJSON` covers empty, whitespace-only, `{}`, and populated inputs
- [x] `Test_Client_auths` empty-file subtest returns no auths, no error
- [x] `Test_SystemConfig_UserConfig` empty-file subtest returns zero-valued config, no error
- [x] `Test_ReadProjectConfigFile` empty-file subtest returns zero-valued config, no error
- [x] `AppClient.readDeployedApps` empty-file subtest returns empty apps and writes `{}` back to disk
- [x] `AppClient.readLocalApps` empty-file subtest returns empty local apps and writes `{}` back to disk
- [x] `go test ./internal/... ./cmd/...` passes
- [x] `make lint` passes with zero issues

## Follow-ups (not in this PR)

- **Atomic writes** — replace `afero.WriteFile` with a temp-file + `Rename` pattern across the five state writers so the durable fix is that these files can never reach a 0-byte state in the first place.
- **Centralise the read-JSON-state pattern** — the "stat → read → check empty → unmarshal → wrap error" sequence is now duplicated across five call sites. Move it into `goutils` (or a new `internal/statefile` package) so future state files pick up the guard for free.
- **`internal/api/activity.go`** — `ErrHTTPResponseInvalid` on the activity poll loop shows the same "thousands of events off one broken response" shape; worth a matching guard.
- **`internal/update/sdk.go`** — retry with backoff on persistently malformed SDK hook response so a broken SDK doesn't dominate the error telemetry either.


internal/config/system.go
```diff
@@ -127,6 +128,17 @@ func (c *SystemConfig) UserConfig(ctx context.Context) (*SystemConfig, error) {
...
 	}
 
+	// Treat an empty (or whitespace-only) config file as "no user config saved
+	// yet" rather than a parse error. This state occurs when a prior write was
+	// truncated but not completed (e.g. a process was interrupted between
+	// afero.WriteFile's O_TRUNC and the actual write). Without this guard,
+	// every subsequent CLI invocation reads the same 0-byte file and logs
+	// ErrUnableToParseJSON in a hot loop.
+	if goutils.IsEmptyJSON(configFileBytes) {
+		config.Surveys = map[string]SurveyConfig{}
```
mwbrooks, line 138: note: I thought there might be an initialize function, but this same pattern is used elsewhere.

mwbrooks (commented):
Comments for the kind reviewers

internal/app/app_client.go
```diff
@@ -297,6 +298,22 @@ func (ac *AppClient) readDeployedApps() error {
 		return err
 	}
 
+	// Treat an empty (or whitespace-only) file as "no apps saved yet" rather
+	// than a parse error. This state occurs when a prior write was truncated
+	// but not completed (e.g. a process was interrupted between afero.WriteFile's
+	// O_TRUNC and the actual write). Without this guard, every subsequent CLI
+	// invocation logs ErrUnableToParseJSON on the same file.
```
zimeg, line 305: 🔭 thought: The follow up "atomic" writes is a clever and good workaround to have! I'm worried this patch might cause existing project apps to seem missing instead of a JSON error appearing, but the root cause is unrelated to this PR.

internal/auth/auth.go
```diff
@@ -216,6 +216,16 @@ func (c *Client) auths(ctx context.Context) (map[string]types.SlackAuth, error)
...
+	// stored" rather than a parse error. This state occurs when a prior write
+	// was truncated but not completed (e.g. a process was interrupted between
+	// afero.WriteFile's O_TRUNC and the actual write). Without this guard,
+	// every subsequent CLI invocation reads the same 0-byte file and logs
+	// ErrUnableToParseJSON in a hot loop.
+	if goutils.IsEmptyJSON(raw) {
+		return types.AuthByTeamID{}, nil
+	}
+
 	err = json.Unmarshal(raw, &auths)
```
zimeg, line 229: 👾 thought: Perhaps covered in the "atomic" reads and writes once more, but I find these functions pair well together and am curious if our own `JSON` package might handle blank files instead of addressing this at each callsite?

internal/goutils/json.go
```diff
@@ -84,3 +84,14 @@ func JSONUnmarshal(data []byte, v interface{}) error {
...
+// IsEmptyJSON returns true when the provided bytes are empty or contain only
+// whitespace. This matches the state left behind when a JSON state file was
+// truncated but not yet rewritten, e.g. after a process was interrupted between
+// afero.WriteFile's O_TRUNC and the actual write. Callers that read on-disk
+// config/auth/app state should treat this case as "empty state" rather than a
+// parse error, otherwise every subsequent CLI invocation logs
+// ErrUnableToParseJSON until the user manually clears the file.
+func IsEmptyJSON(data []byte) bool {
+	return len(bytes.TrimSpace(data)) == 0
+}
```
zimeg, line 97: 🪬 quibble: Calling this "empty JSON" to me is confusing because I'd expect:

```json
{}
```

And instead might prefer to call this `IsBlankFile` or something similar but of course the description above is amazing at making this clear.

zimeg (approved):
@mwbrooks LGTM! I think this is alright to merge with plans to follow up with "atomic" writes. I emphasize this because guarding against these empty error cases isn't as ideal as preventing such. The pattern you suggest is fascinating to me 🧠 💡

I'm thinking we should merge this for next release after finding testing performs as expected!

srtaalej (approved):
tested and it looks great!

PR #617, opened 2026-07-16
feat: add --force-remote flag to manifest sync

srtaalej:
## Summary

- Adds `--force-remote` flag to `slack manifest sync` that pulls app settings to the local project manifest without prompting
- Symmetric counterpart to `--force` (which pushes local → remote)
- Validates mutual exclusion: using both `--force` and `--force-remote` returns a clear error
- Updates non-TTY remediation message to mention both force options

## Context

Follow-up to #543 (two-way manifest sync). That PR's notes call out `--force-remote` as a planned follow-up for non-interactive "remote wins" workflows — CI pipelines, onboarding devs seeding a local manifest from deployed state, etc.

## Test plan

```bash
# 1. Create a new project
  ./bin/slack create my-force-remote-test
```
```bash
  # 2. Install the app (creates remote manifest on app settings)
  cd my-force-remote-test
  ../bin/slack install -e manifest-sync
```
```bash
  # 3. Verify sync shows no differences initially
  ../bin/slack manifest sync -e manifest-sync
```
4. Make a local change to manifest.json (to create drift) e.g., edit display_information.description in manifest.json
```bash
  # 5. Test --force-remote pulls remote values (overwrites your local edit)
  ../bin/slack manifest sync --force-remote -e manifest-sync
```
```bash
  # 6. Test mutual exclusion error
  ../bin/slack manifest sync --force --force-remote -e manifest-sync
```

For testing remote-side drift (the more interesting case for --force-remote), edit the manifest at https://app.slack.com/app-settings after step 2, then run step 5 — it should pull the remote change into your local manifest.json without prompting.


internal/config/config.go
```diff
@@ -53,6 +53,7 @@ type Config struct {
 	DeprecatedWorkspaceFlag string
 	DisableTelemetryFlag    bool
 	ForceFlag               bool
+	ForceRemoteFlag         bool
```
zimeg, line 56: 👁️‍🗨️ note: We should avoid adding command flags to these configurations. For now it's alright with suggestions of #543 making it more clear that this is a flag for the `sync` command.

👾 ramble: I do find it curious to workaround the flag structures otherwise but fear it blurs concern of what a command can do here.

zimeg (approved):
@srtaalej LGTM! Thanks for keeping this fast follow separate! 🌠 

I'm leaving a note of change that we can revisit with wider refactors of earlier changes. No blockers from me!

PR #624, opened 2026-08-05
feat: conclude manifest-sync experiment

srtaalej:
## Summary

- Removes the `manifest-sync` experiment flag and enables two-way manifest sync by default
- Unhides `slack manifest sync` command
- Replaces the old confirm prompt flow in `shouldUpdateManifest` with a call to `manifest.Sync`
- Updates experiment docs

## Test plan

- [x] `make lint` passes
- [x] `make test` passes
- [ ] Manual test: `slack manifest sync` works without `--experiment manifest-sync`
- [ ] Manual test: `slack deploy` triggers sync on manifest mismatch


internal/pkg/apps/install_test.go
```diff
@@ -571,17 +545,14 @@ func TestInstall(t *testing.T) {
...
-					nil,
-				)
+			originalSyncFunc := manifestSyncFunc
+			manifestSyncFunc = func(_ context.Context, _ *shared.ClientFactory, _ types.App, _ types.SlackAuth) (*manifestpkg.SyncResult, error) {
+				if tc.mockManifestSyncError != nil {
+					return nil, tc.mockManifestSyncError
+				}
+				return &manifestpkg.SyncResult{}, nil
 			}
+			t.Cleanup(func() { manifestSyncFunc = originalSyncFunc })
```
zimeg, line 555: 🧪 thought: Significant logic is mocked with this pattern and I think we should avoid it to instead **keep** the `manifestpkg.Sync` function call inlined. Confidence in installations is meaningful to keep high!

🔬 ramble: I've opened an adjacent WIP PR https://github.com/slackapi/slack-cli/pull/626/ to collapse `Install` and `InstallLocalApp` functions into one because most logic is identical. I hope it improves the testing ergonomics here.

zimeg (changes requested):
@srtaalej Thanks for getting this started. I'm requesting changes now because we still have a few edges to polish before this seems stable IMHO:

- Recommending a CLI-managed `manifest.json` file as the source of truth. We should remove `get-manifest` from our own hook implementations as part of this. I'm not meaning to deprecate it altogether, but change our preferred default. A frustrating experience appears often otherwise when remote changes aren't saved to file because of a custom hook implementation.

```
 ! No manifest.json found in project root — merged manifest was not written locally
```

- Missing flags in commands calling `manifest sync`. The "run" command shows this and makes me want to revisit the option of https://github.com/slackapi/slack-cli/pull/617 while still in experiment to use perhaps:

```
--manifest-source=project|remote
```

But "local" might be a more common term at the moment. Here's the error for reference:

```
$ lack run --app=A0BP2KFNZSQ | cat

App Manifest
   Found 1 difference(s) between project and app settings


  features.app_home.messages_tab_enabled
    Project:      false
    App settings: true

Check /Users/eden.zimbelman/.slack/logs/slack-debug-20260808.log for error logs
The app manifest was not updated (app_manifest_update_error)

Suggestion
   Run `lack manifest sync` interactively to resolve manifest differences, or pass --force to push the project manifest to app settings or --force-remote to pull app settings to project

$ lack run --app=A0BP2KFNZSQ --force-remote | cat
Check /Users/eden.zimbelman/.slack/logs/slack-debug-20260808.log for error logs
unknown flag: --force-remote
```

zimeg:
🗣️ Changelog for once it's stable.

zimeg (approved):
🌠 @srtaalej Super duper appreciate getting this to a stable spot! Let's merge this for next release.

PR #629, opened 2026-08-10
feat: fall back to manifest.json when get-manifest hook is unavailable

### LINKED ISSUE #627: Default to manifest.json over get-manifest hook in SDK templates
When manifest sync resolves differences and tries to write the merged result locally, projects relying on a `get-manifest` hook instead of a static `manifest.json` file hit:

```
! No manifest.json found in project root — merged manifest was not written locally
```

We should update our SDK hook implementations to prefer `manifest.json` as the default source of truth.

Context: https://github.com/slackapi/slack-cli/pull/624#pullrequestreview-4889752193

srtaalej:
## Summary

- `GetManifestLocal` now falls back to reading `manifest.json` directly from the project root when no `get-manifest` hook is available
- Renamed `readManifestFile` to `getManifestFromFile` to match the `getManifestFromHook` naming convention

Closes #627

## Context

Projects without a `get-manifest` hook but with a static `manifest.json` had no way to read the local manifest. This adds `manifest.json` as a fallback source when the hook isn't defined, allowing the CLI to manage the manifest directly for projects that don't use the hook.

The hook remains the preferred source when available — a hook means the SDK manages the manifest, while a bare `manifest.json` means it's CLI-managed.

## Test plan

- [x] `make test` passes
- [x] Existing tests updated to verify hook-first, file-fallback behavior

### Manual testing

Create a Bolt JS app using `slack create` with the `bolt-js-starter-template`:

```bash
slack create my-test-app --template bolt-js-starter-template
cd my-test-app
```

**Test 1: Hook path still works (existing behavior)**

```bash
slack manifest info --skip-update
```

Should display the manifest via the `get-manifest` hook.

**Test 2: Fallback to manifest.json when no hook is defined**

```bash
# Replace hooks with an empty definition
echo '{"hooks":{}}' > .slack/hooks.json

# CLI should fall back to reading manifest.json directly
slack manifest info --skip-update

# Restore
git checkout .slack/hooks.json
```

Should display the manifest read from `manifest.json`.


internal/app/manifest.go
```diff
@@ -59,17 +64,44 @@ func SetManifestEnvTeamVars(manifestEnv map[string]string, appTeamDomain string,
...
 
-// GetManifestLocal gathers manifest content from the "get-manifest" hook
+// GetManifestLocal reads the local manifest, preferring a static manifest.json
+// file in the project root. Falls back to the "get-manifest" hook when no file exists.
 func (c *ManifestClient) GetManifestLocal(ctx context.Context, sdkConfig hooks.SDKCLIConfig, hookExecutor hooks.HookExecutor) (types.SlackYaml, error) {
+	manifestPath := filepath.Join(sdkConfig.WorkingDirectory, manifestFileName)
+	if exists, _ := afero.Exists(c.fs, manifestPath); exists {
+		return c.readManifestFile(manifestPath)
+	}
+	return c.getManifestFromHook(ctx, sdkConfig, hookExecutor)
```
zimeg: 🔭 thought: We should reverse this order to check if a `get-manifest` hook exists and fallback to reading "manifest.json" file. Deprecating the hook might happen at the **hook** package while the CLI continues to support it.

zimeg (changes requested):
@srtaalej Super appreciate the changes toward this 🌟 

I left a comment on precedence of manifestations and am thinking that the fallback `manifest.json` file allows us to consider the manifest "CLI-managed" while with a hook ought mean not.

This pattern has a slow rollout with hooks needing to be updated too, but I'm most optimistic for the changes it'll unlock with confidence in changing a project manifest.

internal/manifest/writeback.go
```diff
@@ -45,9 +45,14 @@ func WriteManifestLocal(fs afero.Fs, workingDir string, manifest types.AppManife
...
-		}, nil
+		fresh, err := marshalFresh(manifest)
+		if err != nil {
+			return WriteBackResult{}, fmt.Errorf("failed to serialize merged manifest: %w", err)
+		}
+		if err := atomicWriteFile(fs, manifestPath, fresh, 0644); err != nil {
+			return WriteBackResult{}, fmt.Errorf("failed to write %s: %w", manifestFileName, err)
+		}
+		return WriteBackResult{Written: true, FilePath: manifestPath}, nil
 	}
```
zimeg, line 51: 🦠 suggestion: Let's keep this change for a separate PR!

internal/app/manifest.go
```diff
@@ -59,17 +64,44 @@ func SetManifestEnvTeamVars(manifestEnv map[string]string, appTeamDomain string,
...
+// when available. Falls back to reading manifest.json directly from the project root.
 func (c *ManifestClient) GetManifestLocal(ctx context.Context, sdkConfig hooks.SDKCLIConfig, hookExecutor hooks.HookExecutor) (types.SlackYaml, error) {
+	if sdkConfig.Hooks.GetManifest.IsAvailable() {
+		return c.getManifestFromHook(ctx, sdkConfig, hookExecutor)
+	}
+	manifestPath := filepath.Join(sdkConfig.WorkingDirectory, manifestFileName)
+	return c.readManifestFile(manifestPath)
+}
+
+func (c *ManifestClient) readManifestFile(path string) (types.SlackYaml, error) {
```
zimeg:
```suggestion
func (c *ManifestClient) getManifestFromFile(sdkConfig hooks.SDKCLIConfig) (types.SlackYaml, error) {
```

🌵 thought: I think we should match the function name convention here and also keep path specific logic contained within this function. I'm less confident of the second thought FWIW!

zimeg (changes requested):
🔍 Before testing more I want to cut scope on this change. I think other PRs can follow to ensure we're not clobbering an SDK managed manifest!

internal/app/manifest.go
```diff
@@ -59,17 +64,44 @@ func SetManifestEnvTeamVars(manifestEnv map[string]string, appTeamDomain string,
...
+			WithRootCause(err).
+			WithCode(slackerror.ErrInvalidManifest)
+	}
+	return sl, nil
+}
+
+func (c *ManifestClient) getManifestFromHook(ctx context.Context, sdkConfig hooks.SDKCLIConfig, hookExecutor hooks.HookExecutor) (types.SlackYaml, error) {
 	var sl types.SlackYaml
 
 	if !sdkConfig.Hooks.GetManifest.IsAvailable() {
```
mwbrooks: issue: This guard is now unreachable. GetManifestLocal only routes here when sdkConfig.Hooks.GetManifest.IsAvailable() is true, so the ErrSDKHookNotFound branch can never fire and the diagnostic it used to produce is lost. A project with no get-manifest hook and no manifest.json now reports "Failed to read manifest file" with ErrInvalidManifest, which points at a malformed manifest rather than a missing hook. Either delete the dead guard and keep the file path honest, or have getManifestFromFile distinguish "no manifest.json and no hook" from "manifest.json is invalid".

internal/manifest/sync_test.go
```diff
@@ -173,52 +173,45 @@ func Test_Sync(t *testing.T) {
...
+		t.Run(name, func(t *testing.T) {
+			f := newSyncTestFixture(t)
+			f.projectConfig.On("GetManifestSource", mock.Anything).Return(config.ManifestSourceLocal, nil)
+			f.manifestMock.On("GetManifestLocal", mock.Anything, mock.Anything, mock.Anything).
+				Return(localManifest, nil)
+			f.manifestMock.On("GetManifestRemote", mock.Anything, mock.Anything, mock.Anything).
+				Return(remoteManifest, nil)
+			f.clients.Config.ForceFlag = tc.forceFlag
+			f.clients.Config.ForceRemoteFlag = tc.forceRemoteFlag
+			f.clientsMock.API.On("UpdateApp", mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything).
```
mwbrooks: issue: The table-driven rewrite dropped the argument assertion. The previous test pinned the call with AssertCalled(t, "UpdateApp", mock.Anything, "xoxb-test", "A123", mock.Anything, true, true); this version matches every argument with mock.Anything and only checks the merged description. The force and force-remote paths are precisely where the wrong token, app ID, or boolean pair would be the bug, so the coverage that mattered went away even though the case count went up. Keep a per-case arg assertion.

internal/app/manifest.go
```diff
@@ -59,23 +64,45 @@ func SetManifestEnvTeamVars(manifestEnv map[string]string, appTeamDomain string,
...
-		return sl, slackerror.New(slackerror.ErrSDKHookNotFound).
-			WithMessage("The `get-manifest` script was not found")
+func (c *ManifestClient) getManifestFromFile(sdkConfig hooks.SDKCLIConfig) (types.SlackYaml, error) {
+	var sl types.SlackYaml
+	manifestPath := filepath.Join(sdkConfig.WorkingDirectory, manifestFileName)
+	data, err := afero.ReadFile(c.fs, manifestPath)
+	if err != nil {
+		return sl, slackerror.New("Failed to read manifest file").
+			WithRootCause(err).
+			WithCode(slackerror.ErrNoFile)
```
mwbrooks: issue: Removing the dead guard was the right call, but `ErrNoFile` isn't the code to land on. In `internal/slackerror/errors.go` it's the code-bundle upload error (message `Couldn't upload your bundled code to server`, remediation `Please try again`), and it has no other production use in the repo.

There's a subtlety worth knowing here, because it makes the code look correct when it isn't: `WithCode` only assigns the code field, and `Error()` backfills the registry's message and remediation **only when `Message` is empty** (`internal/slackerror/error.go:84-90`). Since `New("Failed to read manifest file")` already set a message, the registry text is discarded. So what a user actually sees for a project with no `get-manifest` hook and no `manifest.json` is:

```
Failed to read manifest file (no_file)
   open /project/manifest.json: file does not exist
```

No `Suggestion:` block, and a machine-readable code that means "upload failed". Before this PR the same situation produced `The `get-manifest` script was not found (sdk_hook_not_found)` plus remediation pointing at `slack.json` / `.slack/hooks.json`. That's the regression I was worried about in the last pass, and it's still here, just relocated.

`cmd/platform/deploy.go:214` (`errorMissingDeployHook`) is the existing idiom for this exact shape:

```go
return sl, slackerror.New(slackerror.ErrSDKHookNotFound).
    WithMessage("No `get-manifest` hook and no %s found in the project", manifestFileName).
    WithRemediation(...)  // point at the hooks docs and at creating manifest.json
```

Keeping `ErrSDKHookNotFound` also preserves the code contract for anything parsing `--json` output. If you'd rather have a distinct code, adding a new one is fine; reusing `no_file` isn't.

internal/manifest/sync_test.go
mwbrooks: issue: Still open from the last pass, and I don't think it was intentional: this file is untouched by the new commit, so the `UpdateApp` argument assertion is still gone. `grep AssertCalled` over the whole file returns nothing, and the consolidated loop sets up six `mock.Anything` and then asserts only `HasDifferences`, `WriteBack.Written`, and the merged description.

What it used to assert:

```go
f.clientsMock.API.AssertCalled(t, "UpdateApp", mock.Anything, "xoxb-test", "A123", mock.Anything, true, true)
```

Those last two `true`s are the force and force-remote flags, which is the behaviour these cases exist to pin down. Without them the test passes whether or not we pass the flags through correctly. Worth restoring on at least the force / force-remote cases.

internal/app/manifest.go
```diff
@@ -59,23 +64,45 @@ func SetManifestEnvTeamVars(manifestEnv map[string]string, appTeamDomain string,
...
 	}
+	if err := json.Unmarshal(data, &sl); err != nil {
+		return sl, slackerror.New("Failed to parse manifest file").
+			WithRootCause(err).
+			WithCode(slackerror.ErrInvalidManifest)
+	}
+	return sl, nil
+}
+
+func (c *ManifestClient) getManifestFromHook(ctx context.Context, sdkConfig hooks.SDKCLIConfig, hookExecutor hooks.HookExecutor) (types.SlackYaml, error) {
```
mwbrooks: issue: The split makes an existing asymmetry easier to see, and I think it's worth closing while you're in here. `getManifestFromFile` wraps its unmarshal failure as `ErrInvalidManifest` (line 99), but the hook path ends with a bare `err = json.Unmarshal(...); return sl, err` and returns the raw `*json.SyntaxError`. So a hook that emits output starting with `{` but not valid JSON produces an error with no code and no CLI framing, and any caller doing a `*slackerror.Error` type assertion on it panics rather than fails. Wrapping it the same way line 99 does is a one-liner. There's also no `hookTests` case for that shape (the existing one has no brace at all, which takes the other branch).

internal/app/manifest_test.go
```diff
@@ -68,77 +69,128 @@ func Test_AppManifest_SetManifestEnvTeamVars(t *testing.T) {
...
+			}
+
 			mockHookExecutor := &hooks.MockHookExecutor{}
-			if tc.mockManifestInfo != "" {
-				mockSDKConfig.Hooks.GetManifest = hooks.HookScript{
-					Name:    "GetManifest",
-					Command: "cat manifest.json",
-				}
+			if tc.hookCommand != "" {
 				mockHookExecutor.On("Execute", mock.Anything, mock.Anything).
```
mwbrooks: suggestion: The old version of this test set `configMock.ManifestEnv = map[string]string{"EXAMPLE": "12"}`, and `ManifestEnv` no longer appears anywhere in this file. That leaves the `for name, val := range c.Env { manifestHookOpts.Env[name] = val }` loop in `getManifestFromHook` with no test exercising it, which is the same shape of gap as the `UpdateApp` one. `mock.MatchedBy` on the `HookExecOpts` argument would let you assert the source arg, `DENO_AUTH_TOKENS`, and the merged env in one place instead of `mock.Anything`.

internal/app/manifest_test.go
```diff
@@ -68,77 +69,128 @@ func Test_AppManifest_SetManifestEnvTeamVars(t *testing.T) {
...
+			manifestClient := NewManifestClient(&api.APIMock{}, configMock, fsMock)
+			result, err := manifestClient.GetManifestLocal(ctx, mockSDKConfig, mockHookExecutor)
+
+			if tc.expectedErrCode != "" {
+				require.Error(t, err)
+				assert.Equal(t, tc.expectedErrCode, err.(*slackerror.Error).Code)
 			} else {
-				mockSDKConfig.Hooks.GetManifest = hooks.HookScript{Name: "GetManifest"}
+				require.NoError(t, err)
+				assert.Equal(t, tc.expectedName, result.DisplayInformation.Name)
```
mwbrooks: suggestion: This compares only `DisplayInformation.Name` where the old test compared the whole `types.SlackYaml`. Cheap to restore struct equality on at least the happy path, and it catches the class of bug where the fallback reads the right file but drops fields.

internal/app/manifest_test.go
```diff
@@ -68,77 +69,128 @@ func Test_AppManifest_SetManifestEnvTeamVars(t *testing.T) {
...
-		mockManifestErr  error
-		expectedErr      error
-		expectedManifest types.SlackYaml
+	fallbackTests := map[string]struct {
+		hookCommand     string
+		hookOutput      string
+		manifestFile    string
+		expectedName    string
+		expectedErrCode string
+		expectHookCall  bool
```
mwbrooks: suggestion: `expectHookCall` is derivable from `hookCommand != ""`, which the loop below already branches on twice. Two sources of truth for the same fact can disagree: add a hook case, forget the flag, and you get an `AssertNotCalled` failure that reads like a production bug. Dropping the field makes the table shorter and harder to get wrong.

internal/app/manifest.go
```diff
@@ -59,23 +64,45 @@ func SetManifestEnvTeamVars(manifestEnv map[string]string, appTeamDomain string,
...
+	}
+	return c.getManifestFromFile(sdkConfig)
+}
 
-	if !sdkConfig.Hooks.GetManifest.IsAvailable() {
-		return sl, slackerror.New(slackerror.ErrSDKHookNotFound).
-			WithMessage("The `get-manifest` script was not found")
+func (c *ManifestClient) getManifestFromFile(sdkConfig hooks.SDKCLIConfig) (types.SlackYaml, error) {
+	var sl types.SlackYaml
+	manifestPath := filepath.Join(sdkConfig.WorkingDirectory, manifestFileName)
```
mwbrooks: issue: Guard against an empty `WorkingDirectory` here. It is only ever set at `internal/shared/clients.go:239` after `InitSDKConfig` locates a `hooks.json`, and when that fails it stays `""`, which is the same sentinel `SDKCLIConfig.Exists()` reads as "not a project directory". `filepath.Join("", "manifest.json")` gives you the relative path `manifest.json`, and afero's OsFs resolves that against the process CWD, so `slack env list` in an unrelated folder could pick up a stray manifest. An early `if _, err := sdkConfig.Exists(); err != nil { return sl, err }` covers it.

internal/app/manifest.go
```diff
@@ -59,23 +64,45 @@ func SetManifestEnvTeamVars(manifestEnv map[string]string, appTeamDomain string,
...
 	}
 	return client
 }
 
-// GetManifestLocal gathers manifest content from the "get-manifest" hook
+// GetManifestLocal reads the local manifest, preferring the "get-manifest" hook
+// when available. Falls back to reading manifest.json from the project root.
 func (c *ManifestClient) GetManifestLocal(ctx context.Context, sdkConfig hooks.SDKCLIConfig, hookExecutor hooks.HookExecutor) (types.SlackYaml, error) {
-	var sl types.SlackYaml
+	if sdkConfig.Hooks.GetManifest.IsAvailable() {
```
mwbrooks: note: This quietly retires the last production source of `ErrSDKHookNotFound` from the manifest path, and `cmd/root.go:399` was upgrading that code to `ErrRuntimeNotFound` whenever `SDKConfig.Runtime` was empty. A project with a `hooks.json`, no `get-manifest` hook, no runtime, and no `manifest.json` used to get runtime-install guidance and now gets a bare read failure. Worth confirming that trade is intentional, and note that `cmd/env/init_test.go` and the `cmd/datastore/*_test.go` files still mock this call returning `ErrSDKHookNotFound`, so those cases now exercise an error the real client cannot produce.

internal/app/manifest_test.go
```diff
@@ -68,77 +69,128 @@ func Test_AppManifest_SetManifestEnvTeamVars(t *testing.T) {
...
+
+			if tc.expectHookCall {
+				mockHookExecutor.AssertCalled(t, "Execute", mock.Anything, mock.Anything)
+			} else {
+				mockHookExecutor.AssertNotCalled(t, "Execute", mock.Anything, mock.Anything)
 			}
+		})
+	}
+
+	hookTests := map[string]struct {
```
mwbrooks: suggestion: Add a case where the hook is available, the hook errors, and a valid `manifest.json` sits on disk, asserting the hook error propagates. Right now precedence is only proven on the happy path, and the invariant that matters most is that a failing hook never silently falls through to a possibly stale file in an SDK-managed project. The behaviour is correct today, but nothing in the suite would catch a future refactor that added a fallback on error.

internal/app/manifest.go
mwbrooks: issue: `getManifestFromFile` joins `sdkConfig.WorkingDirectory` with no emptiness guard. `WorkingDirectory` is set in exactly one place (`internal/shared/clients.go:239`), only after `InitSDKConfig` walks up and finds a `hooks.json`. When that lookup fails it stays `""`, which is the same sentinel `SDKCLIConfig.Exists()` uses for "not a project" (`internal/hooks/sdk_config.go:51`). `filepath.Join("", "manifest.json")` resolves to the relative path `manifest.json`, which afero's OsFs resolves against the process CWD, not the project root. `cmd/env/env.go:88` (`isHostedRuntime`) calls `GetManifestLocal` and swallows the error without gating on a valid project directory, so this can silently read an unrelated `manifest.json` from wherever the user happens to be standing.

internal/app/manifest.go
```diff
@@ -59,23 +64,45 @@ func SetManifestEnvTeamVars(manifestEnv map[string]string, appTeamDomain string,
...
 
-	if !sdkConfig.Hooks.GetManifest.IsAvailable() {
-		return sl, slackerror.New(slackerror.ErrSDKHookNotFound).
-			WithMessage("The `get-manifest` script was not found")
+func (c *ManifestClient) getManifestFromFile(sdkConfig hooks.SDKCLIConfig) (types.SlackYaml, error) {
+	var sl types.SlackYaml
+	manifestPath := filepath.Join(sdkConfig.WorkingDirectory, manifestFileName)
+	data, err := afero.ReadFile(c.fs, manifestPath)
+	if err != nil {
+		return sl, slackerror.New("Failed to read manifest file").
```
mwbrooks: issue: The missing-file error is coded `ErrNoFile`, whose canonical definition (`internal/slackerror/errors.go:1267`) is "Couldn't upload your bundled code to server" / "Please try again". That's the only non-test production use of that code in the repo, and it doesn't fit a manifest read failure. Because `slackerror.New("Failed to read manifest file")` doesn't match a known error, `getErrorIfKnown` misses and builds a bare `Error` carrying only that message; `WithCode` sets `Code` but never `Remediation`. Net effect: a message with no path, no remediation, and a `no_file` code shipped to telemetry via `cmd/root.go:411` under a bundle-upload meaning, which will muddy any dashboard filtering on that code.

internal/app/manifest.go
mwbrooks: note: Hook-wins ordering is correct, including on the error path. Good design.

internal/app/manifest.go
```diff
@@ -59,23 +64,45 @@ func SetManifestEnvTeamVars(manifestEnv map[string]string, appTeamDomain string,
...
 
-// GetManifestLocal gathers manifest content from the "get-manifest" hook
+// GetManifestLocal reads the local manifest, preferring the "get-manifest" hook
+// when available. Falls back to reading manifest.json from the project root.
 func (c *ManifestClient) GetManifestLocal(ctx context.Context, sdkConfig hooks.SDKCLIConfig, hookExecutor hooks.HookExecutor) (types.SlackYaml, error) {
-	var sl types.SlackYaml
+	if sdkConfig.Hooks.GetManifest.IsAvailable() {
+		return c.getManifestFromHook(ctx, sdkConfig, hookExecutor)
+	}
+	return c.getManifestFromFile(sdkConfig)
```
mwbrooks: question: Dropping the `ErrSDKHookNotFound` return here is a behaviour change wider than the stated fallback. `cmd/root.go:399` upgrades `ErrSDKHookNotFound` to `ErrRuntimeNotFound` when `clients.SDKConfig.Runtime == ""`, so a project with a `hooks.json` but no `get-manifest` hook and no runtime previously surfaced runtime-install guidance. That mapping can no longer fire from this path. Several command tests still mock `GetManifestLocal` returning `ErrSDKHookNotFound` (`cmd/env/init_test.go:72,88,104,122`, the nine `cmd/datastore/*_test.go` files, `cmd/app/uninstall_test.go:65`), so they now assert against an error the real client can't produce. Was this intentional, or should the runtime-not-found path be preserved alongside the fallback?

internal/app/manifest_test.go
mwbrooks: suggestion: Hook precedence is only tested on the success path. Add a case with the hook available, the hook erroring, and a valid `manifest.json` present, asserting the error propagates instead of silently falling back. Correct today because `GetManifestLocal` returns `getManifestFromHook`'s result directly, but that unguarded invariant is exactly what a future refactor would break.

internal/manifest/sync_test.go
```diff
@@ -173,52 +173,45 @@ func Test_Sync(t *testing.T) {
...
-		assert.Equal(t, "Remote", result.Merged.DisplayInformation.Description)
-	})
+	mergeStrategyTests := map[string]struct {
+		forceFlag       bool
+		forceRemoteFlag bool
+		expectedDesc    string
+	}{
+		"force flag merges all local": {
+			forceFlag:    true,
+			expectedDesc: "Local",
```
mwbrooks: note: This force/force-remote table refactor drops the `AssertCalled(t, "UpdateApp", mock.Anything, "xoxb-test", "A123", mock.Anything, true, true)` check from the force-flag case, so the `UpdateApp` argument shape is no longer verified anywhere here. It gains a `Description == "Local"` assertion in exchange, a fair trade but not equivalent, and this whole file's changes are unrelated to the PR's stated purpose.

internal/app/manifest_test.go
```diff
@@ -68,77 +69,128 @@ func Test_AppManifest_SetManifestEnvTeamVars(t *testing.T) {
...
+			result, err := manifestClient.GetManifestLocal(ctx, mockSDKConfig, mockHookExecutor)
+			if tc.expectedErr != "" {
 				require.Error(t, err)
-				assert.Equal(t,
-					tc.expectedErr.(*slackerror.Error).Code, err.(*slackerror.Error).Code)
+				assert.Equal(t, tc.expectedErr, err.(*slackerror.Error).Code)
 			} else {
 				require.NoError(t, err)
-				assert.Equal(t, tc.expectedManifest, actualManifest)
+				assert.Equal(t, tc.expectedName, result.DisplayInformation.Name)
```
zimeg: 🧪 suggestion: Might we keep the entire manifest assertion here? These unit tests are capturing a meaningful interface to build confidence to IMHO.

internal/app/manifest_test.go
```diff
@@ -68,77 +69,128 @@ func Test_AppManifest_SetManifestEnvTeamVars(t *testing.T) {
...
 			}
+		})
+	}
+
+	hookTests := map[string]struct {
+		hookOutput   string
+		hookErr      error
+		expectedName string
+		expectedErr  string
+	}{
```
zimeg: 🧪 suggestion: Let's combine these test cases with the ones above for this function. This'll help us extend these later if needed with a solid foundation.

internal/manifest/sync_test.go
```diff
@@ -173,52 +173,45 @@ func Test_Sync(t *testing.T) {
...
-		result, err := Sync(f.ctx, f.clients, testApp, testAuth)
-
-		require.NoError(t, err)
-		require.NotNil(t, result)
-		assert.True(t, result.HasDifferences)
-		assert.True(t, result.WriteBack.Written)
-		// Verify remote value was used — the merged manifest should have "Remote" description
-		assert.Equal(t, "Remote", result.Merged.DisplayInformation.Description)
-	})
+	mergeStrategyTests := map[string]struct {
```
zimeg, line 176: 👁️‍🗨️ question: Were these test expectations changed? I'm noticing we don't assert a call to the `UpdateApp` function now which concerns me somewhat since logic adjacent is being changed in this PR too.

zimeg (commented):
@srtaalej Nice I think this'll be a big unlock before the experiment concludes! I'm leaving a few comments but logic LGTM.

- 🧪 Tests can perhaps be scoped more narrow to the hook and fallback changes? I'm interested in this so we keep confidence in both what's changing and what's not.
- 🔭 The `sync` command might have follow up to require a manifest hook doesn't exist. IMHO this is important so we're confident in writing back to expected JSON instead of whatever a hook might gather from.
- 🪝 We should discuss a plan to remove `get-manifest` from the Bolt hook implementations. I forget if we can confirm current CLI versions as part of hooks but we might prefer the plain JSON for apps going forward.

If test changes seem alright and so do discussion points above let me know for a final test and review! 🪬 ✨

internal/app/manifest.go
mwbrooks: issue: This removes the only place the CLI returns `slackerror.ErrSDKHookNotFound`, and `cmd/root.go` still special-cases it: `slackerror.Is(err, slackerror.ErrSDKHookNotFound) && clients.SDKConfig.Runtime == ""`. After this change a hookless project with no `manifest.json` surfaces `ErrNoFile` instead, so that handler no longer fires and the guidance it prints is unreachable. `cmd/env/init_test.go` also mocks `GetManifestLocal` returning that code twice, so the suite asserts behaviour the production path can no longer produce. Worth either keeping the hook-missing code on the no-hook branch or updating `cmd/root.go` and those tests in the same PR.

internal/app/manifest.go
```diff
@@ -17,18 +17,23 @@ package app
...
 
 	"github.com/slackapi/slack-cli/internal/api"
 	"github.com/slackapi/slack-cli/internal/config"
 	"github.com/slackapi/slack-cli/internal/hooks"
 	"github.com/slackapi/slack-cli/internal/shared/types"
 	"github.com/slackapi/slack-cli/internal/slackerror"
+	"github.com/spf13/afero"
 )
 
+const manifestFileName = "manifest.json"
```
mwbrooks: suggestion: `internal/manifest/writeback.go` already declares an identical unexported `const manifestFileName = "manifest.json"` for the same file this code reads. Two private constants naming one on-disk artifact will drift, so consider promoting one to a shared exported const rather than adding a second.

internal/manifest/sync_test.go
srtaalej: i mustve missed that when i refactored the tests! thanks for catching, i brought it back in commit [5293b38](https://github.com/slackapi/slack-cli/pull/629/commits/5293b38772901ee80a61f6bca4556b00f92b03b4) ✨

zimeg (approved):
@srtaalej 🏆 Super appreciated changes! Thanks for keeping tests focused here. It's a significant update to what we can recommend going forward and I'm feeling confident about these changes so please merge when convenient 🍾 🚢 💨

internal/manifest/sync_test.go
zimeg: @srtaalej Ah I was unsure if `sync` needed changes as part of the hook fallback! Thanks for sharing update!

PR #630, opened 2026-08-10
feat: add --manifest-source flag to run and deploy commands

### LINKED ISSUE #628: Add --force/--force-remote flags to run and deploy commands
When `manifest sync` is triggered internally during `slack run` or `slack deploy` and finds differences, the error suggests using `--force` or `--force-remote` — but those flags only exist on `manifest sync`, not on `run`/`deploy`.

```
$ slack run --app=A0BP2KFNZSQ --force-remote | cat
unknown flag: --force-remote
```

We should either add those flags to commands that call sync internally, or introduce a `--manifest-source=project|remote` flag.

Context: https://github.com/slackapi/slack-cli/pull/624#pullrequestreview-4889752193

srtaalej:
## Summary

- Adds a `--manifest-source=project|remote` flag to `slack run` and `slack deploy` commands
- When manifest sync detects differences during run/deploy, this flag allows non-interactive resolution without requiring `slack manifest sync` to be run separately
- Skips the "Overwrite manifest on app settings?" confirmation prompt during install when `--manifest-source` is set — `project` auto-approves the overwrite, `remote` skips it entirely
- Updates the non-TTY error remediation to reference `--manifest-source` instead of `--force`/`--force-remote` (which are only available on `manifest sync`)

Closes #628

## Test plan

- [x] `make lint` passes
- [x] `make test` passes
- [ ] Manual test: `slack run --manifest-source=project` pushes local manifest without prompting
- [ ] Manual test: `slack run --manifest-source=remote` pulls app settings without prompting
- [ ] Manual test: `slack deploy --manifest-source=project` works in non-TTY (CI) environments
- [ ] Manual test: `slack run --manifest-source=invalid` returns a clear validation error
- [ ] Manual test: `slack run --manifest-source=remote` skips the "Overwrite manifest?" prompt on reinstall


internal/manifest/sync.go
```diff
@@ -91,8 +91,8 @@ func Sync(ctx context.Context, clients *shared.ClientFactory, app types.App, aut
 		return nil, slackerror.New(slackerror.ErrAppManifestUpdate).
 			WithRemediation("Run %s interactively to resolve manifest differences, or pass %s to push the project manifest to app settings or %s to pull app settings to project",
 				style.Commandf("manifest sync", false),
-				style.CommandText("--force"),
-				style.CommandText("--force-remote"),
+				style.CommandText("--manifest-source=project"),
```
mwbrooks: issue: This remediation names a flag the command it fires from does not accept. --manifest-source is registered on platform deploy and platform run only; cmd/manifest/sync.go still registers just --force and --force-remote, and its own examples read `manifest sync --force`. So a user who hits this error from manifest sync and copies the suggestion gets "unknown flag: --manifest-source". Either register the flag on manifest sync too (and update its examples), or keep this remediation on --force / --force-remote.

internal/manifest/sync.go
```diff
@@ -77,12 +77,12 @@ func Sync(ctx context.Context, clients *shared.ClientFactory, app types.App, aut
 
 	var merged types.AppManifest
 	switch {
-	case clients.Config.ForceFlag:
+	case clients.Config.ManifestSourceFlag == "project" || clients.Config.ForceFlag:
```
mwbrooks: issue: Two problems stacked here. The switch compares raw "project" / "remote" literals while flags.go defines ManifestSourceProject and ManifestSourceRemote for exactly these values, so the constants are declared and then bypassed. And because manifest sync never registers --manifest-source, ManifestSourceFlag is always "" on that path, which makes the first clause of each case dead there and leaves ForceFlag doing all the work. The code reads as though manifest sync honours --manifest-source when it cannot.

internal/cmdutil/flags.go
```diff
@@ -35,6 +37,29 @@ var OrgGrantWorkspaceDescription = func() string {
...
 }
 
+// ManifestSourceFlag values
+const (
+	ManifestSourceProject = "project"
+	ManifestSourceRemote  = "remote"
+)
+
+// ValidateManifestSourceFlag checks that --manifest-source has a valid value if set
+func ValidateManifestSourceFlag(clients *shared.ClientFactory) error {
```
mwbrooks: praise: Taking the whole ClientFactory rather than a bare string keeps the validator callable from PreRunE without threading config through, and wiring it into both platform deploy and platform run means an invalid value fails before any network call.

internal/manifest/sync.go
```diff
@@ -91,8 +92,8 @@ func Sync(ctx context.Context, clients *shared.ClientFactory, app types.App, aut
 		return nil, slackerror.New(slackerror.ErrAppManifestUpdate).
 			WithRemediation("Run %s interactively to resolve manifest differences, or pass %s to push the project manifest to app settings or %s to pull app settings to project",
 				style.Commandf("manifest sync", false),
-				style.CommandText("--force"),
-				style.CommandText("--force-remote"),
+				style.CommandText("--manifest-source=project / --force"),
+				style.CommandText("--manifest-source=remote / --force-remote"),
```
zimeg: 🪓 question: Are we alright to replace the `--force` and `--force-remote` options altogether while the `sync` command is under experiment?

internal/config/config.go
```diff
@@ -54,6 +54,7 @@ type Config struct {
 	DisableTelemetryFlag    bool
 	ForceFlag               bool
 	ForceRemoteFlag         bool
+	ManifestSourceFlag      string
 	LogstashHostResolved    string
```
zimeg: 🧮 suggestion: Let's keep this in alphabetical order!

internal/cmdutil/flags.go
```diff
@@ -35,6 +37,29 @@ var OrgGrantWorkspaceDescription = func() string {
 		style.Secondary("(or 'all' for all workspaces in the org)"))
 }
 
+// ManifestSourceFlag values
+const (
+	ManifestSourceProject = "project"
+	ManifestSourceRemote  = "remote"
+)
```
zimeg:
```suggestion
const (
	ManifestSourceProject = "local"
	ManifestSourceRemote  = "remote"
)
```

🪬 suggestion(blocking): Earlier suggestion might've hinted at "project" terms but we should match existing configuration options I realize. Perhaps reusing logic from this package instead of validations here?

https://github.com/slackapi/slack-cli/blob/73dde37cd9abc011b7058f363ca1fcba7f87b878/internal/config/manifest.go#L20-L23

cmd/platform/deploy.go
```diff
@@ -59,6 +59,9 @@ func NewDeployCommand(clients *shared.ClientFactory) *cobra.Command {
 			{Command: "platform deploy --team T0123456", Meaning: "Deploy to a specific team"},
 		}),
 		PreRunE: func(cmd *cobra.Command, args []string) error {
+			if err := cmdutil.ValidateManifestSourceFlag(clients); err != nil {
+				return err
+			}
```
zimeg, line 64:
```suggestion
```

🪓 quibble: I'd favor this validation happening with the `switch` case in `internal/manifest/sync.go` to avoid duplicate checks in code, although I understand this might error earlier.

zimeg (changes requested):
@srtaalej I appreciate the most patient request for review 🙏 ✨ I'm requesting a few changes with hopes these comments move this in evermore stable directions. I call out:

- Removing `--force-remote` flag altogether alongside this change: I'd like to avoid multiple options unless follow up is planned to remove this too?
- Favoring the existing terms and implementations of manifest source: Our configuration file has some logic we might reuse here!

If I can share more to these please let me know! I'm optimistic we include this in upcoming release 🚀 🔮

internal/cmdutil/flags.go
```diff
@@ -35,6 +37,29 @@ var OrgGrantWorkspaceDescription = func() string {
 		style.Secondary("(or 'all' for all workspaces in the org)"))
 }
 
+// ManifestSourceFlag values
+const (
+	ManifestSourceProject = "project"
```
mwbrooks: praise: Using the shared constants in sync.go's switch closes the exact gap the last review flagged. A future rename of either value now breaks at compile time instead of silently going dead on the manifest sync path.

internal/manifest/sync_test.go
```diff
@@ -220,6 +221,68 @@ func Test_Sync(t *testing.T) {
...
+
+		result, err := Sync(f.ctx, f.clients, testApp, testAuth)
+
+		require.NoError(t, err)
+		require.NotNil(t, result)
+		assert.True(t, result.HasDifferences)
+		assert.Equal(t, "Remote", result.Merged.DisplayInformation.Description)
+	})
+
+	t.Run("non-TTY error mentions --manifest-source in remediation", func(t *testing.T) {
```
mwbrooks: praise: The non-TTY remediation test asserting on all four flag strings (--manifest-source=project/remote plus --force/--force-remote) is the right regression guard for the confusion the last draft caught.

internal/cmdutil/flags.go
```diff
@@ -35,6 +38,23 @@ var OrgGrantWorkspaceDescription = func() string {
 		style.Secondary("(or 'all' for all workspaces in the org)"))
 }
 
+// ValidateManifestSourceFlag checks that --manifest-source has a valid value if set
+func ValidateManifestSourceFlag(clients *shared.ClientFactory) error {
+	v := clients.Config.ManifestSourceFlag
+	if v == "" {
+		return nil
+	}
+	if v != string(config.ManifestSourceLocal) && v != string(config.ManifestSourceRemote) {
```
mwbrooks: note: The PR description advertises `--manifest-source=project|remote`, but the validation here accepts `local|remote` (and `flags_test.go` asserts `"project"` is invalid). Whichever spelling is intended, the description and the flag help should match the code before this lands, since the description is what ends up in the changelog.

internal/manifest/sync.go
```diff
@@ -91,8 +91,8 @@ func Sync(ctx context.Context, clients *shared.ClientFactory, app types.App, aut
 		return nil, slackerror.New(slackerror.ErrAppManifestUpdate).
 			WithRemediation("Run %s interactively to resolve manifest differences, or pass %s to push the project manifest to app settings or %s to pull app settings to project",
 				style.Commandf("manifest sync", false),
-				style.CommandText("--force"),
-				style.CommandText("--force-remote"),
+				style.CommandText("--manifest-source=local / --force"),
```
mwbrooks: suggestion: `style.CommandText("--manifest-source=local / --force")` styles the ` / ` separator as part of the command, so the remediation renders as one unrunnable command string rather than two alternatives. Two separate `style.CommandText` args with a plain ` or ` between them reads correctly when copied out of the terminal.

internal/pkg/apps/install.go
```diff
@@ -701,6 +701,12 @@ func shouldUpdateManifest(ctx context.Context, clients *shared.ClientFactory, ap
 	if err != nil {
 		return false, err
 	}
+	if clients.Config.ManifestSourceFlag == string(config.ManifestSourceRemote) {
+		return false, nil
+	}
+	if clients.Config.ManifestSourceFlag == string(config.ManifestSourceLocal) {
```
mwbrooks: note: The flag short-circuits ahead of the stored `manifestSource`, so `--manifest-source` silently wins over a project's configured source rather than erroring on the conflict. That is probably the intent for a per-invocation override, but it is worth a line in the flag help so a developer whose project is pinned to `remote` is not surprised when a local push succeeds.

internal/manifest/sync.go
srtaalej: yes i think thats a nice consolidation!

internal/cmdutil/flags.go
srtaalej: I agree!

cmd/platform/deploy.go
srtaalej: 🫡

internal/pkg/apps/install.go
mwbrooks: note: `--manifest-source` still short-circuits ahead of the stored `manifestSource` with no mention in the flag help ("resolve manifest differences using this source (local or remote)"). Same finding as the prior draft against 392db1e, still open: worth a line documenting that this overrides a pinned project source.

cmd/platform/deploy.go
srtaalej: we actually do need the check to happen in run and deploy as well otherwise invalid values in the flag will fall through

internal/pkg/apps/install.go
mwbrooks: note: `--manifest-source` still short-circuits ahead of the stored `manifestSource` with no mention in the flag help ("resolve manifest differences using this source (local or remote)"). Same finding carried through two redrafts now: worth a line documenting that this overrides a pinned project source.

internal/config/config.go
```diff
@@ -55,6 +55,7 @@ type Config struct {
 	ForceFlag               bool
 	ForceRemoteFlag         bool
```
zimeg, line 56:
```suggestion
```

🪓 note: I'm eager to replace this as part of these changes while the initial implementation is under experiment still. IMHO a replacement of these instances makes safer review without handling multiple cases.

internal/manifest/sync.go
```diff
@@ -77,12 +86,12 @@ func Sync(ctx context.Context, clients *shared.ClientFactory, app types.App, aut
 
 	var merged types.AppManifest
 	switch {
-	case clients.Config.ForceFlag:
+	case clients.Config.ManifestSourceFlag == string(config.ManifestSourceLocal) || clients.Config.ForceFlag:
```
zimeg:
```suggestion
	case config.ManifestSource(clients.Config.ManifestSourceFlag).Equals(config.ManifestSourceLocal):
```

🪬 suggestion: Am hoping to compare objects without converting to string here.

internal/manifest/sync.go
zimeg, line 99:
```suggestion
    case config.ManifestSource(clients.Config.ManifestSourceFlag).Exists() && !config.ManifestSource(clients.Config.ManifestSourceFlag).IsValid():
        return nil, slackerror.New(slackerror.ErrInvalidFlag).
			WithMessage("Invalid value %q for %s flag", clients.Config.ManifestSourceFlag, style.CommandText("--manifest-source")).
			WithRemediation("Valid values are %s or %s",
				style.Highlight(string(config.ManifestSourceLocal)),
				style.Highlight(string(config.ManifestSourceRemote)),
			)
	case !clients.IO.IsTTY():
```

🍾 suggestion: We might introduce a new "IsValid" method to the manifest source configurations to compare and error here?

internal/manifest/sync.go
```diff
@@ -36,6 +36,15 @@ type SyncResult struct {
...
 func Sync(ctx context.Context, clients *shared.ClientFactory, app types.App, auth types.SlackAuth) (*SyncResult, error) {
+	if v := clients.Config.ManifestSourceFlag; v != "" && !(v == string(config.ManifestSourceLocal) || v == string(config.ManifestSourceRemote)) {
+		return nil, slackerror.New(slackerror.ErrInvalidFlag).
+			WithMessage("Invalid value %q for %s flag", v, style.CommandText("--manifest-source")).
+			WithRemediation("Valid values are %s or %s",
+				style.Highlight(string(config.ManifestSourceLocal)),
+				style.Highlight(string(config.ManifestSourceRemote)),
+			)
+	}
+
```
zimeg, line 47:
```suggestion
```

🪓 suggestion: Am hoping the checks below can cover this validation without causing issue?

🌲 note: I lean towards keeping validation in `cmd` overall which I might also have suggestion these files shouldn't be in internal... But that can be a later change!

internal/cmdutil/flags.go
```diff
@@ -35,6 +38,23 @@ var OrgGrantWorkspaceDescription = func() string {
...
+	if !(v == string(config.ManifestSourceLocal) || v == string(config.ManifestSourceRemote)) {
+		return slackerror.New(slackerror.ErrInvalidFlag).
+			WithMessage("Invalid value %q for %s flag", v, style.CommandText("--manifest-source")).
+			WithRemediation("Valid values are %s or %s",
+				style.Highlight(string(config.ManifestSourceLocal)),
+				style.Highlight(string(config.ManifestSourceRemote)),
+			)
+	}
+	return nil
+}
```
zimeg, line 56: 👁️‍🗨️ note: Adjacent comment suggests adding "IsValid" to the manifest source configurations that I think might be useful instead of flag specific checks here?

👾 note: I'm not so confident with `PreRunE` flag checks but am thinking that might return both truth values with an optional error:

```go
func (ms ManifestSource) IsValid() (bool, error)
```

zimeg (commented):
💬 @srtaalej Handfulls more comment around consolidating on config and flag options across commands! The PRs looking solid but I'd like to keep internal representations together with this!

internal/manifest/sync.go
```diff
@@ -77,12 +86,12 @@ func Sync(ctx context.Context, clients *shared.ClientFactory, app types.App, aut
 
 	var merged types.AppManifest
 	switch {
-	case clients.Config.ForceFlag:
+	case config.ManifestSource(clients.Config.ManifestSourceFlag).Equals(config.ManifestSourceLocal):
```
mwbrooks: issue: This drops `|| clients.Config.ForceFlag` from the local-merge case, so `--force` alone (no `--manifest-source=local`) no longer takes the local-merge path and falls through to the interactive-required error instead. Required check `Lints and Tests` is failing at this head on exactly that: `Test_Sync/force_flag_merges_all_local_and_pushes_to_API` plus four cascading subtests in the same table. `--force` is still a live flag on `manifest sync` per this PR (only `--force-remote` on `sync` got folded away), so this needs the same `|| clients.Config.ForceFlag` the remote branch below kept.

internal/pkg/apps/install.go
mwbrooks: note: `--manifest-source` still short-circuits ahead of the stored `manifestSource` with no mention in the flag help text ("resolve manifest differences using this source (local or remote)"). Same finding carried through three redrafts now: worth a line documenting that this overrides a pinned project source.

internal/config/config.go
```diff
@@ -53,8 +53,8 @@ type Config struct {
 	DeprecatedWorkspaceFlag string
 	DisableTelemetryFlag    bool
 	ForceFlag               bool
-	ForceRemoteFlag         bool
 	LogstashHostResolved    string
+	ManifestSourceFlag      string
```
zimeg, line 57: 🌟 praise: Super appreciate keeping shared flags here!

internal/pkg/apps/install.go
```diff
@@ -701,6 +701,12 @@ func shouldUpdateManifest(ctx context.Context, clients *shared.ClientFactory, ap
...
+		return true, nil
+	}
```
zimeg: 🏁 quibble(non-blocking): Our flags overall might need some care but it'd be cool to use something as:

```go
config.ManifestSource(clients.Config.ManifestSourceFlag).Equals(config.ManifestSourceRemote)
```

zimeg (approved):
@srtaalej Kind thanks for all the iterations once more and LGTM let's get this merged 🚢 💨

PR #639, opened 2026-08-17
ci: use client-id instead of deprecated app-id for cli releaser github app

zimeg:
## What

`actions/create-github-app-token` deprecated its **`app-id`** input in favor of **`client-id`**. Every run of `release.yml` and `dependencies.yml` was logging:

> `Input 'app-id' has been deprecated with message: Use 'client-id' instead.`

This switches both token steps to `client-id`.

## ✅ Secret in place — ready to merge

`client-id` takes the App's **Client ID** (string like `Iv23li…`), not the numeric **App ID** that `app-id` took. The required repo secret **`GH_APP_CLIENT_ID_RELEASER`** has been added (value = the releaser App's Client ID). Both workflow lines reference it. `GH_APP_PRIVATE_KEY_RELEASER` is unchanged.

## Files
- `.github/workflows/release.yml`
- `.github/workflows/dependencies.yml`

## Notes
- The old `GH_APP_ID_RELEASER` secret is no longer referenced by any workflow — safe to delete after merge, or leave in place (harmless).
- Non-breaking on its own — `app-id` still functioned; this clears the deprecation warning and future-proofs against its eventual removal.


mwbrooks:
Nice cleanup — clears the deprecation warning and the diff is minimal. LGTM on the two workflow files.

One follow-up before merge: `.github/MAINTAINERS_GUIDE.md:537` still lists `GH_APP_ID_RELEASER` in the "Credentials and secrets" section. After this PR, no workflow references that secret, so anyone bootstrapping the releaser app from the maintainer guide (fork, rotation, new environment) would set the wrong secret name and hit an empty `client-id` at runtime. Suggest updating that bullet to `GH_APP_CLIENT_ID_RELEASER` in this PR so the docs and workflows stay in sync.

While you're at it, since you already noted the old `GH_APP_ID_RELEASER` secret is unreferenced, it'd be worth deleting it from the repo secrets as part of this change rather than leaving it as a post-merge TODO — keeps the secret surface tidy and avoids the "harmless leftover" drifting into confusion later.

mwbrooks (approved):

PR #642, opened 2026-08-19
fix: use docs site paths for command reference See also links

lukegalbraithrussell:
## Summary

The generated command reference "See also" links pointed to bare slugs (e.g. `slack_app_delete`) that do not properly resolve on the docs site, thus dead-linking. 

This changes `docgen` to emit absolute docs site paths (`/tools/slack-cli/reference/commands/<name>/`), matching the link convention already used in the hand-authored reference docs (e.g. `docs/reference/experiments.md`).

## Changes

- Added a `commandsDocsURLPath` constant and `commandDocsURL()` helper in `cmd/docgen/docgen.go`.
- Applied it to both the parent and child links in the "See also" section.
- Regenerated all command reference files via `slack docgen ./docs/reference`.


zimeg (approved):
@lukegalbraithrussell Thanks immense for catching this 🤓

Before merging it'd be great to add a few tests to prevent regressing on this logic but no blocker!

lukegalbraithrussell:
@zimeg Added some unit tests!

haleychaas (approved):

cmd/docgen/docgen_test.go
```diff
@@ -171,3 +172,47 @@ func TestNewDocsCommand(t *testing.T) {
 		return NewCommand(clients)
 	})
 }
+
+func Test_commandDocsURL(t *testing.T) {
```
zimeg, line 176: 🧪 🏆 praise: Thanks for keeping this confident!

🎁 note: I'm unsure if `docgen` or the internal `docs` package is better for these patterns and test but for now it's no blocker!

cmd/docgen/docgen_test.go
```diff
@@ -171,3 +172,47 @@ func TestNewDocsCommand(t *testing.T) {
...
+		},
+	}
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			assert.Equal(t, tc.expected, commandDocsURL(tc.commandPath))
+		})
+	}
+}
+
+func Test_genMarkdownCommand(t *testing.T) {
```
zimeg, line 201: 👁️‍🗨️ note: I'm more curious if this can be added to an existing test case above but we can keep this in mind for ongoing changes too.

zimeg (approved):
@lukegalbraithrussell Thanks for bringing this to better paths! LGTM!

cmd/docgen/docgen.go
```diff
@@ -34,6 +34,15 @@ import (
 // defaultDocsDirPath is the default value for the <path> argument
 const defaultDocsDirPath = "docs"
 
+// commandsDocsURLPath is the docs site base path for command reference pages.
+// "See also" links must resolve against this path rather than a bare slug.
+const commandsDocsURLPath = "/tools/slack-cli/reference/commands/"
+
+// commandDocsURL returns the docs site link for a command path (e.g. "slack app").
+func commandDocsURL(commandPath string) string {
+	return commandsDocsURLPath + strings.ReplaceAll(commandPath, " ", "_") + "/"
```
mwbrooks: praise: One helper for the base path plus the space-to-underscore rule is the right shape, and Test_genMarkdownCommand asserting NotContains "](slack_app_delete)" pins the actual regression instead of only the new happy path, so a future refactor that reverts to bare slugs fails the test rather than shipping broken "See also" links.

PR #646, opened 2026-08-21
feat: add 'app request' to check install approval requests status

AmyScript:
**Changelog**
Added a `slack app request` command that checks the status of your most recent request to have an app approved for install.
Fixed the help template so hidden subcommands are no longer listed under their parent command.

**Summary**
Installing an app on a team with admin approval required creates an approval request, but there was no way to check what happened to it from the CLI. This adds a `slack app request` command (aliased as requests) that reports the most recent request for the selected app on each team in the token's scope.

Requests are searched on the team of the authenticated account. An account of a workspace belonging to an organization also searches that organization, while an account of an organization searches the organization alone. Other workspaces of an organization can be searched with `--workspace-ids`, which the API accepts up to 50 of.

Apps saved to a project are chosen with the usual prompt. Since that prompt only offers apps recorded in the project, an app can also be named by ID with `--app`, which skips both the project requirement and the project app list and gathers a token from the authenticated accounts instead.

Output includes the request ID, status, and timestamps, plus the actor that cancelled a request and a hint when the account can install the app without approval. Only team IDs are returned by the endpoint, so a team is titled by name when it happens to be the team of the authenticated account:

```
🔒 App Install Approval Requests
   App ID:       A01AX7UTK19
   teamname (T0123456789):
     Request ID:   Ar0123456789
     Status:       pending
     Requested:    2026-08-21 15:04:05 -04:00
```

A denied request adds the moment it was reviewed:

```
🔒 App Install Approval Requests
   App ID:       A01AX7UTK19
   teamname (T0123456789):
     Request ID:   Ar0123456789
     Status:       denied
     Requested:    2026-08-21 15:04:05 -04:00
     Resolved:     2026-08-22 09:30:00 -04:00
```

A cancelled request names the kind of actor that cancelled it, which is an admin, the system, or you:

```
🔒 App Install Approval Requests
   App ID:       A01AX7UTK19
   teamname (T0123456789):
     Request ID:   Ar0123456789
     Status:       cancelled
     Requested:    2026-08-21 15:04:05 -04:00
     Resolved:     2026-08-22 09:30:00 -04:00
     Cancelled by: an admin
```

A pending request where the user can install without approval says so when the account turns out to be able to install the app without approval, because users who can install without approval are still able to create requests:

```
🔒 App Install Approval Requests
   App ID:       A01AX7UTK19
   teamname (T0123456789):
     Request ID:   Ar0123456789
     Status:       pending
     Requested:    2026-08-21 15:04:05 -04:00
     You can install this app without approval. Please cancel the request.
```

### Design considerations

`slack app request` was chosen as the command because Slack's vocabulary splits by role rather than by feature: the person who wants the app requests it, and the admin approves it. This command runs as the requester, reading their own request status, so it sits on the request half.

On the requester side, request also shows up in our CLI and UI:

- The CLI install flow: "Your request to install the app is pending", "Please submit a request to install or update your app" (internal/pkg/apps/install.go), and "Cancel the current request to install this app?" (internal/api/app.go).
- On the app settings page under Install App, when approval is on, the button is Request to Install, and Slackbot follows up with "Your request to install [APP] on [WORKSPACE] has been sent for review."
- The member-facing help article has a "Request app approval" section: https://slack.com/help/articles/202035138-Add-apps-to-your-Slack-workspace

### Preview

No recording yet — the command is unreleased and hidden, and the output above is taken from the unit tests rather than a live run.

### Testing

- Unit tests cover the API client against a mocked server and the output formatting for every status and cancellation actor.
- Both app selection paths are covered: the project prompt and the `--app` app ID path, including a failed team selection and a team without a token.
- Verified the experiment gate rejects the command without `--experiment app-approval-status`, that a project is still required unless an app ID is given, and that the API error for more than 50 workspaces surfaces to the developer.
- Output is sorted by team and leaves the requests of the caller in the order they arrived.
- The error codes returned by the endpoint are asserted through their rendered message and remediation rather than the bare code.

### Notes

- The command is registered unconditionally and marked `Hidden`, matching how `manifest sync` gates itself; the experiment check lives in `PreRunE`.
- Adds the `feature_not_enabled` and `restricted_action` error codes returned by the endpoint, named `ErrAPIFeatureNotEnabled` and `ErrAPIRestrictedAction` so it is clear they mirror API responses rather than errors raised by the CLI.
- Fixes a pre-existing bug in the help template where hidden subcommands were still listed under a parent command. Without it, `slack app --help` lists this command despite `Hidden`. It is a one line change that can be reverted on its own if it ever needs to be.
- `PromptTeamSlackAuth` now calls `SetSelectedAuth` when only one account is authenticated, which the two multi-account paths already did. This sends API calls to the host that account belongs to, and also applies to `app link` and `sandbox`, which use the same prompt.
- `app_id` is required by the endpoint, so requests are reported one app at a time rather than for every app the account owns.

### Requirements


cmd/app/requests.go
```diff
@@ -0,0 +1,242 @@
...
+		requestsText = append(requestsText, field("Request ID", request.ID))
+		requestsText = append(requestsText, field("Status", formatRequestStatus(request.Status)))
+		requestsText = append(requestsText, field("Requested", formatRequestTime(request.DateCreated)))
+		if request.DateResolved > 0 {
+			requestsText = append(requestsText, field("Resolved", formatRequestTime(request.DateResolved)))
+		}
+		if request.CancelledBy != "" {
+			requestsText = append(requestsText, field("Cancelled by", formatRequestCancelledBy(request.CancelledBy)))
+		}
+		if request.CanSelfApprove {
```
mwbrooks: issue: The self-approve hint fires on `CanSelfApprove` alone, so it prints under every status, and the `Status` field two lines up can already read `cancelled`, `denied`, or `approved`. On a cancelled request the output reads "Status: cancelled" followed by "Please cancel the request", which is contradictory, and on approved or denied there is nothing left to cancel. Gating the remediation sentence on `request.Status == api.AppsApprovalsRequestStatusPending` (or splitting the "you can self-install" fact from the "cancel this" instruction) keeps it accurate in all four states.

cmd/app/requests_test.go
```diff
@@ -0,0 +1,364 @@
...
+			},
+			Expected: []string{"Cancelled by: you"},
+		},
+		"a request cancelled without an actor of its own": {
+			Requests: []api.AppsApprovalsRequest{
+				{ID: "Ar1234", TeamID: "T1234", Status: api.AppsApprovalsRequestStatusCancelled, DateCreated: mockRequestCreated, DateResolved: mockRequestResolved, CancelledBy: api.AppsApprovalsRequestCancelledBySystem},
+			},
+			Expected: []string{"Cancelled by: the system"},
+		},
+		"a denied request that the account can approve itself": {
```
mwbrooks: note: This case locks in the behaviour flagged on `requests.go:193`, asserting the "Please cancel the request" line on a `denied` request. Worth flipping to assert the hint is absent once the status guard lands, so the test pins the intended behaviour rather than the current one.

cmd/app/requests.go
```diff
@@ -0,0 +1,242 @@
...
+	}
+	if selection.App.AppID == "" {
+		return "", "", slackerror.New(slackerror.ErrAppNotFound)
+	}
+	return selection.App.AppID, selection.Auth.Token, nil
+}
+
+// FormatRequestsSuccess formats the install request of each team for an app
+func FormatRequestsSuccess(appID string, requests []api.AppsApprovalsRequest) (secondaryText []string) {
+	sort.Slice(requests, func(i, j int) bool {
```
mwbrooks: note: `sort.Slice` reorders the caller's slice in place, so `FormatRequestsSuccess` quietly mutates `result.Requests` for anything that reads it after formatting. Nothing does today and sorting by team ID is the right call for stable output, but a formatter is a surprising place to lose the API's ordering. Copying into a local slice before sorting sidesteps it.

cmd/help/help.go
```diff
@@ -96,7 +96,7 @@ const charmHelpTemplate string = `{{.Long | ToDescription}}
 {{if eq .Name (GetProcessName)}}{{Header "Commands"}}{{range .Commands}}{{if and (.HasAvailableSubCommands) (not .Hidden)}}
   {{.Name | ToGroupName }}{{range .Commands}}{{if (not .Hidden)}}
     {{rpad .Name .NamePadding | ToCommandText}} {{.Short | ToDescription}}{{end}}{{end}}{{end}}{{end}}{{if and (.HasAvailableSubCommands) (not .Hidden)}}{{range .Commands}}{{if and (not .HasAvailableSubCommands) (not .Hidden)}}{{if not (IsAlias .Name $.Data.Aliases)}}
-  {{(rpad .Name .NamePadding) | ToGroupName }}{{.Short | ToDescription}}{{end}}{{end}}{{end}}{{end}}{{else}}{{Header "Subcommands"}}{{if and (.HasAvailableSubCommands) (not .Hidden)}}{{range .Commands}}{{if not .HasAvailableSubCommands}}
+  {{(rpad .Name .NamePadding) | ToGroupName }}{{.Short | ToDescription}}{{end}}{{end}}{{end}}{{end}}{{else}}{{Header "Subcommands"}}{{if and (.HasAvailableSubCommands) (not .Hidden)}}{{range .Commands}}{{if and (not .HasAvailableSubCommands) (not .Hidden)}}
```
mwbrooks: suggestion: Good catch on a pre-existing bug, and it changes help output well beyond this PR: every parent command with hidden children now filters them out of the Subcommands block, here and in the legacy template on line 142. Nothing pins that behaviour, so a future template edit can regress it without a red test. A small golden-output assertion that a hidden subcommand is absent from `slack app --help` would protect the fix that makes `app requests` hidden in the first place.

cmd/app/requests_test.go
```diff
@@ -0,0 +1,364 @@
...
+	"github.com/stretchr/testify/mock"
+)
+
+// mockRequestCreated is the moment a mocked request was made
+var mockRequestCreated = time.Date(2026, 8, 21, 15, 4, 5, 0, time.UTC).Unix()
+
+// mockRequestResolved is the moment a mocked request was reviewed
+var mockRequestResolved = time.Date(2026, 8, 22, 9, 30, 0, 0, time.UTC).Unix()
+
+func TestRequestsCommand(t *testing.T) {
```
mwbrooks: praise: Thorough table coverage for a brand new command, and the experiment gate is exercised in both directions rather than only the rejecting one, which is the half that usually gets skipped.

cmd/app/requests.go
```diff
@@ -0,0 +1,242 @@
...
+						requestsTeamsLimit,
+					)
+			}
+			clients.Config.SetFlags(cmd)
+			// An app named by ID is checked without the apps of a project
+			if types.IsAppID(clients.Config.AppFlag) {
+				return nil
+			}
+			// Verify command is run in a project directory
+			return cmdutil.IsValidProjectDirectory(clients)
```
zimeg: 🔬 quibble(non-blocking): It could be nice to have remediation to use the `--app` flag here! I'm a fan of these commands!

cmd/app/requests.go
```diff
@@ -0,0 +1,242 @@
...
+						style.CommandText("--experiment app-approval-status"),
+					)
+			}
+			if len(requestsFlags.teamIDs) > requestsTeamsLimit {
+				return slackerror.New(slackerror.ErrInvalidArguments).
+					WithMessage("The %s flag accepts at most %d teams",
+						style.CommandText("--team-ids"),
+						requestsTeamsLimit,
+					)
+			}
```
zimeg:
```suggestion
```

🪓 question: Is this error also returned from the API? I'm wondering if we can remove it from here to avoid version mismatches if this changes!

cmd/app/requests.go
```diff
@@ -0,0 +1,242 @@
...
+func requestsAppSelection(ctx context.Context, clients *shared.ClientFactory) (appID string, token string, err error) {
+	if types.IsAppID(clients.Config.AppFlag) {
+		auth, err := requestsTeamSelectPromptFunc(ctx, clients, "Select an account to search with", nil)
+		if err != nil {
+			return "", "", err
+		}
+		if auth == nil || auth.Token == "" {
+			return "", "", slackerror.New(slackerror.ErrCredentialsNotFound)
+		}
+		clients.Auth().SetSelectedAuth(ctx, *auth, clients.Config, clients.Os)
```
zimeg:
```suggestion
```

🪓 question: Can this line be kept with `prompts.PromptTeamSlackAuth`? I think it has shared selection helpers for commands using the API.

cmd/app/requests.go
```diff
@@ -0,0 +1,242 @@
...
+			}
+			// Verify command is run in a project directory
+			return cmdutil.IsValidProjectDirectory(clients)
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runRequestsCommand(cmd, clients)
+		},
+	}
+
+	cmd.Flags().StringSliceVar(&requestsFlags.teamIDs, "team-ids", nil, "also check these teams of an organization,\nwith a maximum of 50 teams")
```
zimeg:
```suggestion
	cmd.Flags().StringSliceVar(&requestsFlags.teamIDs, "workspace-ids", nil, "also check these workspaces of an organization,\nwith a maximum of 50 workspaces")
```

🪬 question: Is this flag limited to workspaces within an organization? I'm uncertain if the top-level team is supported but I'm wanting to avoid confusion with the `--team` flag.

cmd/app/requests.go
```diff
@@ -0,0 +1,242 @@
...
+
+// formatRequestCancelledBy names the kind of actor that cancelled a request.
+// Every returned request was made by the authenticated account, so a request
+// cancelled by a user was withdrawn by that same account.
+func formatRequestCancelledBy(actor api.AppsApprovalsRequestCancelledBy) string {
+	switch actor {
+	case api.AppsApprovalsRequestCancelledByAdmin:
+		return "an admin"
+	case api.AppsApprovalsRequestCancelledBySystem:
+		return "the system"
```
zimeg, line 217: 🤖 praise: Solid API patterns I think!

cmd/help/help.go
zimeg, line 99: 🧮 quibble(non-blocking): Separate PRs might be ideal for adjacent changes for possible revert sake later since this is a nice find to keep!

internal/slackerror/errors.go
```diff
@@ -787,6 +789,12 @@ Otherwise start your app for local development with: %s`,
 		Message: "Couldn't save the logs",
 	},
 
+	ErrFeatureNotEnabled: {
+		Code:        ErrFeatureNotEnabled,
+		Message:     "This feature is not enabled for the team",
+		Remediation: "Reach out to an admin for additional information",
+	},
+
```
zimeg, line 797:
```suggestion
```

🪓 note: I don't think this is used in calling code and might be nice to keep out? Would API responses be alright to match this?

internal/slackerror/errors.go
```diff
@@ -1382,6 +1390,12 @@ Otherwise start your app for local development with: %s`,
 		Message: "Must include a request_id or app_id",
 	},
 
+	ErrRestrictedAction: {
+		Code:        ErrRestrictedAction,
+		Message:     "The requested action is not allowed for a specified team",
+		Remediation: "Check that each team belongs to the organization of the authenticated account",
+	},
+
```
zimeg, line 1398:
```suggestion
```

🌲 ramble: I fear checking for errors perhaps unreferenced similar but I'm wondering if this is used in unit tests or when handling the API response? If it's alright to remove I might suggest but all so is rambles!

zimeg (approved):
🏁 @AmyScript Thanks for sending this in! It's LGTM per usual and I leave a few comments that might be nice before merge but some are rambles... 🗣️ 

Overall I lean toward returning API responses or errors instead of expecting certain values and some flag options and error outputs. Nothing blocking! These are exciting experiment to bring to release 🤖⚡

cmd/app/requests.go
AmyScript: yes good catch, it is returned by the API. I'll remove the code here.

cmd/help/help.go
AmyScript: Without that template fix, hidden subcommands are listed under their parent, so slack app --help would show requests even though it's Hidden. Splitting it means this PR is either temporarily leaky or has to wait on the other one to merge.  If you are ok with it, perhaps we can keep this as it's one line revert if we need to do it later

cmd/app/requests.go
AmyScript: It is limited to workspaces within an organization because if the AppId belongs to a standalone workspace, this wouldn't have to be passed in as it would return for the top-level team.  I'll update it to avoid confusion.

cmd/app/requests.go
AmyScript: Yes it can! good catch.

AmyScript: I did have to add `clients.Auth().SetSelectedAuth(ctx, allAuths[0], clients.Config, clients.Os)` inside `PromptTeamSlackAuth` with the existing `if len(allAuths) == 1 {return &allAuths[0], nil}` so that the request goes to the host that account resolved to even if there is only 1 team that is authed.

internal/slackerror/errors.go
AmyScript: It is used, just not through explicit calling code. `slackerror.NewAPIError` calls `getErrorIfKnown`, which looks up `ErrorCodeMap` by whatever code the API returned, so this entry is what turns `feature_not_enabled` into a message and remediation. Without it the user sees the bare code. I added an API test case asserting the rendered message so the wiring is visible in the suite.

internal/slackerror/errors.go
AmyScript: Same mechanism as above. `Test_Client_ListAppApprovalRequests` has an "errors when a team is outside of the organization" case returning `restricted_action`. I strengthened it to assert the mapped message and remediation rather than just the code, so removing the entry would now fail the test rather than silently degrade the output.

cmd/app/requests.go
```diff
@@ -0,0 +1,239 @@
...
+		requestsText = append(requestsText, field("Request ID", request.ID))
+		requestsText = append(requestsText, field("Status", formatRequestStatus(request.Status)))
+		requestsText = append(requestsText, field("Requested", formatRequestTime(request.DateCreated)))
+		if request.DateResolved > 0 {
+			requestsText = append(requestsText, field("Resolved", formatRequestTime(request.DateResolved)))
+		}
+		if request.CancelledBy != "" {
+			requestsText = append(requestsText, field("Cancelled by", formatRequestCancelledBy(request.CancelledBy)))
+		}
+		if request.CanSelfApprove {
```
mwbrooks: issue: The self-approve hint is gated on `CanSelfApprove` alone, so it prints under every status, and the Status field two lines up can already read `approved`, `cancelled`, or `denied`. On a cancelled request the output reads "Status: cancelled" followed by "Please cancel the request", and on approved or denied there is nothing left to cancel. Gating the remediation sentence on `request.Status == api.AppsApprovalsRequestStatusPending` (or splitting the "you can self-install" fact from the "cancel this" instruction) keeps it accurate in all four states.

cmd/app/requests_test.go
```diff
@@ -0,0 +1,373 @@
...
+			},
+			Expected: []string{"Cancelled by: the system"},
+		},
+		"a denied request that the account can approve itself": {
+			Requests: []api.AppsApprovalsRequest{
+				{ID: "Ar1234", TeamID: "T1234", Status: api.AppsApprovalsRequestStatusDenied, CanSelfApprove: true, DateCreated: mockRequestCreated, DateResolved: mockRequestResolved},
+			},
+			Expected: []string{
+				"Status:       denied",
+				"You can install this app without approval. Please cancel the request.",
```
mwbrooks: note: This case locks in the behaviour flagged on `requests.go:190`, asserting the "Please cancel the request" line on a denied request. Worth flipping to assert the hint is absent once the status guard lands, so the suite pins the intended behaviour rather than the current one.

cmd/app/requests.go
```diff
@@ -0,0 +1,239 @@
...
+				)
+			}
+			return nil
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runRequestsCommand(cmd, clients)
+		},
+	}
+
+	cmd.Flags().StringSliceVar(&requestsFlags.workspaceIDs, "workspace-ids", nil, "also check these workspaces of an organization,\nwith a maximum of 50 workspaces")
```
mwbrooks: suggestion: Moving the fifty-workspace check to the API makes sense, and it does cost the message that named the cap. The removed guard said "The --team-ids flag accepts at most 50 teams"; `invalid_arguments` carries `Message: "Slack API request parameters are invalid"` with no `Remediation`, so fifty-one workspaces now surfaces a generic parameter error while this flag's help still promises the limit. Both codes this PR adds (`feature_not_enabled`, `restricted_action`) got a `Remediation`, so the same treatment would keep the failure self-explanatory here. Since `invalid_arguments` is shared, a `WithRemediation` at the `ListAppApprovalRequests` call site is probably the cleaner place for it.

internal/prompts/team_select.go
```diff
@@ -40,6 +40,7 @@ func PromptTeamSlackAuth(ctx context.Context, clients *shared.ClientFactory, pro
 	}
 
 	if len(allAuths) == 1 {
+		clients.Auth().SetSelectedAuth(ctx, allAuths[0], clients.Config, clients.Os)
```
mwbrooks: note: Good place for this. The single-auth early return was the only path in this function that skipped `SetSelectedAuth` (lines 87 and 108 both call it), so fixing it here rather than in `requests.go` closes the gap for every caller. Worth naming the blast radius: `cmd/app/link.go` and `cmd/sandbox/sandbox.go` also call `PromptTeamSlackAuth`, and both now get a `SetSelectedAuth` on the one-account path where they previously did not. Nothing in this PR exercises either, so a quick sanity check on those two before merge would be worthwhile.

internal/prompts/team_select_test.go
```diff
@@ -0,0 +1,56 @@
...
+			clientsMock.Auth.On(Auths, mock.Anything).Return(tc.auths, nil)
+			clientsMock.AddDefaultMocks()
+			clients := shared.NewClientFactory(clientsMock.MockClientFactory())
+
+			auth, err := PromptTeamSlackAuth(ctx, clients, "Select a team", nil)
+
+			require.NoError(t, err)
+			assert.Equal(t, tc.expectedAuth, *auth)
+			clientsMock.Auth.AssertCalled(t, "SetSelectedAuth", mock.Anything, tc.expectedAuth, clients.Config, clients.Os)
+			clientsMock.IO.AssertNotCalled(t, SelectPrompt)
```
mwbrooks: praise: First coverage for this helper, and it asserts the side effect and the absence of a prompt rather than only the returned auth, which is the half that would catch a regression on the line above.

cmd/app/requests.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+	if selection.App.AppID == "" {
+		return "", "", slackerror.New(slackerror.ErrAppNotFound)
+	}
+	return selection.App.AppID, selection.Auth.Token, nil
+}
+
+// FormatRequestsSuccess formats the install request of each team for an app
+func FormatRequestsSuccess(appID string, requests []api.AppsApprovalsRequest) (secondaryText []string) {
+	sort.Slice(requests, func(i, j int) bool {
```
mwbrooks: note: `sort.Slice` reorders the caller's slice in place, so `FormatRequestsSuccess` quietly mutates `result.Requests` for anything that reads it after formatting. Nothing does today and sorting by team ID is the right call for stable output, but a formatter is a surprising place to lose the API's ordering. Copying into a local slice before sorting sidesteps it.

mwbrooks (changes requested):
@AmyScript Please don't merge this yet - I'd like a moment to review it. Just dropping the change request, since @zimeg already approved it which enables merging.

internal/slackerror/errors.go
mwbrooks: It would be good to update the name to `ErrAPIFeatureNotEnabled` so that it's self-documenting that this is an API error and not an internal Slack CLI error. Otherwise, future maintainers may remove it under the same assumption as @zimeg.

internal/slackerror/errors.go
mwbrooks: Same suggestion as above to rename it `ErrAPIRestrictedAction` so that it's self-documenting that it's an API error.

cmd/app/requests.go
```diff
@@ -0,0 +1,239 @@
...
+	workspaceIDs []string
+}
+
+var requestsFlags requestsCmdFlags
+
+// NewRequestsCommand returns a new Cobra command
+func NewRequestsCommand(clients *shared.ClientFactory) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "requests [flags]",
+		Aliases: []string{"approval-requests", "approvals"},
```
AmyScript, line 57: @mwbrooks here are the aliases I was talking about.

cmd/help/help.go
zimeg: @AmyScript No worries keeping it here! I don't know if other hidden subcommands were exposed before this change but it might be nice to also note in chagnelog too if it's worth a callout!

cmd/app/requests.go
zimeg: @AmyScript Thanks for catching that too! I think if more edges are found we should adjust this to set selected auth info in a [`defer`](https://go.dev/blog/defer-panic-and-recover) statement but I think the current approach is great!

cmd/app/requests_test.go
```diff
@@ -0,0 +1,373 @@
...
+			},
+			Expected: []string{"Status:       escalated"},
+		},
+		"an unrecognized cancellation actor is reported as named": {
+			Requests: []api.AppsApprovalsRequest{
+				{ID: "Ar1234", TeamID: "T1234", Status: api.AppsApprovalsRequestStatusCancelled, DateCreated: mockRequestCreated, CancelledBy: api.AppsApprovalsRequestCancelledBy("workflow")},
+			},
+			Expected: []string{"Cancelled by: workflow"},
+		},
+		"requests are sorted by the team ID": {
```
zimeg, line 336: 🧮 praise: Thanks for bringing consistent outputs to this command!

internal/prompts/team_select_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+	"github.com/slackapi/slack-cli/internal/shared"
+	"github.com/slackapi/slack-cli/internal/shared/types"
+	"github.com/slackapi/slack-cli/internal/slackcontext"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
+)
+
+func TestPromptTeamSlackAuth(t *testing.T) {
```
zimeg, line 28: 🦠 praise: Thanks for adding coverage to this prompt!

🔬 ramble: I'm hoping we can improve the mock longterm to avoid stubbing these selections within each command and this is so helpful in building confidence toward this I think.

zimeg (approved):
@AmyScript Super appreciate the fast changes! I'll toss another approval on but defer to @mwbrooks for ongoing review. Nothing is blocking experimental releases for me!

🧪 I do notice the E2E tests aren't running and I think that's an intentional setting to avoid passing secrets to forked branches... It catches us often and I'm unsure a workaround remains but I share this for possible updates:

🔗 https://github.com/slackapi/slack-cli/blob/main/.github/MAINTAINERS_GUIDE.md#fork

AmyScript:
> @AmyScript Super appreciate the fast changes! I'll toss another approval on but defer to @mwbrooks for ongoing review. Nothing is blocking experimental releases for me!
> 
> 🧪 I do notice the E2E tests aren't running and I think that's an intentional setting to avoid passing secrets to forked branches... It catches us often and I'm unsure a workaround remains but I share this for possible updates:
> 
> 🔗 https://github.com/slackapi/slack-cli/blob/main/.github/MAINTAINERS_GUIDE.md#fork

Should I have just branched off main?  I read about the fork so I forked 😆. Is there a way for me to run the e2e tests locally with a command?

cmd/app/requests.go
```diff
@@ -0,0 +1,239 @@
...
+type requestsCmdFlags struct {
+	workspaceIDs []string
+}
+
+var requestsFlags requestsCmdFlags
+
+// NewRequestsCommand returns a new Cobra command
+func NewRequestsCommand(clients *shared.ClientFactory) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "requests [flags]",
```
mwbrooks, line 56: suggestion: After some thought, I'd like to suggest renaming this to `approval` (singular)

The reason is that `app approval` matches the API (`apps.approvals.requests.list`) and leaves room for `slack app approval [create|cancel|list]` later. 

We've had requests for a cancel feature, so this opens the namespace for it. 

The bare `approval` also matches the bare `app → app list`, `trigger → trigger list`, `auth → auth list`, and `manifest → manifest list` convention. So, if we expand the command with `create | cancel | list` subcommands, we've made decisions that allow for a smooth forward rollout.

The `approvals` command also matches the terminology used in our documentation, where as `requests` is more from the admin-perspective:
- https://docs.slack.dev/admins/managing-app-approvals/
- https://slack.com/help/articles/222386767-Manage-app-approval-for-your-workspace

We should alias the plural `approvals`. The CLI usually has the canonical command as singular while aliasing the plural. Long ago, we mixed-and-matched singular/plural depending on the command and it was confusing for developers. So, we try to use singular when possible.

cmd/app/requests.go
mwbrooks: issue: This hint prints under every status, including approved and cancelled. Gate it on a pending request too.

cmd/app/requests_test.go
mwbrooks: note: This case pins the hint onto a `denied` request, so the behaviour above is locked in. Update it alongside the fix.

cmd/app/requests.go
mwbrooks, line 111: thought: Is the 50 workspace limit enforced by the API?  A brief look and it appears the API returns `invalid_arguments` with no remediation. If so, we may want to validate the length here and include it in the error message since it's easy to overlook the help documentation.

internal/prompts/team_select.go
mwbrooks, line 43: note: Right place for this, but it also changes `app link` and `sandbox`, which now get the selected auth set on the single-account path. Worth confirming neither relies on it being unset.

cmd/app/requests.go
mwbrooks, line 168: suggestion: `sort.Slice` reorders the caller's original slice. Harmless today, but this function mutating its input will be surprise later if used elsewhere and can introduce unexpected bugs. I'd recommend sorting a copy of `requests` (`requestsSorted`).

cmd/app/app.go
```diff
@@ -52,6 +52,7 @@ func NewCommand(clients *shared.ClientFactory) *cobra.Command {
 	cmd.AddCommand(NewDeleteCommand(clients))
 	cmd.AddCommand(NewLinkCommand(clients))
 	cmd.AddCommand(NewListCommand(clients))
+	cmd.AddCommand(NewRequestsCommand(clients))
```
mwbrooks, line 55: praise: Appreciate the attention to detail on naming conventions and alphabetical order. 🙇🏻

cmd/app/requests.go
```diff
@@ -0,0 +1,239 @@
...
+const requestsTimeFormat = "2006-01-02 15:04:05 Z07:00"
+
+// Handle to a function used for testing
+var requestsAppSelectPromptFunc = prompts.AppSelectPrompt
+
+// Handle to a function used for testing
+var requestsTeamSelectPromptFunc = prompts.PromptTeamSlackAuth
+
+// Flags
+
```
mwbrooks, line 46: suggestion: Remove the extra whitespace.
```suggestion
```

cmd/app/requests.go
mwbrooks: suggestion: Assuming we go with `app approval`, alias `approvals` and remove the others. It's hard to remove an alias once we add it, but it's easy to add one in the future.
```suggestion
		Aliases: []string{"approvals"},
```

cmd/app/requests.go
```diff
@@ -0,0 +1,239 @@
...
+}
+
+var requestsFlags requestsCmdFlags
+
+// NewRequestsCommand returns a new Cobra command
+func NewRequestsCommand(clients *shared.ClientFactory) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "requests [flags]",
+		Aliases: []string{"approval-requests", "approvals"},
+		Short:   "Check requests to install the app",
```
mwbrooks, line 58: suggestion: Most developers aren't familiar with AAA, so I think the extra word can go along way and aligns with our documentation.
```suggestion
		Short:   "Check approval requests to install the app",
```

cmd/app/requests.go
```diff
@@ -0,0 +1,239 @@
...
+			"with the --team flag or a prompt.",
+			"",
+			"Apps saved to a project are chosen with a prompt, but any app can be checked",
+			"by app ID with the --app flag, which does not require a project.",
+		}, "\n"),
+		Hidden: true,
+		Example: style.ExampleCommandsf([]style.ExampleCommand{
+			{Command: "app requests", Meaning: "Check requests to install an app"},
+			{Command: "app requests --app A0123456789", Meaning: "Check requests for an app outside a project"},
+			{Command: "app requests --workspace-ids T0123456789,T9876543210", Meaning: "Check requests on certain workspaces of an organization"},
```
mwbrooks, line 80: suggestion: Assuming we go with `app approval`:
```suggestion
			{Command: "app approval", Meaning: "Check requests to install an app"},
			{Command: "app approval --app A0123456789", Meaning: "Check requests for an app outside a project"},
			{Command: "app approval --workspace-ids T0123456789,T9876543210", Meaning: "Check requests on certain workspaces of an organization"},
```

cmd/app/requests.go
```diff
@@ -0,0 +1,239 @@
...
+
+	cmd.Flags().StringSliceVar(&requestsFlags.workspaceIDs, "workspace-ids", nil, "also check these workspaces of an organization,\nwith a maximum of 50 workspaces")
+
+	return cmd
+}
+
+// runRequestsCommand will execute the requests command
+func runRequestsCommand(cmd *cobra.Command, clients *shared.ClientFactory) error {
+	ctx := cmd.Context()
+	span, ctx := opentracing.StartSpanFromContext(ctx, "cmd.app.requests")
```
mwbrooks, line 119: suggestion: If we accept `app approval` then we should adjust all of our tracing.
```suggestion
	span, ctx := opentracing.StartSpanFromContext(ctx, "cmd.app.approval")
```

cmd/app/requests.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+
+	result, err := clients.API().ListAppApprovalRequests(ctx, token, appID, requestsFlags.workspaceIDs)
+	if err != nil {
+		return err
+	}
+
+	clients.IO.PrintInfo(ctx, false, "\n%s", style.Sectionf(style.TextSection{
+		Emoji:     "lock",
+		Text:      "App Requests",
```
mwbrooks, line 134: suggestion: Adding more clarity for developers who aren't familiar with AAA.
```suggestion
		Text:      "App Approval Requests",
```

mwbrooks (commented):
🙌🏻 Thanks for the PR @AmyScript! Admin approval management has been a painfully missed feature in the CLI, so I'm happy to see you giving it some TLC.

🙇🏻 Some praise goes to you on code quality. You rocked our latest conventions, patterns, and test coverage goals. Thanks for that!

💡 Small tip for next time - please create a branch on this repo instead of forking. External contributors fork, but as a maintainer you can work on a branch allowing our CI/CD to safely run E2E tests. Before merging, we should push this to a branch on the `origin` to ensure the E2E tests pass alright.

🧠 My main ask before approving this PR is to consider `app approval` (alias: `approvals`) instead of `app requests`. I think it reads better from the user/developer perspective, aligns with our documentation, and aligns with our API. It also opens space for 3rd-tier of commands such as `app approval cancel` or `app approval list` (alias `app approval` - what you're landing here).

🧪 I hear you're aiming to land this into production end-of-week or early next week. I'd suggestion you delete the experiment entirely, because we won't be doing a release between now-and-then. All production features should not be behind an experiment.

🚦 When you remove the experiment, please switch this to a `semver:minor`.

📝 The rest of my suggestions are minor. Nice to haves but not blockers if you are short on time.

Once we settle on the name, happy to hit ✅ approve!

cmd/app/requests.go
AmyScript: Good call, I was looking at this and I'm going to change it to App Install Approval Requests to make it more clear.

cmd/app/requests.go
AmyScript: Hey @mwbrooks, thanks for the thought you put into this, and for surfacing that there's been a request for a cancel feature. 

I should have explained the design considerations for using `app requests` in the PR description. My thinking is that Slack's vocabulary splits by role rather than by feature: the person who wants the app *requests* it, and the admin *approves* it. This command runs as the requester, reading their own request status, so it sits on the request half.

That split shows up in the sources you linked, too. Both are written for the approver, and even there, the object is a request: the admin listing method is `admin.apps.requests.list`, the event is `app_requested`, the companion help article is "Manage app requests for your workspace".

On the requester side, request also shows up in our CLI and UI:
- The CLI install flow: "Your request to install the app is pending", "Please submit a request to install or update your app" (`internal/pkg/apps/install.go`), and "Cancel the current request to install this app?" (`internal/api/app.go`).
- On the app settings page under **Install App**, when approval is on, the button is **Request to Install**, and Slackbot follows up with "Your request to install [APP] on [WORKSPACE] has been sent for review."
- The member-facing help article has a "Request app approval" section: https://slack.com/help/articles/202035138-Add-apps-to-your-Slack-workspace

Good point about leaving room for subcommands. The requester side verbs are create, cancel, and list, and those read naturally as `app request cancel`, our own enum comment describes that action as "the requester withdrawing their own request".  Whereas `app approval cancel` sounds like revoking an approval someone granted. That's also why I left `approval` out of the aliases.  If `app request` claims that name now, we can't give it to an admin side `app approval` later if we need it.

You're right that we should use the singular, so I've renamed it to `app request` with `requests` as the only alias.  The command is hidden behind the `app-approval-status` experiment right now, so if you still feel strongly about it, I am happy to switch it if you'd rather go with `approval`.  Let me know what you think!

cmd/app/requests.go
AmyScript: Good suggestion!

internal/prompts/team_select.go
AmyScript: Good call to check. Neither one depends on those fields being unset.

`app link` uses the auth's team ID explicitly when it saves the app (`saveAppToJSON` passes `app.TeamID` to `GetDeployed`/`GetLocal`), so the `TeamFlag` fallback in `getDeployedAppTeamDomain` is never reached, and it has no other team-bound prompts. `sandbox` just passes `auth.Token` to the API and reads nothing else off the config.

cmd/app/requests.go
AmyScript: @zimeg @mwbrooks I need a tiebreak on this one, since the two pieces of feedback point you both provided points in opposite directions and both risks are real.

Eden asked to drop the client-side check so a future change to the cap can't leave a stale CLI rejecting calls the API would accept, and I removed it. Michael then pointed out that the API answers with a bare `invalid_arguments` and no remediation, so the failure doesn't explain itself, and the developer has to go find the limit in the flag help. I think both concerns are valid. I've left the code as is for now, but if there's agreement to go with @mwbrooks' suggestion, I can change it back.

AmyScript:
> 🙌🏻 Thanks for the PR @AmyScript! Admin approval management has been a painfully missed feature in the CLI, so I'm happy to see you giving it some TLC.
> 
> 🙇🏻 Some praise goes to you on code quality. You rocked our latest conventions, patterns, and test coverage goals. Thanks for that!
> 
> 💡 Small tip for next time - please create a branch on this repo instead of forking. External contributors fork, but as a maintainer you can work on a branch allowing our CI/CD to safely run E2E tests. Before merging, we should push this to a branch on the `origin` to ensure the E2E tests pass alright.
> 
> 🧠 My main ask before approving this PR is to consider `app approval` (alias: `approvals`) instead of `app requests`. I think it reads better from the user/developer perspective, aligns with our documentation, and aligns with our API. It also opens space for 3rd-tier of commands such as `app approval cancel` or `app approval list` (alias `app approval` - what you're landing here).
> 
> 🧪 I hear you're aiming to land this into production end-of-week or early next week. I'd suggestion you delete the experiment entirely, because we won't be doing a release between now-and-then. All production features should not be behind an experiment.
> 
> 🚦 When you remove the experiment, please switch this to a `semver:minor`.
> 
> 📝 The rest of my suggestions are minor. Nice to haves but not blockers if you are short on time.
> 
> Once we settle on the name, happy to hit ✅ approve!

@mwbrooks Thanks for your review on my first PR working with this codebase, lots of great suggestions.  The code has been updated for another review.  I've left the naming of the new command for now.  I'm leaving on PTO next week, so that's why I was hoping to launch this by the end of the week, but if we think there are still discussions that need to happen, I'm happy to leave the experiment flag on and merge it for now and pick it back up when I'm back from PTO.

I'll create a branch off the repo once i get a  ✅ from you so we can merge the branch instead.

cmd/app/request.go
```diff
@@ -0,0 +1,259 @@
...
+	if selection.App.AppID == "" {
+		return "", types.SlackAuth{}, slackerror.New(slackerror.ErrAppNotFound)
+	}
+	return selection.App.AppID, selection.Auth, nil
+}
+
+// FormatRequestSuccess formats the install request of each team for an app.
+// Teams found in teamNames are titled by name while others are titled by ID.
+func FormatRequestSuccess(appID string, teamNames map[string]string, requests []api.AppsApprovalsRequest) (secondaryText []string) {
+	sorted := slices.SortedFunc(slices.Values(requests), func(a api.AppsApprovalsRequest, b api.AppsApprovalsRequest) int {
```
mwbrooks, line 171: praise: Thanks for the slice fix!

internal/slackerror/errors.go
```diff
@@ -27,6 +27,8 @@ const (
 	ErrAddAppToProject                               = "add_app_to_project_error"
 	ErrAlreadyLoggedOut                              = "already_logged_out"
 	ErrAlreadyResolved                               = "already_resolved"
+	ErrAPIFeatureNotEnabled                          = "feature_not_enabled"
```
mwbrooks, line 30: praise: Thanks for prefix the `ErrAPI*` codes!

cmd/app/requests.go
mwbrooks: @AmyScript Generally, we'd prefer to have less validation in the CLI and more on the API. So, I'd lean toward having the API enforce the limit. It would be best for the CLI to not document the 50-limit and instead allow the API error to explain the limit.

mwbrooks (changes requested):
@AmyScript Thanks for fixes and changes. 🙇🏻

I understand your reasoning for `app requests` although I still feel it isn't a self-documenting command for approval requests. Regardless, naming is hard and I don't want to block your PR while we try to decide on a better name.

✏️ Last change request before we can merge - please remove the experiment gate. If you want this feature to be available to customers next week, it can't be behind an experiment.

cmd/app/requests.go
AmyScript: Agree!

AmyScript:
Moved to #650, which is the same code on a branch in this repo rather than my fork so that CI can reach the repository secrets.

The only change since the last review here is removing the `app-approval-status` experiment gate and the `Hidden` flag, now that we have the go-ahead to ship the command. Everything else is identical.

Closing this in favor of #650 — thanks for the reviews, and sorry for the thread shuffle.

AmyScript:
Exact branch with passing tests: https://github.com/slackapi/slack-cli/pull/650

mwbrooks (approved):
✅ Thanks @AmyScript, adding an approval to the PR!

🙇🏻 Appreciate all the back and forth. After learning about your tight release schedule, dropping the experiment will allow us to release it anytime.

🧪 Thanks for confirming the our E2E tests pass with PR #650. In the future you can branch off the main repo to avoid that.

PR #652, opened 2026-08-27
fix(install): qualify Windows archive extraction

### LINKED ISSUE #651: Windows installer fails when Pscx shadows Expand-Archive
## Description

The official Windows installer fails in PowerShell when the Pscx module is installed. Pscx 3.3.2 provides an `Expand-Archive` cmdlet that shadows the archive cmdlet from `Microsoft.PowerShell.Archive`. The installer invokes `Expand-Archive` without module qualification and passes `-DestinationPath`, which the Pscx command does not support.

The download succeeds: `%LOCALAPPDATA%\\slack-cli\\slack_cli.zip` is a valid archive containing `bin/slack.exe`. The installer stops before extraction and before updating PATH, so `slack` is not available in a new session.

## CLI Version

Installer script (pre-install failure; no Slack CLI binary version available)

## Operating System

Windows 11; PowerShell 7.6.5; Pscx 3.3.2

## Steps to Reproduce

1. Install/import Pscx 3.3.2 in PowerShell 7.6.5.
2. Verify that `Get-Command Expand-Archive` resolves to the Pscx module.
3. Run the official installer:

   ```powershell
   irm https://downloads.slack-edge.com/slack-cli/install-windows.ps1 | iex
   ```

4. When the installer reaches extraction, observe the parameter error for `-DestinationPath`.

## Expected Result

The installer extracts `bin/slack.exe`, updates the user PATH, and completes successfully regardless of unrelated modules that define commands with the same name.

## Actual Result

The installer downloads the archive but fails at extraction with an error equivalent to:

```text
A parameter cannot be found that matches parameter name 'DestinationPath'.
```

`Get-Command Expand-Archive` resolves to `Pscx` rather than `Microsoft.PowerShell.Archive`, and the installer exits before extraction and PATH update.

## Suggested Fix

Module-qualify the call as `Microsoft.PowerShell.Archive\\Expand-Archive` (or otherwise avoid ambiguous command resolution), for example:

```powershell
Microsoft.PowerShell.Archive\\Expand-Archive -LiteralPath "$slack_cli_dir\\slack_cli.zip" -DestinationPath $slack_cli_dir -Force
```

codywilliamson:
### Changelog

The Windows installer now extracts the CLI reliably when another PowerShell module defines `Expand-Archive`.

### Summary

PowerShell resolves an unqualified command name before binding its parameters. When Pscx 3.3.2 is loaded, the installer selects Pscx's `Expand-Archive`, which does not accept `-DestinationPath`; the ZIP downloads successfully, but extraction and the PATH update never happen. Both release and development installers now address `Microsoft.PowerShell.Archive` explicitly.

Fixes #651.

The regression test is the part worth the closest review. It keeps the change dependency-free: it parses both mirrored installers to require the qualified command, then performs a real extraction while a same-named function shadows unqualified command resolution.

### Testing

```powershell
pwsh -NoProfile -File ./scripts/install-windows-test.ps1
powershell.exe -NoProfile -NonInteractive -ExecutionPolicy Bypass -File ./scripts/install-windows-test.ps1
```

Both commands pass locally. `go test ./...` was also attempted with the repository-requested Go 1.26.6 toolchain on Windows/386, but unrelated command packages simultaneously reached the suite's 11-minute watchdog. The pull request's macOS Go suite remains the authoritative repository-wide check.

### Notes

The CLI runtime and non-Windows installers are unchanged. The added Windows job runs only the focused, Pester-free installer regression.

### Requirements

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -34,6 +34,19 @@ jobs:
           files: ./coverage.out
           fail_ci_if_error: false
 
+  windows-install:
+    name: Windows Installer
+    runs-on: windows-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+      - name: Test Windows installer
+        shell: pwsh
+        run: ./scripts/install-windows-test.ps1
+
   # Monitor code coverage and TODO/FIXME-type comments
   health-score:
     name: Health Score
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -25,8 +25,6 @@ jobs:
         run: go tool golangci-lint run --timeout=5m
       - name: Unit Tests
         run: make test
-      - name: Install Tests
-        run: make test-install
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@fb8b3582c8e4def4969c97caa2f19720cb33a72f # v7.0.0
         with:
@@ -34,16 +32,27 @@ jobs:
           files: ./coverage.out
           fail_ci_if_error: false
 
-  windows-install:
-    name: Windows Installer
-    runs-on: windows-latest
+  install-tests:
+    name: Install Tests
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - macos-latest
+          - ubuntu-latest
+          - windows-latest
+    runs-on: ${{ matrix.os }}
     permissions:
       contents: read
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           persist-credentials: false
-      - name: Test Windows installer
+      - name: Install Tests (Unix)
+        if: runner.os != 'Windows'
+        run: make test-install
+      - name: Install Tests (Windows)
+        if: runner.os == 'Windows'
         shell: pwsh
         run: ./scripts/install-windows-test.ps1
 
--- a/scripts/install-windows-dev.ps1
+++ b/scripts/install-windows-dev.ps1
@@ -163,7 +163,7 @@ function install_slack_cli {
   $slack_cli_new_binary_path = "$($slack_cli_dir)\bin\${confirmed_alias}.exe"
 
   delay 0.3 "Extracting the executable to:`n   $slack_cli_new_binary_path"
-  Expand-Archive "$($slack_cli_dir)\slack_cli.zip" -DestinationPath "$($slack_cli_dir)" -Force
+  Microsoft.PowerShell.Archive\Expand-Archive "$($slack_cli_dir)\slack_cli.zip" -DestinationPath "$($slack_cli_dir)" -Force
   Move-Item -Path $slack_cli_binary_path -Destination $slack_cli_new_binary_path -Force
 
   $User = [System.EnvironmentVariableTarget]::User
--- a/scripts/install-windows-test.ps1
+++ b/scripts/install-windows-test.ps1
@@ -14,23 +14,21 @@
 
 $ErrorActionPreference = "Stop"
 
-$qualifiedCommand = "Microsoft.PowerShell.Archive\Expand-Archive" # https://github.com/slackapi/slack-cli/issues/651
-$installDir = Join-Path $env:LOCALAPPDATA "slack-cli"
+# Guard against the Pscx-shadowing regression from https://github.com/slackapi/slack-cli/issues/651:
+# Pscx 3.3.2 ships its own Expand-Archive that shadows the built-in, so the extraction call must be
+# qualified as Microsoft.PowerShell.Archive\Expand-Archive. We assert this statically (parse the AST,
+# not run the installer) because the installers' post-install courtesy check hangs under pwsh 7 on
+# current Windows runner images -- `& slack _fingerprint | Tee-Object -Variable | Out-Null` blocks
+# with no console -- so a real end-to-end install is not yet runnable in CI. Tracked separately.
+$qualifiedCommand = "Microsoft.PowerShell.Archive\Expand-Archive"
 
-$cases = @(
-  @{ Installer = "install-windows.ps1"; Alias = "slack-test"; Version = "4.6.0"; ExpectVersion = $true },
-  @{ Installer = "install-windows-dev.ps1"; Alias = "slack-dev-test"; Version = "dev"; ExpectVersion = $false }
+$installers = @(
+  "install-windows.ps1",
+  "install-windows-dev.ps1"
 )
 
-function Remove-Install {
-  if (Test-Path -LiteralPath $installDir) {
-    Remove-Item -LiteralPath $installDir -Recurse -Force
-  }
-}
-
-foreach ($case in $cases) {
-  $installer = Join-Path $PSScriptRoot $case.Installer
-  $aliasBinary = Join-Path $installDir "bin\$($case.Alias).exe"
+foreach ($name in $installers) {
+  $installer = Join-Path $PSScriptRoot $name
 
   $tokens = $null
   $parseErrors = $null
@@ -51,26 +49,8 @@ foreach ($case in $cases) {
     }, $true)
 
   if ($archiveCommands.Count -ne 1 -or $archiveCommands[0].GetCommandName() -ne $qualifiedCommand) {
-    throw "$installer must call $qualifiedCommand exactly once"
+    throw "$installer must call $qualifiedCommand exactly once (issue #651)"
   }
 
-  try {
-    Remove-Install
-
-    & $installer -Alias $case.Alias -Version $case.Version -SkipGit $true
-
-    if (!(Test-Path -LiteralPath $aliasBinary)) {
-      throw "$($case.Installer) did not place $($case.Alias).exe at $aliasBinary (extraction failed?)"
-    }
-
-    $versionOutput = & $aliasBinary --version
-    if ($case.ExpectVersion -and $versionOutput -notmatch [regex]::Escape($case.Version)) {
-      throw "Version mismatch: expected '$($case.Version)' in output, got '$versionOutput'"
-    }
-
-    Write-Host "$($case.Installer) E2E passed: $($case.Alias).exe installed, version '$versionOutput'"
-  }
-  finally {
-    Remove-Install
-  }
+  Write-Host "$name calls $qualifiedCommand exactly once (issue #651 guard passed)"
 }
--- a/scripts/install-windows-test.ps1
+++ b/scripts/install-windows-test.ps1
@@ -14,13 +14,7 @@
 
 $ErrorActionPreference = "Stop"
 
-# Guard against the Pscx-shadowing regression from https://github.com/slackapi/slack-cli/issues/651:
-# Pscx 3.3.2 ships its own Expand-Archive that shadows the built-in, so the extraction call must be
-# qualified as Microsoft.PowerShell.Archive\Expand-Archive. We assert this statically (parse the AST,
-# not run the installer) because the installers' post-install courtesy check hangs under pwsh 7 on
-# current Windows runner images -- `& slack _fingerprint | Tee-Object -Variable | Out-Null` blocks
-# with no console -- so a real end-to-end install is not yet runnable in CI. Tracked separately.
-$qualifiedCommand = "Microsoft.PowerShell.Archive\Expand-Archive"
+$qualifiedCommand = "Microsoft.PowerShell.Archive\Expand-Archive" # https://github.com/slackapi/slack-cli/issues/651
 
 $installers = @(
   "install-windows.ps1",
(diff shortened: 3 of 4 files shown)


.github/workflows/tests.yml
```diff
@@ -34,6 +34,19 @@ jobs:
...
+    runs-on: windows-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+      - name: Test Windows installer
+        shell: pwsh
+        run: ./scripts/install-windows-test.ps1
```
mwbrooks: note: Every other install test in this repo is reachable through `make test-install`, which runs `scripts/install-test.sh` and `scripts/install-dev-test.sh`. This one is only reachable from CI, so a contributor on Windows has no obvious local entry point (your PR description is currently the only place the invocation is written down). A `test-install-windows` target would close that, though `make` on Windows is uncommon enough that CI-only is a defensible call.

scripts/install-windows-test.ps1
```diff
@@ -0,0 +1,70 @@
...
+    throw "PowerShell parser errors in $installerPath"
+  }
+
+  $archiveCommands = $ast.FindAll({
+      param($node)
+      $node -is [System.Management.Automation.Language.CommandAst] -and
+        $node.GetCommandName() -like "*Expand-Archive"
+    }, $true)
+
+  if ($archiveCommands.Count -ne 1 -or $archiveCommands[0].GetCommandName() -ne $qualifiedCommand) {
```
mwbrooks: suggestion: This asserts the call count, but the invariant that matters is qualification. If a future installer change legitimately extracts twice (both qualified), the test fails and the message points at the count rather than the thing that is actually wrong. Consider asserting that every `*Expand-Archive` command resolves to `$qualifiedCommand` and that there is at least one, which tests the property directly and survives an extra call site.

scripts/install-windows-test.ps1
```diff
@@ -0,0 +1,70 @@
...
+$sourcePath = Join-Path $testRoot "source"
+$destinationPath = Join-Path $testRoot "destination"
+$archivePath = Join-Path $testRoot "slack_cli.zip"
+
+try {
+  New-Item -ItemType Directory -Path (Join-Path $sourcePath "bin") -Force | Out-Null
+  Set-Content -LiteralPath (Join-Path $sourcePath "bin\slack.exe") -Value "test"
+  Microsoft.PowerShell.Archive\Compress-Archive -Path (Join-Path $sourcePath "*") -DestinationPath $archivePath
+
+  function Expand-Archive {
```
mwbrooks: praise: Shadowing with a local function rather than pulling in Pscx keeps this dependency-free, and it is a stronger control than the bug it guards: PowerShell resolves Alias > Function > Cmdlet, so a function shadows more aggressively than Pscx's cmdlet does via import order. Surviving this means surviving the real case.

scripts/install-windows-dev.ps1
```diff
@@ -163,7 +163,7 @@ function install_slack_cli {
   $slack_cli_new_binary_path = "$($slack_cli_dir)\bin\${confirmed_alias}.exe"
 
   delay 0.3 "Extracting the executable to:`n   $slack_cli_new_binary_path"
-  Expand-Archive "$($slack_cli_dir)\slack_cli.zip" -DestinationPath "$($slack_cli_dir)" -Force
+  Microsoft.PowerShell.Archive\Expand-Archive "$($slack_cli_dir)\slack_cli.zip" -DestinationPath "$($slack_cli_dir)" -Force
```
zimeg, line 166: 🌟 praise: Thanks for making this step more stable!

zimeg (approved):
@codywilliamson Super appreciate this changeset landing here! I'm finding it continues to work as we hope and the tests are nice to confirm this.

Before merging I'm hoping to make a few small changes to tests and CI so we can keep this confidence high 🏆

.github/workflows/tests.yml
```diff
@@ -25,15 +25,37 @@ jobs:
...
-      - name: Install Tests
-        run: make test-install
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@fb8b3582c8e4def4969c97caa2f19720cb33a72f # v7.0.0
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
           files: ./coverage.out
           fail_ci_if_error: false
 
+  install-tests:
```
mwbrooks: issue: Moving `make test-install` out of `Lints and Tests` and into this job takes install-test coverage off the merge gate. `main`'s required checks are `salesforce-cla`, `Check that license headers are in place`, `Lints and Tests`, `build`, and `e2e-test`, so `Install Tests (macos-latest|ubuntu-latest|windows-latest)` can go red and the PR still merges. This PR is the live demo: the Windows leg never passed and the branch still reads as mergeable. **This part is ours, not yours** - we need to add the three contexts to branch protection. Flagging it here so the gap is on the record next to the change that created it, and worth a note in the PR description so whoever merges knows to do it.

.github/workflows/tests.yml
```diff
@@ -25,15 +25,37 @@ jobs:
...
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+      - name: Install Tests (Unix)
+        if: runner.os != 'Windows'
+        run: make test-install
+      - name: Install Tests (Windows)
```
mwbrooks: issue: This job has no `timeout-minutes`, so when the Windows leg wedged it sat for 12m19s and then surfaced as CANCELLED rather than failed. Cancelled reads as neutral in the checks list, which is the worst outcome: a hang looks like nothing happened. `timeout-minutes: 15` on the job turns that into a red X.

scripts/install-windows-test.ps1
```diff
@@ -0,0 +1,76 @@
...
+    $installer,
+    [ref]$tokens,
+    [ref]$parseErrors
+  )
+
+  if ($parseErrors.Count -gt 0) {
+    throw "PowerShell parser errors in $installer"
+  }
+
+  $archiveCommands = $ast.FindAll({
```
mwbrooks: suggestion: Worth keeping the function-shadow case from the earlier revision alongside this. `windows-latest` has no Pscx, and the job runs `shell: pwsh` (PowerShell 7) while Pscx 3.3.2 is a 5.1 module, so the end-to-end run would stay green even if the qualification were reverted. That leaves the AST scan below as the only thing standing between us and #651 coming back. Defining a same-named `Expand-Archive` function was a faithful stand-in (a function outranks a cmdlet in resolution order, same observable as Pscx's import), and it was fast, cross-platform, and needed no network. The E2E is genuinely valuable, it just covers a different thing.

scripts/install-windows-test.ps1
```diff
@@ -0,0 +1,76 @@
...
+      param($node)
+      $node -is [System.Management.Automation.Language.CommandAst] -and
+        $node.GetCommandName() -like "*Expand-Archive"
+    }, $true)
+
+  if ($archiveCommands.Count -ne 1 -or $archiveCommands[0].GetCommandName() -ne $qualifiedCommand) {
```
mwbrooks: suggestion: This is sound today but asserts "exactly one call, and it is qualified" rather than the invariant that matters, "every `Expand-Archive` call site is qualified". A future installer that legitimately extracts twice fails the test, and the throw message points at the count instead of the qualification, which trains the next person to loosen it. Something like:

```powershell
if ($archiveCommands.Count -lt 1) { throw "$installer must call $qualifiedCommand" }
$unqualified = $archiveCommands | Where-Object { $_.GetCommandName() -ne $qualifiedCommand }
if ($unqualified) { throw "$installer must qualify every Expand-Archive call as $qualifiedCommand" }
```

Also worth running this scan over both installers up front, before the E2E loop. Right now it sits behind the previous case's network install, and that bit us here: case 1 wedged, so `install-windows-dev.ps1` was never checked at all.

scripts/install-windows-test.ps1
```diff
@@ -0,0 +1,76 @@
...
+    throw "$installer must call $qualifiedCommand exactly once"
+  }
+
+  try {
+    Remove-Install
+
+    & $installer -Alias $case.Alias -Version $case.Version -SkipGit $true
```
mwbrooks: issue: This line is where CI hangs, and it's case 1, not case 2. The log's last output is `Sweet! You're all set to start developing!` (`install-windows.ps1:262`), and the next statement is `terms_of_service`, which calls `check_slack_binary_exist`, which runs `& $SLACK_CLI_NAME _fingerprint` at line 58. The in-process `$Env:Path` update means `Get-Command slack-test` now resolves, so that is the first ever execution of the freshly downloaded binary, and nothing was emitted for the following twelve minutes.

One concrete lead: `install-windows-dev.ps1:27` sets `SLACK_DISABLE_TELEMETRY` process-wide and `install-windows.ps1` does not, and case 1 is the one that wedged. Cheapest bisect is `env: SLACK_DISABLE_TELEMETRY: "true"` on the job.

Separately, invoking with `& $installer` in the current session means the installer under test inherits `$ErrorActionPreference = "Stop"` from line 15, while real users run it under the default `Continue` with the installer's own `trap`. Running it as `pwsh -NoProfile -File $installer ...` with stdin redirected fixes both problems at once: it exercises the path users actually get, and a first-run prompt waiting on stdin can no longer wedge the job.

zimeg:
@codywilliamson Thanks once more for sending this in! I'm hoping in follow up changes we can strengthen CI coverage of this installation script but for now let's fix this issue so I will merge the PR.

.github/workflows/tests.yml
mwbrooks: issue: Splitting this into its own job takes install-test coverage off the merge gate. `make test-install` used to be a step inside `Lints and Tests`, which is a required check on `main`, so a broken installer blocked merge. `Install Tests (macos-latest)`, `(ubuntu-latest)`, and `(windows-latest)` are new contexts and are not in the required list, so right now installer breakage lands silently.

This half is ours to fix in branch protection, not yours, and it is worth doing with a second clause: fork PRs on this repo sit at `action_required` until a maintainer approves the workflow run, so making these contexts required without also approving fork-PR runs as part of the review rotation would park community PRs rather than gate them. Flagging so the coverage gap is on the record with the PR that created it.

.github/workflows/tests.yml
```diff
@@ -25,15 +25,37 @@ jobs:
...
+  install-tests:
+    name: Install Tests
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - macos-latest
+          - ubuntu-latest
+          - windows-latest
+    runs-on: ${{ matrix.os }}
```
mwbrooks: suggestion: Worth adding `timeout-minutes` here. It mattered more when the Windows leg ran a live install, but the macOS and Ubuntu legs still run `make test-install`, which downloads from the release CDN. A network stall there hits the 6-hour default and surfaces as CANCELLED, which reads as neutral in the checks list rather than as a failure. Something like `timeout-minutes: 10` is generous against the current 19s to 37s runtimes.

scripts/install-windows-test.ps1
```diff
@@ -0,0 +1,64 @@
...
+# limitations under the License.
+
+$ErrorActionPreference = "Stop"
+
+# Guard against the Pscx-shadowing regression from https://github.com/slackapi/slack-cli/issues/651:
+# Pscx 3.3.2 ships its own Expand-Archive that shadows the built-in, so the extraction call must be
+# qualified as Microsoft.PowerShell.Archive\Expand-Archive. We assert this statically (parse the AST,
+# not run the installer) because the installers' post-install courtesy check hangs under pwsh 7 on
+# current Windows runner images -- `& slack _fingerprint | Tee-Object -Variable | Out-Null` blocks
+# with no console -- so a real end-to-end install is not yet runnable in CI. Tracked separately.
```
mwbrooks: suggestion: "Tracked separately" would be more useful with the issue link. #651 is cited by number three times just above, so the contrast is noticeable, and the pwsh 7 hang is the reason this file is a parse rather than an install. Whoever revisits it will want the issue rather than this commit's archaeology.

The mechanism you captured is the useful part and I would keep it verbatim in the issue: `& slack _fingerprint | Tee-Object -Variable | Out-Null` blocking with no console attached.

scripts/install-windows-test.ps1
```diff
@@ -0,0 +1,64 @@
...
+$qualifiedCommand = "Microsoft.PowerShell.Archive\Expand-Archive"
+
+$installers = @(
```
mwbrooks: suggestion: A fixed list means a third installer, or extraction moving into a helper script, is silently uncovered rather than failing. `Get-ChildItem -Path $PSScriptRoot -Filter 'install-windows*.ps1'` (excluding this test file) keeps the guard honest as the directory changes, and the `Count -lt 1` check below already handles a file that legitimately does no extraction.

Not blocking, and the current two names are the complete set today.

PR #661, opened 2026-08-29
fix: honor explicit apihost throughout deploy

### LINKED ISSUE #129: bug: --apihost flag limitations
### Description

The `--apihost` flag is mainly used by the maintainers of this project to test various use cases. Using it as follows `--apihost=api.shhhh.slack.com` mostly works as expected. But when combine with the `deploy` command having the `api` domain prefix breaks the logic that sets the `SLACK_API_URL` environment variable

### Version
```
Using lack v3.3.0
```
### OS Info
```
ProductName:            macOS
ProductVersion:         15.5
BuildVersion:           24F74
Darwin Kernel Version 24.5.0: Tue Apr 22 19:54:49 PDT 2025; root:xnu-11417.121.6~2/RELEASE_ARM64_T6000
```
#### Steps to reproduce:

Share the commands to run, source code, and project settings:

1. Create an app
2. Use commands with the `--apihost` flag set to a value prefixed with `api.`
     a. everything should work as expected 
3. Use `slack deploy --apihost=api.your.test.endpoint -vv` 
4. No error is raised and no trace of a request to set the `SLACK_API_URL` environment variable can be found
5. If the flag value is set without the `api` prefix the `SLACK_API_URL` environment variable is set and things work as expected

### Expected result:

If the `apihost` value is rejected the CLI expects then we should inform the user

### Actual result:

Nothing indicates that the `apihost` value is rejected by the CLI

### Requirements

* [x] I've read and understood the [Contributing guidelines](/blob/main/.github/contributing.md) and have done my best effort to follow them.
* [x] I've searched for any related issues and avoided creating a duplicate issue.

mvanhorn:
### Changelog

- Treat a non-empty `APIHostFlag` as the signal that this invocation is intentionally targeting a custom endpoint: preserve its existing precedence in `ResolveAPIHost`, and bypass automatic rotation of saved credentials so rotation cannot contact a different stored host. In the deploy path, use the resolved explicit host for `SLACK_API_URL` regardless of whether its first hostname label begins with `dev`, while retaining current behavior for invocations without an explicit custom host. Move the environment-variable update into a small production helper called by `deployApp` so host-selection and error behavior can be tested directly, and propagate an `AddVariable` failure through the existing deploy error path rather than discarding it.

### Summary

Treat a non-empty `APIHostFlag` as the signal that this invocation is intentionally targeting a custom endpoint: preserve its existing precedence in `ResolveAPIHost`, and bypass automatic rotation of saved credentials so rotation cannot contact a different stored host. In the deploy path, use the resolved explicit host for `SLACK_API_URL` regardless of whether its first hostname label begins with `dev`, while retaining current behavior for invocations without an explicit custom host. Move the environment-variable update into a small production helper called by `deployApp` so host-selection and error behavior can be tested directly, and propagate an `AddVariable` failure through the existing deploy error path rather than discarding it.

The hidden `--apihost` flag is intended to direct a CLI invocation to a custom Slack API endpoint, but deploy only writes `SLACK_API_URL` when the resolved hostname matches the narrower `dev*` convention. As a result, an explicit endpoint such as `api.your.test.endpoint` can handle the CLI's API requests while being silently omitted from the deployed app environment. The thread also identifies that credential loading can rotate tokens against a saved host before the explicitly requested host takes effect. The fix must make the explicit flag authoritative for the invocation and report failures instead of silently completing a deploy with inconsistent configuration.

Fixes #129

### Preview

(Add screenshots, GIFs, or recordings that show the changes)

### Testing

(List the steps used to verify these changes)

### Notes

(Add any additional context, trade-offs, or follow-up items)

### Requirements

AI was used for assistance.

### DIFF
--- a/internal/auth/auth.go
+++ b/internal/auth/auth.go
@@ -238,7 +238,11 @@ func (c *Client) auths(ctx context.Context) (map[string]types.SlackAuth, error) 
 	var updatedAuthsByTeamID types.AuthByTeamID
 	var hasUpdateRotation, hasUpdateNaming bool
 
-	updatedAuthsByName, hasUpdateRotation = c.rotateTokenAll(ctx, auths)
+	if c.config.APIHostFlag != "" {
+		updatedAuthsByName = auths
+	} else {
+		updatedAuthsByName, hasUpdateRotation = c.rotateTokenAll(ctx, auths)
+	}
 
 	// As of v2.4.0 we migrate users credentials json to storing auths by team_id
 	updatedAuthsByTeamID, hasUpdateNaming = c.migrateToAuthByTeamID(ctx, updatedAuthsByName)
--- a/internal/auth/auth_test.go
+++ b/internal/auth/auth_test.go
@@ -374,6 +374,31 @@ func Test_AuthsRotation(t *testing.T) {
 		require.NoError(t, err, "Should not return an error when the contents of credentials are valid")
 	})
 
+	t.Run("token rotation is skipped with an explicit API host", func(t *testing.T) {
+		ctx, authClient := setup(t)
+		fiveMinutesAgo := int(time.Now().Unix()) - 60*5
+		savedAPIHost := "https://saved.slack.test"
+		authClient.config.APIHostFlag = "api.your.test.endpoint"
+
+		workspaceAuth := types.SlackAuth{
+			APIHost:      &savedAPIHost,
+			Token:        "expiredToken",
+			RefreshToken: "valid-refresh-token",
+			ExpiresAt:    fiveMinutesAgo,
+			TeamDomain:   "workspace-a",
+			TeamID:       "T123456789A",
+		}
+		_, err := authClient.setAuths(ctx, types.AuthByTeamDomain{
+			workspaceAuth.TeamID: workspaceAuth,
+		})
+		require.NoError(t, err)
+
+		updatedAuths, err := authClient.auths(ctx)
+
+		require.NoError(t, err)
+		require.Equal(t, workspaceAuth, updatedAuths[workspaceAuth.TeamID])
+	})
+
 	t.Run("token rotation returns an error", func(t *testing.T) {
 		// Setup
 		ctx, authClient := setup(t)
@@ -566,6 +591,7 @@ func Test_SetSelectedAuth(t *testing.T) {
 	mockAPIHost := "dev.slack.com"
 	tests := map[string]struct {
 		auth            types.SlackAuth
+		apiHostFlag     string
 		expectedAPIHost string
 		expectedAPIURL  string
 	}{
@@ -577,10 +603,20 @@ func Test_SetSelectedAuth(t *testing.T) {
 			expectedAPIHost: fmt.Sprintf("https://%s", mockAPIHost),
 			expectedAPIURL:  fmt.Sprintf("https://%s/api/", mockAPIHost),
 		},
+		"explicit API host overrides authentication host": {
+			auth: types.SlackAuth{
+				TeamID:  "T002",
+				APIHost: &mockAPIHost,
+			},
+			apiHostFlag:     "api.your.test.endpoint",
+			expectedAPIHost: "https://api.your.test.endpoint",
+			expectedAPIURL:  "https://api.your.test.endpoint/api/",
+		},
 	}
 	for name, tc := range tests {
 		t.Run(name, func(t *testing.T) {
 			ctx, authClient, osMock := setup(t)
+			authClient.config.APIHostFlag = tc.apiHostFlag
 			authClient.SetSelectedAuth(ctx, tc.auth, authClient.config, osMock)
 			assert.Equal(t, authClient.config.TeamFlag, tc.auth.TeamID)
 			assert.Equal(t, authClient.config.APIHostResolved, tc.expectedAPIHost)
--- a/internal/pkg/platform/deploy.go
+++ b/internal/pkg/platform/deploy.go
@@ -185,19 +185,27 @@ func deployApp(ctx context.Context, clients *shared.ClientFactory, app types.App
 	var elapsedDeploy = time.Since(startDeploy)
 	var deployTime = fmt.Sprintf("%.1fs", elapsedDeploy.Seconds())
 
-	// Set the SLACK_API_URL environment variable for development workspaces
-	//
-	// Note: This errors silently to continue deployment without any problem
-	var apiHost = clients.Config.APIHostResolved
-	if clients.Auth().IsAPIHostSlackDev(apiHost) {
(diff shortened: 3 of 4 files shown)


internal/pkg/platform/deploy.go
```diff
@@ -200,6 +195,19 @@ func deployApp(ctx context.Context, clients *shared.ClientFactory, app types.App
 	return nil
 }
 
+// setAPIHostVariable configures deployed apps to use the resolved custom API
+// host when the invocation explicitly requests one or targets Slack development.
+func setAPIHostVariable(ctx context.Context, clients *shared.ClientFactory, appID string) error {
+	apiHost := clients.Config.APIHostResolved
+	if clients.Config.APIHostFlag == "" && !clients.Auth().IsAPIHostSlackDev(apiHost) {
```
mwbrooks: suggestion: Consider keying this on the resolved host rather than on the flag: `if apiHost == "" || clients.Auth().IsAPIHostSlackProd(apiHost) { return nil }`. As written, a developer logged into a saved custom or enterprise host who deploys without `--apihost` still gets no `SLACK_API_URL`, which is the same shape of bug as #129. `run.go:119` already sets the variable unconditionally from `APIHostResolved` for local runs, so this would bring deploy in line with run and drop the need to consult two config fields.

internal/pkg/platform/deploy.go
```diff
@@ -185,13 +185,8 @@ func deployApp(ctx context.Context, clients *shared.ClientFactory, app types.App
...
 
-	// Set the SLACK_API_URL environment variable for development workspaces
-	//
-	// Note: This errors silently to continue deployment without any problem
-	var apiHost = clients.Config.APIHostResolved
-	if clients.Auth().IsAPIHostSlackDev(apiHost) {
-		apiHostURL := fmt.Sprintf("%s/api/", apiHost)
-		_ = clients.API().AddVariable(ctx, token, app.AppID, "SLACK_API_URL", apiHostURL)
+	if err := setAPIHostVariable(ctx, clients, app.AppID); err != nil {
+		return err
```
mwbrooks: suggestion: Worth wrapping this one, since by the time it fires `UploadApp` has already succeeded and the app is live on the platform. Every other error in `deployApp` gets context (`error uploading app: %s`, `slackerror.Wrapf`), so a bare pass-through here surfaces something like `app_not_found` with no hint that the deploy itself worked and only the variable write did not. A wrap with remediation pointing at `slack env set SLACK_API_URL <host>` would make the failure actionable, and it addresses the "we should inform the user" ask in #129 directly.

internal/pkg/platform/deploy.go
```diff
@@ -200,6 +195,19 @@ func deployApp(ctx context.Context, clients *shared.ClientFactory, app types.App
...
+// host when the invocation explicitly requests one or targets Slack development.
+func setAPIHostVariable(ctx context.Context, clients *shared.ClientFactory, appID string) error {
+	apiHost := clients.Config.APIHostResolved
+	if clients.Config.APIHostFlag == "" && !clients.Auth().IsAPIHostSlackDev(apiHost) {
+		return nil
+	}
+
+	token := config.GetContextToken(ctx)
+	apiHostURL := fmt.Sprintf("%s/api/", apiHost)
+	return clients.API().AddVariable(ctx, token, appID, "SLACK_API_URL", apiHostURL)
```
mwbrooks: question: Since this writes app-level state that outlives the invocation, what happens on the next deploy without `--apihost`? The resolved host is prod, the guard above returns early, and the app keeps calling the custom endpoint indefinitely. The old `dev*` guard kept that footgun narrow, and widening it to any explicit host makes a stale value more likely, so it may be worth resetting or removing the variable when the resolved host comes back to production (happy to track that as a follow-up rather than growing this PR).

internal/auth/auth.go
```diff
@@ -238,7 +238,11 @@ func (c *Client) auths(ctx context.Context) (map[string]types.SlackAuth, error)
 	var updatedAuthsByTeamID types.AuthByTeamID
 	var hasUpdateRotation, hasUpdateNaming bool
 
-	updatedAuthsByName, hasUpdateRotation = c.rotateTokenAll(ctx, auths)
+	if c.config.APIHostFlag != "" {
```
mwbrooks: suggestion: This skips rotation for every saved auth, including the one this invocation is about to use, so an expired-but-refreshable token for the host you explicitly asked for is no longer refreshed and the API returns `invalid_auth` instead. Narrowing the skip to auths whose stored `APIHost` does not match the resolved flag host would still stop rotation from reaching unrelated endpoints (the behaviour @zimeg asked for in #129) while keeping the transparent refresh on the host you are targeting.

internal/auth/auth.go
```diff
@@ -238,7 +238,11 @@ func (c *Client) auths(ctx context.Context) (map[string]types.SlackAuth, error)
...
+		updatedAuthsByName = auths
+	} else {
+		updatedAuthsByName, hasUpdateRotation = c.rotateTokenAll(ctx, auths)
```
mwbrooks: note: Taking a position that this guard treats the symptom rather than the cause. In `rotateToken` below, the `if err != nil` branch returns before `c.api.SetHost(activeAPIHostBeforeRotation)` runs, and `clients.API()` hands back a single shared client, so one failed rotation leaves the whole process pointed at that auth's saved host. Map iteration order decides which host wins, which matches the nondeterministic `dev2126` vs `dev1815` mismatch reported in #129. Deferring the restore would fix it for invocations without `--apihost` too, either here or as a follow-up.

internal/auth/auth_test.go
```diff
@@ -374,6 +374,31 @@ func Test_AuthsRotation(t *testing.T) {
...
+		}
+		_, err := authClient.setAuths(ctx, types.AuthByTeamDomain{
+			workspaceAuth.TeamID: workspaceAuth,
+		})
+		require.NoError(t, err)
+
+		updatedAuths, err := authClient.auths(ctx)
+
+		require.NoError(t, err)
+		require.Equal(t, workspaceAuth, updatedAuths[workspaceAuth.TeamID])
```
mwbrooks: suggestion: This assertion would also pass if rotation had run and failed, because `rotateTokenAll` swallows errors and returns the auth unchanged. Right now the only thing catching a regression is a panic on the nil API client from `NewClient(nil, nil, ...)`. Installing an `api.NewFakeClient` whose handler fails the test on any call (the way the `token rotation needed for at least one of the auths` case above sets one up) would assert the security-relevant part, that no request leaves for `https://saved.slack.test`. Worth noting too that this fixture pins the new behaviour where an expired token with a valid refresh token stays expired, so it is the right place to say so in a comment.

vegeris:
Hi there, 

Thank you for contributing! Given that the `SLACK_API_URL` override would only be applicable to internal users, I'm going to close this PR to let maintainers handle it

PR #669, opened 2026-09-09
revert: charmbracelet group bump #636 (lipgloss + ultraviolet) — lingering selection output

zimeg:
### Changelog

Fix interactive selection prompts (e.g. the `slack create` category picker) leaving stale/lingering rendered lines on screen after a selection is made.

### Summary

Reverts the charmbracelet group bump from #636 across **both** deps it changed. #636 was titled as a `charm.land/lipgloss/v2` v2.0.5 → v2.0.6 bump, but its go.mod diff also carried `github.com/charmbracelet/ultraviolet` `f5a850f9` → `006e29f9`. The lingering-selection regression is in **ultraviolet's terminal renderer**, not lipgloss:

- `slack create` → select a category leaves residual lines on screen after the pick.
- The teardown/erase behavior is owned by ultraviolet (huh → bubbletea → ultraviolet's renderer), which was substantially rewritten between `f5a850f9` and `006e29f9` (new line-drift/repaint handling and changed inline erase-height logic). The wide selector glyph exercises exactly that path.
- lipgloss v2.0.6 requires the new ultraviolet via MVS, so the two are coupled — pinning lipgloss alone left this repo resolving ultraviolet to `006e29f9`, and the regressed renderer was never actually reverted. Both are pinned back together: lipgloss → v2.0.5, ultraviolet → `f5a850f9`.

An earlier revision of this PR pinned only lipgloss; that did **not** fix the symptom (verified on a real terminal — selections still lingered), which is what surfaced ultraviolet as the true cause.

### Testing

- `go build ./...` passes with lipgloss v2.0.5 + ultraviolet `f5a850f9`.
- `slack create` → **Automation App** category selection verified by hand on a real terminal: prompt tears down cleanly, no lingering lines. The prior config (lipgloss v2.0.5, ultraviolet still `006e29f9`) reproduced the lingering; pinning ultraviolet back to `f5a850f9` resolves it.
- The lingering artifact only renders through the interactive inline renderer on a real TTY (accessible-mode / model-stepping tests bypass it), so it is verified by hand rather than in headless CI.

### Notes

- The dependabot `ignore` rule fences **both** `charm.land/lipgloss/v2` (`>= 2.0.6`) and `github.com/charmbracelet/ultraviolet` (`>= 006e29f9`) so the next charmbracelet group bump can't silently re-raise either half and undo the revert.
- This is an **interim pin** — a follow-up issue upstream will track the proper ultraviolet fix (the renderer's erase-sizing on lines with wide glyphs), so we can un-pin both deps together once it ships.

### Requirements


mwbrooks (approved):

srtaalej (approved):

zimeg:
@mwbrooks @srtaalej Super appreciate the fast reviews! More testing finds that pinning the `ultraviolet` dependencies required from **lipgloss** keeps this change stable 💄 ✨ 

Let's merge this now but search for upstream fix to remove our @dependabot changes here since I don't believe we can unit test this with confidence at this immediate time.

PR #641, opened 2026-08-17
feat: support the code_channels manifest feature

zimeg:
### Changelog

Add support for the `code_channels` app manifest feature.

### Summary

Adds the `code_channels` feature block to the manifest types so the CLI can read, write, and round-trip it (`enabled`, optional `slash_command_url`). Follows the existing feature-block pattern; validation is handled by the manifest validation endpoint.

### Requirements

- [x] I've read and understood the [Contributing Guidelines](https://github.com/slackapi/slack-cli/blob/main/.github/CONTRIBUTING.md) and have done my best effort to follow them.
- [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).

PR #670, opened 2026-09-09
feat: support features.code_channels in app manifest round trip

mcodik:
### Changelog

Lets devs set code_channels feature in their manifests, and have it display in `manifest export` 

### Summary

AppFeatures had no field for code_channels, so Go's json.Unmarshal silently dropped it before the manifest was sent to apps.manifest.validate/update. Add it as a RawJSON passthrough so any future sub-fields survive without further struct changes.

### Preview

Before
Screenshot 2026-09-09 at 2 24 12 PM

After

Screenshot 2026-09-09 at 2 23 43 PM

### Testing

Run `manifest export` on app that has the code_channels feature set in the manifest

### Notes

(Add any additional context, trade-offs, or follow-up items)

### Requirements

- [ ] I've read and understood the [Contributing Guidelines](https://github.com/slackapi/slack-cli/blob/main/.github/CONTRIBUTING.md) and have done my best effort to follow them.
- [ ] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).

### DIFF
--- a/internal/shared/types/app_manifest.go
+++ b/internal/shared/types/app_manifest.go
@@ -79,6 +79,7 @@ type AppFeatures struct {
 	ManifestSlashCommandsItems []ManifestSlashCommandsItem `json:"slash_commands,omitempty" yaml:"slash_commands,flow,omitempty"`
 	Search                     *Search                     `json:"search,omitempty" yaml:"search,flow,omitempty"`
 	RichPreviews               *RichPreviews               `json:"rich_previews,omitempty" yaml:"rich_previews,flow,omitempty"`
+	CodeChannels               *RawJSON                    `json:"code_channels,omitempty" yaml:"code_channels,flow,omitempty"`
 }
 
 type RichPreviews struct {
--- a/internal/shared/types/app_manifest_test.go
+++ b/internal/shared/types/app_manifest_test.go
@@ -250,6 +250,15 @@ func Test_AppManifest_AppFeatures(t *testing.T) {
 			},
 			want: `{"app_home":{},"bot_user":{"display_name":"business_bot"},"rich_previews":{"entity_types":["slack#/entities/file"]}}`,
 		},
+		"includes code channels when provided": {
+			features: AppFeatures{
+				BotUser: BotUser{
+					DisplayName: "codebot",
+				},
+				CodeChannels: ToRawJSON(`{"enabled":true}`),
+			},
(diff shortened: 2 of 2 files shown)


internal/shared/types/app_manifest.go
```diff
@@ -79,6 +79,7 @@ type AppFeatures struct {
 	ManifestSlashCommandsItems []ManifestSlashCommandsItem `json:"slash_commands,omitempty" yaml:"slash_commands,flow,omitempty"`
 	Search                     *Search                     `json:"search,omitempty" yaml:"search,flow,omitempty"`
 	RichPreviews               *RichPreviews               `json:"rich_previews,omitempty" yaml:"rich_previews,flow,omitempty"`
+	CodeChannels               *RawJSON                    `json:"code_channels,omitempty" yaml:"code_channels,flow,omitempty"`
```
zimeg, line 82: 👁️‍🗨️ question: Should we extend values for `code_channels` to include the following or are these values sometimes changed?:

```go
type CodeChannels struct {
	Enabled         *bool  `json:"enabled,omitempty" yaml:"enabled,omitempty"`
	SlashCommandURL string `json:"slash_command_url,omitempty" yaml:"slash_command_url,omitempty"`
}
```

zimeg (approved):
@mcodik Kind thanks for following up with this! 🏆 ✨ 

I think this is good to merge whenever but would like to align on changes of https://github.com/slackapi/slack-cli/pull/641 🪬

internal/shared/types/app_manifest.go
zimeg: > Add it as a RawJSON passthrough so any future sub-fields survive without further struct changes.

🌚 note: Am open to this as well but think we benefit from strict structures overall after stable API releases!

zimeg:
📚 note: I'm open to keeping this with `changelog` despite a limited GA so that updates to the upcoming release can guarantee support instead of magic later.

mcodik:
nice, I hadnt seen #641 . I'm ok with you merging that instead! re RawJSON vs specified, I picked RawJSON here because its a part of the manifest the CLI doesnt process, so if we ever add new fields to that feature those wont silently get dropped when handled by old CLIs that dont know about the field

zimeg:
@mcodik Thanks for some confirmation to these fields 👾 ✨ 

I'm leaning toward merging #641 for now with much willingness to fast follow releases as needed if more fields appear. So let's merge that then close this!

zimeg:
@mcodik Now is the time to close this PR. Thanks once more for bumping these features 🪬

